PR #5, opened 2017-12-26
Codec in subproject

alari:


codec/kryo/src/main/scala/fluence/codec/kryo/KryoCodecs.scala
```diff
@@ -42,18 +42,13 @@ class KryoCodecs[F[_], L <: HList] private (pool: KryoPool)(implicit F: Applicat
...
-            F.raiseError[Array[Byte]](new NullPointerException("Obj is null, encoding is impossible"))
-        }
-
-      override def decode(binary: Array[Byte]): F[T] =
-        F.catchNonFatal(pool.fromBytes(binary).asInstanceOf[T])
-    }
+    Codec(
+      obj ⇒ Option(obj) match {
+        case Some(o) ⇒
+          pool.toBytesWithClass(o).pure[F]
```
RicoGit: This method isn't pure, and may throws unchecked exception or returns null. It's should be wrapped, perhaps

PR #7, opened 2017-12-27
Btree refactoring

RicoGit:


build.sbt
```diff
@@ -165,17 +165,33 @@ lazy val `storage` = project.in(file("storage"))
...
+  .settings(commons)
+  .settings(
+    libraryDependencies ++= Seq(
+      monix3,
+      logback,
+      scalatest
+    )
+  ).dependsOn(`crypto`)
+
+lazy val `b-tree-protocol` = project.in(file("b-tree/protocol"))
```
alari: Protocol should not depend on monix and logback. Protocol should contain only DTOs and RPC traits, possibly with cats as dependency. Otherwise it's not a protocol, it's a part of protocol implementation.

PR #15, opened 2018-01-25
Add hash of dataset to merkle tree

RicoGit:


project/build.properties
```diff
@@ -1 +1 @@
-sbt.version = 1.1.0
\ No newline at end of file
+sbt.version = 1.0.4
```
alari: May you avoid committing it?

RicoGit: ok

PR #25, opened 2018-02-01
Edsa signature

DieMyst:


dataset/client/src/main/scala/fluence/dataset/contract/ContractRead.scala
```diff
@@ -114,42 +118,42 @@ object ContractRead {
...
+    def checkOfferSignature[F[_]](signature: Signature, checker: SignatureChecker)(implicit F: MonadError[F, Throwable]): F[Boolean] =
       checker.check(signature, getOfferBytes)
 
     /**
      * @return Whether this contract is a valid blank offer (with no participants, with client's signature)
      */
-    def isBlankOffer(checker: SignatureChecker): Boolean =
-      participants.isEmpty && checkOfferSeal(checker)
+    def isBlankOffer[F[_]](checker: SignatureChecker)(implicit F: MonadError[F, Throwable]): F[Boolean] =
+      checkOfferSeal(checker).map(_ && participants.isEmpty)
```
alari: Idea was to check that `participants.isEmpty` first, as this operation is cheap compared to signature checking

dataset/client/src/main/scala/fluence/dataset/contract/ContractRead.scala
```diff
@@ -158,31 +162,37 @@ object ContractRead {
...
      *
      * @param checker Signature checker
      */
-    def checkAllParticipants(checker: SignatureChecker): Boolean =
-      participants.size == participantsRequired &&
-        participants.forall(participantSigned(_, checker))
+    def checkAllParticipants[F[_]](checker: SignatureChecker)(implicit F: MonadError[F, Throwable]): F[Boolean] = {
+      for {
+        participants ← Traverse[List].traverse(participants.map(participantSigned(_, checker)).toList)(identity)
+      } yield participants.size == participantsRequired && participants.forall(identity)
```
alari: The same here, comparing sizes is cheap, so it's better to avoid traversing signatures until participants size fits

dataset/client/src/main/scala/fluence/dataset/contract/ContractWrite.scala
```diff
@@ -50,14 +51,14 @@ object ContractWrite {
   implicit class WriteOps[C](contract: C)(implicit read: ContractRead[C], write: ContractWrite[C]) {
     import ContractRead.ReadOps
 
-    def sealOffer(signer: Signer): C =
-      write.setOfferSeal(contract, signer.sign(contract.getOfferBytes))
+    def sealOffer[F[_]](signer: Signer)(implicit F: MonadError[F, Throwable]): F[C] =
+      signer.sign(contract.getOfferBytes).flatMap(s ⇒ F.catchNonFatal(write.setOfferSeal(contract, s)))
```
alari: `write.setOfferSeal` is a pure function, as it's just a lens call. No need to do flatMapping and catching here

dataset/client/src/main/scala/fluence/dataset/contract/ContractRead.scala
```diff
@@ -88,7 +89,7 @@ trait ContractRead[C] {
 
 object ContractRead {
 
-  implicit class ReadOps[C](contract: C)(implicit read: ContractRead[C]) {
+  implicit class ReadOps[F[_], C](contract: C)(implicit read: ContractRead[C], F: MonadError[F, Throwable]) {
```
alari: Previously we've had possibility to read static fields from the contract, like `.id`, without `MonadError` in scope. Now we have to define F and provide MonadError, otherwise implicit conversion to `ReadOps` is undefined.

crypto/src/main/scala/fluence/crypto/algorithm/Ecdsa.scala
```diff
@@ -36,104 +37,99 @@ import scala.util.control.NonFatal
...
-  }
+  private lazy val getSignatureProvider =
+    nonFatalHandling(Signature.getInstance(scheme, BouncyCastleProvider.PROVIDER_NAME))("Cannot get signature instance.")
 }
 
 object Ecdsa {
-  val ecdsa_secp256k1_sha256 = new Ecdsa("secp256k1", "SHA256withECDSA")
+  def ecdsa_secp256k1_sha256[F[_]](implicit F: MonadError[F, Throwable]) = new Ecdsa("secp256k1", "SHA256withECDSA")
+
+  class EcdsaSigner(keyPair: KeyPair) extends DataSigner {
```
alari: It should be just `Signer`?

crypto/src/main/scala/fluence/crypto/signature/DataSigner.scala
```diff
@@ -18,29 +18,20 @@
 package fluence.crypto.signature
 
 import cats.MonadError
-import fluence.crypto.algorithm.Ecdsa
 import fluence.crypto.keypair.KeyPair
 import scodec.bits.ByteVector
 
-trait Signer {
+trait DataSigner {
```
alari: Why to rename it? "Data" is meaningless here. JDK-specific names should not be used outside Crypto submodule, so even if you make things simpler here, but at the cost of higher complexity for module users.

PR #28, opened 2018-02-05
File storage with init info

DieMyst:


node/src/main/scala/fluence/node/NodeApp.scala
```diff
@@ -27,81 +33,134 @@ import monix.execution.Scheduler.Implicits.global
...
+    val appDir = new File(fluencePath)
+    if (!appDir.exists()) {
+      Files.createParentDirs(appDir)
+      appDir.mkdir()
+    }
+    ()
+  }
 
-  val log = LoggerFactory.getLogger(getClass)
+  def readConfig(configPath: Option[String]): Config = {
```
alari: Why? Isn't it already managed with ConfigFactory?

node/src/main/scala/fluence/node/NodeApp.scala
```diff
@@ -27,81 +33,134 @@ import monix.execution.Scheduler.Implicits.global
...
+          .withFallback(ConfigFactory.defaultReference())
+          .resolve()
+    }
 
-  val gitHash = ConfigFactory.load().getString("fluence.gitHash")
+    config
+  }
 
-  println(Console.CYAN + "Git Commit Hash: " + gitHash + Console.RESET)
+  def getKeyPair(keyPath: String): Either[Throwable, KeyPair] = {
```
alari: `Either[Throwable, T]` is the same as `Try[T]`

crypto/src/main/scala/fluence/crypto/algorithm/JavaAlgorithm.scala
```diff
@@ -23,16 +23,22 @@ import fluence.crypto.keypair.KeyPair
...
+import scala.language.{higherKinds, implicitConversions}
 
-private[algorithm] trait JavaAlgorithm[F[_]] extends Algorithm[F] {
+private[crypto] trait JavaAlgorithm extends Algorithm {
   JavaAlgorithm.addProvider
 }
 
 object JavaAlgorithm {
+  /**
+   * getEncoded method return ASN.1 encoding in PKCS#8 standard for secret key and X.509 standard for public
```
alari: Could you please provide some links describing why these standards matter?

crypto/src/main/scala/fluence/crypto/FileKeyStorage.scala
```diff
@@ -0,0 +1,79 @@
...
+import fluence.crypto.algorithm.JavaAlgorithm
+import fluence.crypto.keypair.KeyPair
+import io.circe.parser._
+import io.circe.syntax._
+import io.circe.{ Decoder, Encoder, HCursor, Json }
+import scodec.bits.ByteVector
+
+case class KeyStore(keyPair: KeyPair)
+
+class FileKeyStorage[F[_]](file: File)(implicit F: MonadError[F, Throwable]) extends JavaAlgorithm {
```
alari: Could you please explain in docs why `FileKeyStorage` extends `JavaAlgorithm`? Can't it simply import some functions from it?  Why should it?

crypto/src/main/scala/fluence/crypto/algorithm/JavaAlgorithm.scala
```diff
@@ -23,16 +23,22 @@ import fluence.crypto.keypair.KeyPair
 import org.bouncycastle.jce.provider.BouncyCastleProvider
 import scodec.bits.ByteVector
 
-import scala.language.higherKinds
+import scala.language.{higherKinds, implicitConversions}
 
-private[algorithm] trait JavaAlgorithm[F[_]] extends Algorithm[F] {
+private[crypto] trait JavaAlgorithm extends Algorithm {
```
alari: Documentation needed

crypto/src/test/scala/fluence.crypto/SignatureSpec.scala
```diff
@@ -58,5 +62,29 @@ class SignatureSpec extends WordSpec with Matchers with BeforeAndAfterAll {
       the[CryptoErr] thrownBy Ecdsa.Checker.check(sign.copy(sign = rndByteVector(10)), data).get
       the[CryptoErr] thrownBy Ecdsa.Checker.check(sign.copy(publicKey = sign.publicKey.copy(value = rndByteVector(10))), data).get
     }
+
+    "store and read key from file" in {
+      val seed = rndBytes(32)
+      val keys = Ecdsa.ecdsa_secp256k1_sha256.generateKeyPair(new SecureRandom(seed)).get
+
+      val keyFile = new File("/tmp/key")
```
alari: Let's use `File.createTempFile` or smth like that

node/src/main/scala/fluence/node/NodeApp.scala
```diff
@@ -27,81 +33,114 @@ import monix.execution.Scheduler.Implicits.global
...
+import scala.util.{ Failure, Success, Try }
 
-object NodeApp extends App {
+object NodeApp {
 
-  val log = LoggerFactory.getLogger(getClass)
+  def initDirectory(fluencePath: String): Unit = {
+    val appDir = new File(fluencePath)
+    if (!appDir.exists()) {
+      Files.createParentDirs(appDir)
```
alari: Is it possible to do the same without `com.google.common`? E.g. https://stackoverflow.com/questions/3090761/how-to-create-a-new-file-together-with-missing-parent-directories

node/src/main/scala/fluence/node/NodeApp.scala
```diff
@@ -27,81 +33,114 @@ import monix.execution.Scheduler.Implicits.global
...
+      appDir.mkdir()
+    }
+    ()
+  }
 
-  val gitHash = ConfigFactory.load().getString("fluence.gitHash")
+  def getKeyPair(keyPath: String): Try[KeyPair] = {
+    val keyFile = new File(keyPath)
+    val keyStorage = new FileKeyStorage[Try](keyFile)
+    val keyPair = if (keyFile.exists()) {
```
alari: Probably it worth moving to `FileKeyStorage`: it might check that file exists, and if it doesn't, use provided `=> F[KeyPair]` to generate, store and return a new one. In this case, we won't make a mistake doing the same "read keypair from path" again, and keep this file more concise & clean

node/src/main/scala/fluence/node/NodeApp.scala
```diff
@@ -27,81 +33,114 @@ import monix.execution.Scheduler.Implicits.global
...
+    keyPair
+  }
 
-  // For demo purposes
-  val keySeed = StdIn.readLine(Console.CYAN + "Who are you?\n> " + Console.RESET).getBytes()
-  val keyPair = KeyPair.fromBytes(keySeed, keySeed)
+  def cmd(s: String): Unit = println(Console.BLUE + s + Console.RESET)
 
-  val nodeComposer = new NodeComposer(keyPair, () ⇒ Task.now(NodeInfo(gitHash)))
+  def main(args: Array[String]): Unit = {
```
alari: btw, why do we need `main`? Why is it better than `... extends App`? Just curious

node/src/main/scala/fluence/node/NodeApp.scala
```diff
@@ -27,81 +33,114 @@ import monix.execution.Scheduler.Implicits.global
...
-
-      case "s" | "seed" ⇒
-        server.contact.map(_.b64seed).runAsync.foreach(cmd)
-
-      case "q" | "quit" | "x" | "exit" ⇒
-        cmd("exit")
-        System.exit(0)
-
-      case _ ⇒
+    if (keyPairE.isFailure) {
```
alari: Probably it's better to do `keyPairE.failed.foreach{e => ...}` -- to avoid ".get", and ".isFailure". `System.exit` will work the same way as it does it now

project/SbtCommons.scala
```diff
@@ -50,6 +51,12 @@ object SbtCommons {
   val scodecBits = "org.scodec" %% "scodec-bits" % "1.1.5"
   val bouncyCastle = "org.bouncycastle" % "bcprov-jdk15on" % "1.59"
 
+  val circe: Seq[ModuleID] = Seq(
+    "io.circe" %% "circe-core",
+    "io.circe" %% "circe-generic",
```
alari: Do you always need all the three modules?

PR #33, opened 2018-02-07
Dataset composition

alari:


transport/grpc/src/main/scala/fluence/transport/grpc/server/GrpcServerConf.scala
```diff
@@ -28,9 +31,10 @@ case class GrpcServerConf(
 object GrpcServerConf {
   val ConfigPath = "fluence.transport.grpc.server"
 
-  def read(path: String = ConfigPath, config: Config = ConfigFactory.load()): GrpcServerConf = {
-    import net.ceedubs.ficus.Ficus._
-    import net.ceedubs.ficus.readers.ArbitraryTypeReader._
-    config.as[GrpcServerConf](path)
-  }
+  def read[F[_]](config: Config, path: String = ConfigPath)(implicit F: ApplicativeError[F, Throwable]): F[GrpcServerConf] =
```
DieMyst: How about to generalize this method somewhere?

PR #34, opened 2018-02-09
refactor sign classes

DieMyst:
store in private key only S, in public key only Q point without wrappers

### DIFF
--- a/crypto/src/main/scala/fluence/crypto/SignAlgo.scala
+++ b/crypto/src/main/scala/fluence/crypto/SignAlgo.scala
@@ -1,0 +1,28 @@
+package fluence.crypto
+
+import java.security.SecureRandom
+
+import cats.MonadError
+import fluence.crypto.algorithm.{ DumbSign, KeyGenerator, SignatureFunctions }
+import fluence.crypto.keypair.KeyPair
+import fluence.crypto.signature.{ Signature, SignatureChecker, Signer }
+import scodec.bits.ByteVector
+
+import scala.language.higherKinds
+
+class SignAlgo[F[_]](algo: KeyGenerator[F] with SignatureFunctions[F])(implicit F: MonadError[F, Throwable]) {
+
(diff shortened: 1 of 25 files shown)


crypto/src/main/scala/fluence/crypto/SignAlgo.scala
```diff
@@ -0,0 +1,28 @@
+package fluence.crypto
```
alari: Add license header please

crypto/src/main/scala/fluence/crypto/SignAlgo.scala
```diff
@@ -0,0 +1,28 @@
...
+
+import cats.MonadError
+import fluence.crypto.algorithm.{ DumbSign, KeyGenerator, SignatureFunctions }
+import fluence.crypto.keypair.KeyPair
+import fluence.crypto.signature.{ Signature, SignatureChecker, Signer }
+import scodec.bits.ByteVector
+
+import scala.language.higherKinds
+
+class SignAlgo[F[_]](algo: KeyGenerator[F] with SignatureFunctions[F])(implicit F: MonadError[F, Throwable]) {
```
alari: Some docs should be written

crypto/src/main/scala/fluence/crypto/signature/SignatureChecker.scala
```diff
@@ -17,18 +17,10 @@
 
 package fluence.crypto.signature
 
-import cats.MonadError
 import scodec.bits.ByteVector
 
 import scala.language.higherKinds
 
-trait SignatureChecker {
-  def check[F[_]](signature: Signature, plain: ByteVector)(implicit F: MonadError[F, Throwable]): F[Boolean]
```
alari: `MonadError` in method meant that there are three possible results:
- Signature is malformed -- exception in F
- Signature is well formed and matches -- true
- Signature is well formed but doesn't match -- false

With simple F this meaning of possibility of error (hence of algo-specific validation of signature inside the method) is completely lost.

PR #42, opened 2018-02-13
sha1 for js and tests

DieMyst:


crypto/js/src/main/scala/fluence/crypto/facade/Hash.scala
```diff
@@ -30,3 +30,10 @@ class SHA256() extends js.Object {
   def update(msg: js.Array[Byte]): Unit = js.native
   def digest(enc: String): String = js.native
 }
+
+@js.native
+@JSImport("hash.js", "sha1")
+class SHA1() extends js.Object {
```
alari: - Each hasher should be wrapped with CryptoHasher trait.
- Is there JVM implementation for sha1 hasher?
- Is there any platform-agnostic way to get hasher implementation for a particular algorithm?

PR #43, opened 2018-02-14
Comments for BasicContract grpc DTO

alari:


dataset/grpc/src/main/scala/fluence/dataset/grpc/BasicContractCodec.scala
```diff
@@ -81,6 +83,9 @@ object BasicContractCodec {
           def read[T](name: String, f: BasicContract ⇒ T): F[T] =
             Option(f(g)).fold[F[T]](F.raiseError(new IllegalArgumentException(s"Required field not found: $name")))(F.pure)
 
+          def readFromOpt[T](name: String, f: BasicContract ⇒ Option[T]): F[T] =
+            Option(f(g)).flatten.fold[F[T]](F.raiseError(new IllegalArgumentException(s"Required field not found: $name")))(F.pure)
```
DieMyst: You don't need to wrap f(g) with Option here

dataset/grpc/src/main/scala/fluence/dataset/grpc/BasicContractCodec.scala
```diff
@@ -81,6 +83,9 @@ object BasicContractCodec {
           def read[T](name: String, f: BasicContract ⇒ T): F[T] =
             Option(f(g)).fold[F[T]](F.raiseError(new IllegalArgumentException(s"Required field not found: $name")))(F.pure)
 
+          def readFromOpt[T](name: String, f: BasicContract ⇒ Option[T]): F[T] =
```
DieMyst: This is some sort of 'require' or 'unsafe' method, maybe rename it in this way?

dataset/grpc/src/main/scala/fluence/dataset/grpc/BasicContractCodec.scala
alari: Are you completely sure? GRPC code is not always null safe

dataset/grpc/src/main/scala/fluence/dataset/grpc/BasicContractCodec.scala
alari: Why is it unsafe?

DieMyst: It is like option.getOrElse(throw Exception) but under F, as i see

alari: It reads into `F`. I think it's ok that `F` uses `raiseError` effect. We could improve it with `Either` later, but now, having `MonadError` in scope, what's the problem?

PR #44, opened 2018-02-14
new module for monix implementation of KademliaClient

DieMyst:


node/src/main/scala/fluence/node/NodeComposer.scala
```diff
@@ -92,6 +92,13 @@ class NodeComposer(
   private lazy val grpcClientConf =
     GrpcClientConf.read[Task](config).memoizeOnSuccess
 
+  private def readConfig[F[_]](config: Config, path: String = "fluence.network.kademlia")(implicit F: ApplicativeError[F, Throwable]): F[KademliaConf] =
```
alari: it should be named "readKademliaConfig" or smth similar

node/src/main/scala/fluence/node/NodeComposer.scala
```diff
@@ -115,7 +122,7 @@ class NodeComposer(
 
       override val signer: Signer = algo.signer(keyPair)
 
-      override lazy val kademlia: Kademlia[Task, Contact] = new KademliaService(
+      override lazy val kademlia: Kademlia[Task, Contact] = new KademliaMVar(
```
alari: Are you sure about `new` here?

node/src/test/scala/fluence/node/NetworkSimulationSpec.scala
```diff
@@ -73,7 +74,7 @@ class NetworkSimulationSpec extends WordSpec with Matchers with ScalaFutures wit
 
     val kademliaClient = client.service[KademliaClient[Task]] _
 
-    val kad = new KademliaService(
+    val kad = new KademliaMVar(
```
alari: You have `new` here as well.

PR #45, opened 2018-02-14
moveup RocksDb from ContractCach

RicoGit:


node/src/main/scala/fluence/node/NodeComposer.scala
```diff
@@ -55,18 +54,17 @@ class NodeComposer(
     keyPair: KeyPair,
     algo: SignAlgo,
     config: Config,
-    contractsCacheStoreName: String = "fluence_contractsCache"
+    contractCacheStore: KVStore[Task, Array[Byte], Array[Byte]],
+    cryptoHasher: JdkCryptoHasher = JdkCryptoHasher.Sha256
```
alari: Why only `JskCryptoHasher` is suitable?

storage/core/src/main/resources/reference.conf
```diff
@@ -15,4 +15,6 @@ fluence.node.storage.rocksDb {
   # create dataDir if it not exists
   createIfMissing: true
 
-}
\ No newline at end of file
+}
+
+# TODO remove 'node' part from all properties
```
alari: Why?

node/src/main/scala/fluence/node/ContractsCacheStore.scala
```diff
@@ -147,7 +145,16 @@ object ContractsCacheStore {
...
+  /**
+   * Creates [[KVStore]] for caching contracts. Wraps 'binary store' with key and value codecs.
+   *
+   * @param contractCacheBinaryStore Task based key/value store for binary data
+   * @param F An Effect
+   * @return contract cache key/value Store
+   */
+  def apply[F[_]](
+    contractCacheBinaryStore: KVStore[Task, Array[Byte], Array[Byte]]
+  )(implicit F: MonadError[F, Throwable]): F[KVStore[Task, Key, ContractRecord[BasicContract]]] = {
```
alari: Why do you need to use `F` here? Previously it was used to catch errors during RocksDBStore creation, but now it's useless

node/src/test/scala/fluence/node/ClientNodeIntegrationSpec.scala
```diff
@@ -32,21 +31,23 @@ import fluence.crypto.hash.TestCryptoHasher
...
 import monix.eval.Task
-import monix.eval.instances.CatsEffectForTask
 import monix.execution.Scheduler.Implicits.global
 import monix.execution.{ CancelableFuture, Scheduler }
-import org.rocksdb.RocksDBException
 import org.scalatest.concurrent.ScalaFutures
 import org.scalatest.time.{ Milliseconds, Seconds, Span }
 import org.scalatest.{ BeforeAndAfterAll, Matchers, WordSpec }
 import scodec.bits.ByteVector
+import shapeless.{ HList, HNil, ops, _ }
```
alari: `import shapeless._` is enough

storage/core/src/main/resources/reference.conf
RicoGit: A storage may be needed on the client.

node/src/main/scala/fluence/node/ContractsCacheStore.scala
RicoGit: Ok

node/src/main/scala/fluence/node/NodeComposer.scala
RicoGit: did not pay attention

storage/core/src/main/resources/reference.conf
alari: It worth adding the reason to comment, so that other developers may understand why to do it.

node/src/test/scala/fluence/node/ClientNodeIntegrationSpec.scala
RicoGit: Ok

PR #53, opened 2018-02-16
Node app flow

alari:


node/src/main/scala/fluence/node/FluenceNode.scala
```diff
@@ -0,0 +1,183 @@
...
+   * Initiates a directory with all its parents
+   *
+   * @param path Directory path to create
+   * @return Existing directory
+   */
+  private def initDirectory(path: String): IO[File] =
+    IO {
+      val appDir = new File(path)
+      if (!appDir.exists()) {
+        appDir.getParentFile.mkdirs()
```
RicoGit: It's redundant, mkdirs() creates folder including any
     necessary but nonexistent parent directories.

node/src/main/scala/fluence/node/NodeGrpc.scala
```diff
@@ -0,0 +1,111 @@
...
+  // TODO: remove it
+  private implicit def runId[F[_]]: F ~> F = new (F ~> F) {
+    override def apply[A](fa: F[A]): F[A] = fa
+  }
+
+  def grpcContact(signer: Signer, serverBuilder: GrpcServer.Builder): IO[Contact] =
+    Contact.buildOwn[IO](
+      serverBuilder.address,
+      serverBuilder.port,
+      0l, // todo protocol version
```
RicoGit: Looks like 01, It will be better use 0L

node/src/main/scala/fluence/node/NodeGrpc.scala
```diff
@@ -0,0 +1,111 @@
...
+    serverBuilder: GrpcServer.Builder,
+    config: Config
+  ): IO[GrpcServer] =
+    for {
+      clientConf ← GrpcClientConf.read[IO](config)
+    } yield {
+      import services._
+
+      val _signAlgo = services.signAlgo
+      import _signAlgo.checker
```
RicoGit: `import services.signAlgo.checker ` May be better?

storage/rocksdb/src/main/scala/fluence/storage/rocksdb/RocksDbStore.scala
```diff
@@ -122,32 +124,53 @@ object RocksDbStore {
...
 
-  def apply[F[_]](dataSet: String, config: RocksDbConf)(implicit F: ApplicativeError[F, Throwable]): F[RocksDbStore] = {
-    val dbRoot = s"${config.dataDir}/$dataSet"
-    val options = createOptionsFromConfig(config, Path(dbRoot))
+    /**
+     * Closes all launched instances of RocksDB
+     */
+    def close: IO[Unit] = IO {
+      logger.info(s"Closing RocksDB instances: ${instances.keys.mkString(", ")}")
+      instances.keySet.flatMap(ds ⇒ instances.remove(ds)).foreach(_.close())
```
RicoGit: If some there will be  an exception, all instances after won't be closed? Is there no test for that?

alari: Yep, exactly.

node/src/main/scala/fluence/node/NodeGrpc.scala
alari: It's definitely better. However, it's not possible

PR #57, opened 2018-02-17
Restart client testcase

RicoGit:


node/src/test/scala/fluence/node/ClientNodeIntegrationSpec.scala
```diff
@@ -89,10 +92,10 @@ class ClientNodeIntegrationSpec extends WordSpec with Matchers with ScalaFutures
     "get an exception" when {
 
       "port is already used" in {
-        val server = new ServerSocket(3000)
+        val port = 6111
```
alari: Ports with numbers starting with ~3200 are noticed to produce bugs in Travis.

node/src/test/scala/fluence/node/ClientNodeIntegrationSpec.scala
```diff
@@ -347,7 +471,7 @@ class ClientNodeIntegrationSpec extends WordSpec with Matchers with ScalaFutures
   private def runNodes(action: Map[Contact, NodeComposer] ⇒ Unit, numberOfNodes: Int = 20): Unit = {
 
     val servers = (0 to numberOfNodes).map { n ⇒
-      val port = 3100 + n
+      val port = 6112 + n
```
alari: It's very likely to fail in Travis

RicoGit: ok, will return 3100

PR #59, opened 2018-02-20
Client cli

DieMyst:


client/src/main/scala/fluence/client/ClientApp.scala
```diff
@@ -17,8 +17,115 @@
...
+      case Some(Get(k)) ⇒
+        val t = for {
+          ds ← fluenceClient.getOrCreateDataset(ac)
+          res ← ds.get(k)
+        } yield res
+
+        val res = Await.result(t.runAsync, 5.seconds) match {
+          case Some(r) ⇒ r
+          case None    ⇒ "None"
+        }
```
RicoGit: Is this really needed?

client/src/main/scala/fluence/client/CommandParser.scala
```diff
@@ -0,0 +1,33 @@
+package fluence.client
+
+import fastparse.all._
+
+object CommandParser {
+  val escape = P("\\" ~ CharIn("\"/\\bfnrt"))
+  val space = P(CharsWhileIn(" \r\n").?)
```
RicoGit: May be several of this field should be private? Is this parser thread save, may be should make it class

client/src/main/scala/fluence/client/ClientApp.scala
DieMyst: It is better to show the real result instead of the "Some("result")"

client/src/main/scala/fluence/client/CommandParser.scala
DieMyst: yes, this parser is thread safe

client/src/main/scala/fluence/client/Operation.scala
```diff
@@ -0,0 +1,23 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package fluence.client
+
+sealed trait Operation
```
alari: Probably it's a perfect job for `Free`

PR #60, opened 2018-02-20
Cli improvements

DieMyst:


client/src/main/scala/fluence/client/ClientApp.scala
```diff
@@ -50,6 +49,10 @@ object ClientApp extends App with slogging.LazyLogging {
   val algo = Ecdsa.signAlgo
   import algo.checker
 
+  //for terminal improvements: history, navigation
+  val terminal = TerminalBuilder.terminal()
```
alari: What happens if client app is runned in non-interactive mode? e.g. inside docker, or via supervisorctl? Could these lines throw exceptions or block in some cases?

client/src/main/scala/fluence/client/ClientApp.scala
```diff
@@ -58,12 +61,12 @@ object ClientApp extends App with slogging.LazyLogging {
         case None             ⇒ ConfigFactory.load()
       }
 
-      val seedsB64 = if (c.seed.isEmpty) config.getStringList("fluence.seed").asScala else c.seed
+      val seedsB64 = if (c.seed.isEmpty) config.getStringList("fluence.client.seed").asScala else c.seed
```
alari: Seed nodes are shared between client and server. No need to put seeds into client or server.

client/src/main/scala/fluence/client/ClientApp.scala
```diff
@@ -72,63 +75,61 @@ object ClientApp extends App with slogging.LazyLogging {
...
 
-  /**
-   * @param line Input string from user
-   * @param fluenceClient Ready to work client
-   * @param ac Keypair of user
-   */
-  def handleCommands(line: String, fluenceClient: FluenceClient, ac: AuthorizedClient): Unit = {
-    val commandOp = CommandParser.parseCommand(line)
-    commandOp match {
+  def handleCmds(fluenceClient: FluenceClient, ac: AuthorizedClient): Task[Unit] = {
```
alari: What is the shared resource here? Why Task is used, not IO?

client/src/main/scala/fluence/client/ClientApp.scala
```diff
@@ -72,63 +75,61 @@ object ClientApp extends App with slogging.LazyLogging {
...
-
       case Some(Get(k)) ⇒
-        val t = for {
+        for {
           ds ← fluenceClient.getOrCreateDataset(ac)
           res ← ds.get(k)
-        } yield res
+          printRes = res match {
+            case Some(r) ⇒ r
+            case None    ⇒ "None"
```
alari: Probably it should be something more specific. "None" looks exactly like a value.

client/src/main/scala/fluence/client/ClientApp.scala
```diff
@@ -72,63 +75,61 @@ object ClientApp extends App with slogging.LazyLogging {
...
+          }
+          _ = logger.info("Result: \"" + printRes + "\"")
+          _ ← handle
+        } yield ()
+      case _ ⇒
+        Task.pure(logger.info("Wrong command.")).flatMap(_ ⇒ handle)
```
alari: `.pure` is eager, while logging is an effect. You should use smth like `suspend` -- e.g. `IO(logger...)` or `Task(logger...)`

client/src/main/scala/fluence/client/ClientApp.scala
```diff
@@ -72,63 +75,61 @@ object ClientApp extends App with slogging.LazyLogging {
...
+    }.asyncBoundary
```
alari: Why not to `shift` here? Why do you need an async boundary?

client/src/main/scala/fluence/client/ClientApp.scala
DieMyst: I don't no for sure. We should write some bash script for connecting to cli process for this cases.

client/src/main/scala/fluence/client/ClientApp.scala
DieMyst: something like <null> without quotas

PR #61, opened 2018-02-21
Config refactoring

alari:


client/src/main/resources/reference.conf
```diff
@@ -1,15 +1,8 @@
 fluence {
-  client {
-    //  seed = []
-    //  keystore = ""
```
DieMyst: there is no class for client config eventually.

co-fail/src/main/scala/fluence/cofail/cofails.scala
```diff
@@ -0,0 +1,76 @@
...
+import scala.language.higherKinds
+import scala.util.control.NonFatal
+
+/**
+ * This is a common supertype for all application-specific errors.
+ * We consider subtyped errors as managed by us, and all others as exceptions.
+ *
+ * @param message Human-readable message
+ */
+private[cofail] sealed abstract class CoFail(val message: String, val errorType: String)
```
RicoGit: Why this class named 'CoFail', not a just Fail? What knowledge does 'Co' bring to us?

co-fail/src/main/scala/fluence/cofail/cofails.scala
```diff
@@ -0,0 +1,76 @@
...
+}
+
+/**
+ * Network-related errors
+ * @param message Human-readable message
+ */
+sealed class NetworkErr(message: String) extends CoFail(message, "network")
+
+object NetworkErr {
+  def unavailable(message: String): NetworkErr = new NetworkErr(message)
```
RicoGit: How should we use this method? It looks like apply.

co-fail/src/main/scala/fluence/cofail/cofails.scala
```diff
@@ -0,0 +1,76 @@
...
+ * @param message Human-readable message
+ */
+sealed class CryptoErr(message: String) extends CoFail(message, "crypto")
+
+object CryptoErr {
+  def apply(message: String): CryptoErr = new CryptoErr(message)
+
+  def apply(reason: Throwable): CryptoErr = new CryptoErr(reason.getMessage)
+
+  def catchNonFatal[F[_] : Applicative, A](errorText: ⇒ String, a: ⇒ A): EitherT[F, CryptoErr, A] =
```
RicoGit: `a: ⇒ A`  what it is actually is?

node/src/main/resources/reference.conf
```diff
@@ -1,22 +1,48 @@
...
+    }
+  }
 
+  // TODO
   contract.cacheDirName = contract_cache
 
-  gitHash = "00000000000000000000" // twenty zeros
-  gitHash = ${?FLUENCE_GIT_HASH}
-}
+}
```
RicoGit: Is an application should be application.conf instead reference.conf. reference.conf used for libraries.

node/src/main/scala/fluence/node/NodeComposer.scala
```diff
@@ -59,7 +59,7 @@ object NodeComposer {
       k ← Key.fromKeyPair[IO](keyPair)
       kadConf ← KademliaConfigParser.readKademliaConfig[IO](config)
       rocksDbFactory = new RocksDbStore.Factory
-      contractCacheStore ← rocksDbFactory[IO](config.getString("fluence.contract.cacheDirName"))
+      contractCacheStore ← rocksDbFactory[IO](config.getString("fluence.contract.cacheDirName"), config)
```
RicoGit: "fluence.contract.cacheDirName" should be inside of ContractsCacheStore() and rocksDbFactory too.
RocksDb should build inside ContractsCacheStore(). What do you think?

node/src/test/scala/fluence/node/ClientNodeIntegrationSpec.scala
```diff
@@ -466,8 +464,8 @@ class ClientNodeIntegrationSpec extends WordSpec with Matchers with ScalaFutures
         s.stop.unsafeRunSync()
 
         // clean all rockDb data from disk
-        if (s.config.getString("fluence.node.storage.rocksDb.dataDir").startsWith(System.getProperty("java.io.tmpdir")))
-          Path(s.config.getString("fluence.node.storage.rocksDb.dataDir")).deleteRecursively()
+        if (s.config.getString("fluence.storage.rocksDb.dataDir").startsWith(System.getProperty("java.io.tmpdir")))
```
RicoGit: Is there enough to clean only "fluence.directory" directory for deleting all data?

node/src/test/scala/fluence/node/ClientNodeIntegrationSpec.scala
```diff
@@ -510,7 +508,7 @@ class ClientNodeIntegrationSpec extends WordSpec with Matchers with ScalaFutures
 
   override protected def afterAll(): Unit = {
     super.afterAll()
-    Path(config.getString("fluence.node.storage.rocksDb.dataDir")).deleteRecursively()
+    Path(config.getString("fluence.storage.rocksDb.dataDir")).deleteRecursively()
```
RicoGit: Is there enough to clean only "fluence.directory" directory for deleting all data?

project/SbtCommons.scala
```diff
@@ -47,14 +47,16 @@ object SbtCommons {
   val AirflameV = "0.38"
   val ScalatestV = "3.0.+"
   val SloggingV = "0.6.1"
+  val ShapelessV = "2.3.+"
```
RicoGit: Do you really think it's good practice to use "2.3.+" instead fixed version?

client/src/main/resources/reference.conf
alari: It's not necessary, as both seed and keystore are defined in other configs.

co-fail/src/main/scala/fluence/cofail/cofails.scala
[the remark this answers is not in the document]
alari: It's going to be used for coproducts of fails.

co-fail/src/main/scala/fluence/cofail/cofails.scala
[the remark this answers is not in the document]
alari: It highlights an idea of having custom builders for custom types of errors. As it's not used anywhere (yet), it looks exactly the same as `apply`

node/src/main/scala/fluence/node/FluenceNode.scala
```diff
@@ -96,24 +98,22 @@ object FluenceNode extends slogging.LazyLogging {
...
-   * Reads seed nodes contacts from config
-   */
-  def readSeedsConfig(conf: Config): IO[SeedsConfig] =
-    IO {
-      import net.ceedubs.ficus.Ficus._
-      import net.ceedubs.ficus.readers.ArbitraryTypeReader._
-      conf.as[SeedsConfig]("fluence.node.join")
+  private def launchUPnP(config: Config, contactConf: ContactConf, grpc: GrpcServerConf): IO[(ContactConf, IO[Unit])] =
+    UPnPConf.read(config).flatMap {
+      case u if !u.isEnabled ⇒ IO.pure(contactConf -> IO.unit)
```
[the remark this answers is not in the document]
alari: It's a matter of taste. `UPnPConf.read` reads the conf. `u` in a map is conf object. If it was a `.filter(_.isEnabled)`, it won't even need any name. For partial function, name is necessary, but as it lives in less then one line (only in a guard clause), and just checks that conf is not enabled, I believe `u` is good enough.

co-fail/src/main/scala/fluence/cofail/cofails.scala
alari: I have already seen this message somewhere above :)

node/src/main/resources/reference.conf
alari: application.conf should be used _only_ by the end user of the library or application, otherwise there could be problems. End user should be able to provide his own `application.conf` with the -Dconfig.file directive. If you already have `application.conf` in any of your modules, it becomes tough and buggy and requires to overwrite all the `application.conf` directives in it.

node/src/main/scala/fluence/node/NodeComposer.scala
alari: - RocksDB should never know how is it used.
- This particular config line should be moved somewhere later.

node/src/test/scala/fluence/node/ClientNodeIntegrationSpec.scala
alari: It should be. You may improve it with a commit or PR.

project/SbtCommons.scala
alari: Sometimes it's very handy. And there's no change here, version is the same.

PR #63, opened 2018-02-22
Client side replication

RicoGit:


client/src/main/scala/fluence/client/ClientReplicationWrapper.scala
```diff
@@ -0,0 +1,67 @@
...
+import monix.eval.Task
+
+import scala.language.higherKinds
+
+/**
+ * Wrapper with dataset replication. Allows writes data to all nodes and reads data if at least one node is alive.
+ * This is naive implementation of replication and it will be removed in future.
+ */
+class ClientReplicationWrapper[K, V](
+    datasetReplicas: List[(ClientDatasetStorageApi[Task, K, V], Contact)]
```
alari: Probably `Contact` should be at the left side, and it even might be a map.

client/src/main/scala/fluence/client/ClientReplicationWrapper.scala
```diff
@@ -0,0 +1,67 @@
...
+   * @param value The value associated with the specified key
+   * @return returns old value if old value was overridden, None otherwise.
+   */
+  override def put(key: K, value: V): Task[Option[V]] = {
+    Task.gatherUnordered(
+      datasetReplicas
+        .map { case (store, _) ⇒ store.put(key, value) }
+    ).map { seq ⇒
+        logger.info(s"$key and $value was written to $replicationFactor nodes")
+        seq.head
```
alari: Is it safe?

RicoGit: yes

client/src/main/scala/fluence/client/ClientReplicationWrapper.scala
RicoGit: Contract is for nice logging only, there is no needed Map

client/src/main/scala/fluence/client/ClientReplicationWrapper.scala
alari: Could you please add comment about why is it safe

client/src/main/scala/fluence/client/ClientReplicationWrapper.scala
alari: Then why not to document that?

client/src/main/resources/reference.conf
```diff
@@ -1,8 +1,18 @@
...
+  kademlia {
+    maxBucketSize: 16
+    maxSiblingsSize: 32
+    parallelism: 3
+    pingExpiresIn: 1 hour
+  }
+
+  network {
+    join {
+      seeds: [
```
alari: seeds are in `fluence.network.seeds` now

client/src/main/scala/fluence/client/FluenceClient.scala
```diff
@@ -147,6 +155,58 @@ class FluenceClient(
...
+      key ← Key.fromKeyPair(ac.kp)
+      bcOp ← contracts.find(key).attempt.map(_.toOption)
+      dataStorages ← bcOp match {
+
+        case Some(basicContract) ⇒ //create storage from existed contract
+          for {
+            nodes ← findContactsOfAllParticipants(basicContract)
+            datasets ← Task.sequence(
+              nodes.map( contact ⇒
+                addNonEncryptedDataset(ac, contact, Some(ClientState(basicContract.executionState.merkleRoot.toArray)))
```
alari: We should move to `ByteVector` for merkle roots everywhere

RicoGit: Ok

client/src/main/resources/reference.conf
RicoGit: Without this client isn't starts. Are you sure?

client/src/main/scala/fluence/client/ClientReplicationWrapper.scala
RicoGit: Cause it not necessary )

PR #73, opened 2018-03-05
Kad proto cross

alari:


crypto/jvm/src/main/scala/fluence/crypto/hash/CryptoHashers.scala
```diff
@@ -0,0 +1,24 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package fluence.crypto.hash
+
+object CryptoHashers {
```
RicoGit: Why CryptoHashers appears twice?

kademlia/protocol/src/main/scala/fluence/kad/protocol/Contact.scala
```diff
@@ -123,21 +107,17 @@ object Contact {
...
-      "ip" -> Json.fromString(data.ip.getHostAddress),
+      "a" -> Json.fromString(data.addr),
       "gp" -> Json.fromInt(data.grpcPort),
       "gh" -> Json.fromString(data.gitHash)
     )
 
     implicit val decodeData: Decoder[JwtData] = c ⇒
       for {
-        ip ← c.downField("ip").as[String]
+        ip ← c.downField("a").as[String]
```
RicoGit: var named 'ip', but it isn't actually 'ip', right?

kademlia/protocol/src/main/scala/fluence/kad/protocol/Contact.scala
```diff
@@ -123,21 +107,17 @@ object Contact {
...
         p ← c.downField("gp").as[Int]
         gh ← c.downField("gh").as[String]
-        ipAddr ← Try(InetAddress.getByName(ip)).fold(
-          t ⇒ Left(DecodingFailure.fromThrowable(t, Nil)),
-          i ⇒ Right(i)
-        )
-      } yield JwtData(ip = ipAddr, grpcPort = p, gitHash = gh)
+      } yield JwtData(addr = ip, grpcPort = p, gitHash = gh)
```
RicoGit: ip -> address

crypto/jvm/src/main/scala/fluence/crypto/hash/CryptoHashers.scala
alari: One instance is for js, another one for jvm. Both share the same duck type.

PR #75, opened 2018-03-06
kademlia-core and kademlia-monix cross builds

alari:


client/src/main/scala/fluence/client/FluenceClient.scala
```diff
@@ -112,9 +109,10 @@ class FluenceClient(
    *
    * Use string only info now
    * private until multiple datasets per authorized client is supported
-   * @param storageRpc transport to dataset
+   *
+   * @param storageRpc  transport to dataset
```
RicoGit: transport -> Transport

kademlia/protocol/src/main/scala/fluence/kad/protocol/Key.scala
```diff
@@ -38,22 +37,18 @@ import scala.util.Try
  *
  * @param value ID wrapped with ByteVector
  */
-final case class Key private (value: ByteVector) extends AnyVal {
-  def id: Array[Byte] = value.toArray
+final case class Key private (value: ByteVector) {
+  lazy val id: Array[Byte] = value.toArray
+
+  lazy val bits: BitVector = value.toBitVector.padLeft(Key.BitLength)
```
RicoGit: BitVector here is perfectly fitted! +1

PR #77, opened 2018-03-07
Btree refactoring for logging

RicoGit:


b-tree/client/src/test/scala/fluence/btree/client/MerkleBTreeClientSpec.scala
```diff
@@ -35,6 +35,15 @@ class MerkleBTreeClientSpec extends WordSpec with Matchers with ScalaFutures {
     def toKey: Key = Key(str.getBytes)
   }
 
+  private implicit class Str2Hash(str: String) {
+    def toHash: Hash = Hash(str.getBytes)
+  }
+
+  private val testHasher = new CryptoHasher[Array[Byte], Hash] {
```
alari: At some moment we probably should add an effect to hasher, and make smth like `.map` and/or `.transform(Codec)` for it, what do you think?

b-tree/client/src/main/scala/fluence/btree/client/BTreeVerifier.scala
```diff
@@ -37,11 +32,11 @@ import scala.collection.Searching.{ Found, InsertionPoint }
  * @param merkleRootCalculator Merkle proof service that allows calculate merkle root from merkle path
  */
 class BTreeVerifier(
-    cryptoHasher: CryptoHasher[Array[Byte], Array[Byte]],
+    cryptoHasher: CryptoHasher[Array[Byte], Hash],
     merkleRootCalculator: MerkleRootCalculator
 ) extends slogging.LazyLogging {
 
-  import BTreeVerifier._
+  import Hash._
```
alari: Why do you need this import?

dataset/node/src/main/scala/fluence/dataset/node/storage/DatasetNodeStorage.scala
```diff
@@ -140,15 +141,20 @@ object DatasetNodeStorage {
       ByteBuffer.allocate(java.lang.Long.BYTES).putLong(_).array()
     )
 
+    val wrappedHasher = new CryptoHasher[Array[Byte], Hash] {
```
alari: All these wrappedHashers are copy-paste. It might worth to put its builder in some single place, or derive it from codec or .map, or do smth similar.

b-tree/client/src/main/scala/fluence/btree/client/BTreeVerifier.scala
alari: I suppose you need it for implicit `Array[Hash]`class? As other implicits are imported from Hash companion object automatically

dataset/node/src/main/scala/fluence/dataset/node/storage/DatasetNodeStorage.scala
RicoGit: It's needed for testing only and in different modules.  For testing copy and paste is ok.

PR #86, opened 2018-03-16
Kad eithert transparent errors

alari:


client/cli/jvm/src/main/scala/fluence/client/cli/Cli.scala
```diff
@@ -34,6 +34,8 @@ import scala.language.higherKinds
 
 object Cli extends slogging.LazyLogging {
 
+  type Err = Throwable
```
DieMyst: can somehow unify the names of errors?

kademlia/protocol/src/main/scala/fluence/kad/protocol/KademliaRpc.scala
```diff
@@ -26,26 +28,28 @@ import scala.language.higherKinds
...
    * Perform a local lookup for a key, return K closest known nodes, going away from the second key.
    *
    * @param key Key to lookup
    */
-  def lookupAway(key: Key, moveAwayFrom: Key, numberOfNodes: Int): F[Seq[Node[C]]]
+  def lookupAway(key: Key, moveAwayFrom: Key, numberOfNodes: Int): EitherT[F, Error, Seq[Node[C]]]
+}
+
+object KademliaRpc {
+  type Aux[F[_], C, E] = KademliaRpc[F, C] { type Error = E }
```
DieMyst: can we avoid using Aux here?

contract/client/src/main/scala/fluence/contract/client/Contracts.scala
```diff
@@ -65,15 +66,15 @@ class Contracts[F[_], Contract : ContractRead : ContractWrite, Contact](
...
-          ME.ensure(c.participantSigned(nc.key))(Contracts.NotFound)(identity) map { _ ⇒ c }
-        },
+        nc ⇒ EitherT(allocatorRpc(nc.contact).offer(contract).flatMap { c ⇒
+          c.participantSigned[F](nc.key).map(Either.cond(_, c, Contracts.NotFound))
+        }),
         contract.participantsRequired,
         maxAllocateRequests(contract.participantsRequired),
         isIdempotentFn = false
-      ).flatMap {
+      ).value.map(_.right.toSeq.flatten).flatMap { // TODO propagate errors
```
DieMyst: so many .value.map(_.right...
maybe we can do something with this?

kademlia/core/src/main/scala/fluence/kad/RoutingTable.scala
```diff
@@ -480,48 +491,53 @@ object RoutingTable {
      * @param numberOfNodes How many nodes to lookupIterative for each peer
      * @param checkNode Test node correctness, e.g. signatures are correct, ip is public, etc.
      * @param parallelism Parallelism factor to perform self-[[lookupIterative()]] in case of successful join
-     * @return F[Unit], possibly a failure if were not able to join any node
+     * @return Either a non-empty list of errors, or the discovered node's neighborhood (up to numberOfNodes)
      */
-    def join(
+    def join[E](
```
DieMyst: i see chain of flatMaps in this method, maybe rewrite it with for-comprehension, so we can easier simplify method in future

contract/client/src/main/scala/fluence/contract/client/Contracts.scala
alari: Yep, that's why there's TODO here

alari:
Seems like not a good way to go

PR #94, opened 2018-03-21
Contract io

alari:


client/core/src/main/scala/fluence/client/core/FluenceClient.scala
```diff
@@ -181,7 +195,7 @@ class FluenceClient(
               )
             )
           } yield Some(datasets)
-        case None ⇒ Task.pure(None)
+        case _ ⇒ Task.pure(None)
```
DieMyst: should we handle errors here?

contract/client/src/main/scala/fluence/contract/client/Contracts.scala
```diff
@@ -82,98 +84,113 @@ object Contracts {
...
-        .flatMap {
-          case sq if sq.nonEmpty ⇒
-            // Contract is found; for the case several different versions are returned, find the most recent
-            sq.map(_._2).maxBy(_.version).pure[F]
-
-          case _ ⇒
-            // Not found -- can't do anything
-            ME.raiseError(Contracts.NotFound)
-        }
+      EitherT(
```
DieMyst: why is this wrapper here? callIterative will return EitherT in future?

contract/node/src/main/scala/fluence/contract/node/ContractAllocator.scala
```diff
@@ -65,81 +67,94 @@ class ContractAllocator[F[_], C: ContractRead: ContractWrite](
...
+
+      contractOpt ← toIO(storage.get(contract.id)).attempt.map(_.toOption)
+
+      contract ← contractOpt match {
+        case Some(cr) ⇒
+          eitherFail(cr.contract.isBlankOffer(), "Different contract is already stored for this ID").flatMap { _ ⇒
+            if (cr.contract =!= contract) {
+              // contract preallocated for id, but it's changed now
+              for {
+                _ ← toIO(storage.remove(contract.id))
```
DieMyst: move this chain to method, and then do toIO(...)

contract/client/src/main/scala/fluence/contract/client/Contracts.scala
alari: `EitherT.apply` converts `F[Either[A, B]]` to `EitherT[F, A, B]` -- that's what we want to get

`callIterative` will return errors collected from the nodes in case of failure.

contract/core/src/main/scala/fluence/contract/ops/ContractRead.scala
```diff
@@ -165,39 +163,45 @@ object ContractRead {
...
+        type M[A] = EitherT[F, CryptoErr, A]
+        Monad[M].tailRecM(participants.toStream) {
+          case pk #:: tail ⇒
+            participantSigned[F](pk).map[Either[Stream[Key], Boolean]] {
+              case true ⇒ Left(tail)
+              case false ⇒ Right(false)
+            }
+          case _ ⇒ Monad[M].pure(Right[Stream[Key], Boolean](true))
+        }
+      } else EitherT.rightT(false)
```
RicoGit: Contract-core module isn't have any test, this is so sad.

PR #97, opened 2018-03-27
Dataset client signature (part 1)

RicoGit:


contract/core/src/main/scala/fluence/contract/ops/ContractRead.scala
```diff
@@ -106,35 +124,32 @@ object ContractRead {
...
+      } yield ()
 
     /**
      * Checks that signature matches contract's offer
      *
      * @param signature Signature to check
      * @param checker Signature checker
      */
-    def checkOfferSignature[F[_]: Monad](
+    private def checkOfferSignature[F[_]: Monad](
```
alari: What is the reason for making it private?

contract/core/src/main/scala/fluence/contract/ops/ContractRead.scala
```diff
@@ -143,64 +158,124 @@ object ContractRead {
...
-        allParticipants ← checkAllParticipants()
-        participantSealResult ← participantsSeal.filter(seal ⇒ Key.checkPublicKey(id, seal.publicKey)) match {
-          case Some(sign) ⇒ checker.check[F](sign, getParticipantsBytes).map(_ ⇒ true)
-          case None ⇒ EitherT.rightT[F, CryptoErr](false)
+        _ ← checkOfferSeal()
+        _ ← checkExecStateSeal()
+        _ ← checkAllParticipants()
+        participantSealResult ← checkParticipantsSeal
+      } yield
+        participantSealResult match {
```
alari: It's the same as `.isDefined`

contract/core/src/main/scala/fluence/contract/ops/ContractRead.scala
```diff
@@ -143,64 +158,124 @@ object ContractRead {
      * @return Whether this contract is a valid blank offer (with no participants, with client's signature)
      */
     def isBlankOffer[F[_]: Monad]()(implicit checker: SignatureChecker): EitherT[F, CryptoErr, Boolean] =
-      if (participants.isEmpty) checkOfferSeal() else EitherT.rightT(false)
-
-    /**
-     * @return Whether this contract offer was signed by a single node and client, but participants list is not sealed yet
-     */
-    def isSignedParticipant[F[_]: Monad]()(implicit checker: SignatureChecker): EitherT[F, CryptoErr, Boolean] =
```
alari: Where has it gone? Can't find out from the PR what is used instead for gathering contract offer signatures on client side

build.sbt
```diff
@@ -440,7 +440,12 @@ lazy val `contract-core` = crossProject(JVMPlatform, JSPlatform)
   .withoutSuffixFor(JVMPlatform)
   .crossType(FluenceCrossType)
   .in(file("contract/core"))
-  .settings(commons)
+  .settings(
+    commons,
+    libraryDependencies ++= Seq(
+      scalatest
```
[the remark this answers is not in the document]
RicoGit: it is, unfortunately only jvm tests can be runned from idea with debugger. I moved tests to crosplatform module.

contract/core/src/main/scala/fluence/contract/ops/ContractRead.scala
RicoGit: This methods isn't used anywhere, why it should be public? This functionality isn't needed outside of contract, for all cases similar there is checkOfferSeal method

contract/core/src/main/scala/fluence/contract/ops/ContractRead.scala
RicoGit: sure, thanks

contract/core/src/main/scala/fluence/contract/ops/ContractRead.scala
RicoGit: This methods wasn't used anywhere, and it seemed to me useless. participantSigned methods is used  instead

PR #98, opened 2018-03-27
Kad contact

alari:


kademlia/grpc/jvm/src/main/scala/fluence/kad/grpc/KademliaGrpcUpdate.scala
```diff
@@ -0,0 +1,57 @@
...
+    update: fluence.kad.protocol.Node[Contact] ⇒ IO[Unit],
+    clientConf: GrpcConf
+  )(implicit checker: SignatureChecker): (Kleisli[Option, String, String], Option[Any]) ⇒ IO[Unit] = {
+    (headers, message) ⇒
+      val remoteNode = for {
+        remoteB64key ← headers(clientConf.keyHeader)
+        remoteKey ← Key.fromB64[Try](remoteB64key).toOption
+
+        remoteSeed ← headers(clientConf.contactHeader)
+        remoteContact ← Contact.readB64seed[Id](remoteSeed).value.toOption
```
DieMyst: it is better to use `try` here, because there may be errors in parsing. Turn it into options it seems wrong.

kademlia/protocol/jvm/src/main/scala/fluence/kad/protocol/ContactSecurity.scala
```diff
@@ -0,0 +1,83 @@
...
+import cats.syntax.eq._
+import cats.syntax.applicative._
+import cats.syntax.apply._
+
+import scala.language.higherKinds
+
+/**
+ * ContactSecurity provides checks that particular Node[Contact] could be saved to a RoutingTable
+ */
+object ContactSecurity {
```
RicoGit: What about test?

kademlia/protocol/js/src/main/scala/fluence/kad/protocol/ContactSecurity.scala
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * ContactSecurity provides checks that particular Node[Contact] could be saved to a RoutingTable
+ * JS version doesn't make use of java.net.InetAddress
+ */
+object ContactSecurity {
+
+  private def checkMaybeLocal[F[_]: Applicative](self: Key): Node[Contact] ⇒ F[Boolean] =
+    node ⇒ Applicative[F].pure(node.key =!= self && Key.checkPublicKey(node.key, node.contact.publicKey))
+
+  def check[F[_]: Applicative](self: Key, acceptLocal: Boolean): Node[Contact] ⇒ F[Boolean] =
```
RicoGit: What about test?

kademlia/grpc/jvm/src/main/scala/fluence/kad/grpc/KademliaGrpcUpdate.scala
alari: Errors are returned on the left side of EitherT, which is simply Either in case of Id monad. As we don't care for error (yet, due to non-EitherT APIs of other parts of for comprehension), turning it into Option to capture the right side is safe.

kademlia/grpc/jvm/src/main/scala/fluence/kad/grpc/KademliaGrpcUpdate.scala
```diff
@@ -0,0 +1,57 @@
...
+object KademliaGrpcUpdate {
+
+  /**
+   * Provides a GRPC server callback to update Kademlia routing table
+   *
+   * @param update Kademlia.update function, converted to IO properly
+   * @param clientConf Grpc Client conf, to get header names from
+   * @return Callback to be registered on GRPC transport
+   */
+  def grpcCallback(
```
RicoGit: What about test?

kademlia/grpc/jvm/src/main/scala/fluence/kad/grpc/KademliaGrpcUpdate.scala
DieMyst: why not use [Option] then?

alari: Because there's no monadic effect at all. And doing `EitherT[Option, A, B].value.flatMap(_.toOption)` to drop Option again is meaningless.

DieMyst: ok, understood

PR #101, opened 2018-03-29
Codecs purify

alari:


client/core/src/main/scala/fluence/client/core/FluenceClient.scala
```diff
@@ -139,7 +139,7 @@ class FluenceClient(
   ): Task[ClientDatasetStorageApi[Task, Observable, String, String]] = {
     import fluence.contract.ops.ContractWrite._
     for {
-      key ← Key.fromKeyPair[Task](keyPair)
+      key ← Key.fromKeyPair.toKleisli[Task].run(keyPair)
```
RicoGit: Why don't you use simple Key.fromBytes.apply[Task](datasetId) ?

codec/core/src/main/scala/fluence/codec/BifuncE.scala
```diff
@@ -0,0 +1,45 @@
...
+
+case class BifuncE[E <: Throwable, A, B](direct: FuncE[E, A, B], inverse: FuncE[E, B, A]) {
+  def swap: BifuncE[E, B, A] = BifuncE(inverse, direct)
+
+  // This is given with Compose, but IDEA fails to perform typecheck due to kind projector
+  def andThen[C](other: BifuncE[E, B, C]): BifuncE[E, A, C] =
+    implicitly[Compose[BifuncE[E, ?, ?]]].andThen(this, other)
+
+  def toCodec[F[_]](implicit F: MonadError[F, Throwable]): Codec[F, A, B] =
+    Codec(direct.toKleisli.run, inverse.toKleisli.run)
```
RicoGit: How I can create codec without MonadError ? For EithertT for example?

codec/core/src/main/scala/fluence/codec/CodecError.scala
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+package fluence.codec
+
+import cats.Eq
+
+import scala.util.control.NoStackTrace
+
+/**
+ * Left side for Codec functions
```
RicoGit: Very strange comment, may be write something about what it is instead where it used.

codec/core/src/main/scala/fluence/codec/FuncE.scala
```diff
@@ -0,0 +1,183 @@
...
+       */
+      override def apply[F[_]: Monad](a: G[A]): EitherT[F, E, G[B]] =
+        Traverse[G].traverse[λ[α ⇒ EitherT[F, E, α]], A, B](a)(funcE.apply)
+    }
+
+  /**
+   * FuncE obeys arrow choice laws with any fixed E
+   * @tparam E Error type
+   */
+  implicit def catsArrowChoice[E <: Throwable]: ArrowChoice[λ[(α, β) ⇒ FuncE[E, α, β]]] = {
```
RicoGit: Can we change unicode symbols to something from a keyboard?

codec/core/src/main/scala/fluence/codec/FuncE.scala
```diff
@@ -0,0 +1,183 @@
...
+    type G[A, B] = FuncE[E, A, B]
+    new ArrowChoice[G] {
+      override def choose[A, B, C, D](f: G[A, C])(g: G[B, D]): G[Either[A, B], Either[C, D]] =
```
RicoGit: ) so many types

codec/core/src/main/scala/fluence/codec/PureCodec.scala
```diff
@@ -0,0 +1,43 @@
...
+ */
+object PureCodec {
+  type Func[A, B] = FuncE[CodecError, A, B]
+
+  def codec[A, B](implicit c: PureCodec[A, B]): PureCodec[A, B] = c
+
+  def apply[A, B](f: Func[A, B], g: Func[B, A]): PureCodec[A, B] =
+    BifuncE(f, g)
+
+  def lift[A, B](f: A ⇒ B, g: B ⇒ A): PureCodec[A, B] =
```
RicoGit: 'pure'? )

codec/core/src/main/scala/fluence/codec/PureCodec.scala
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+package fluence.codec
+
+import scala.language.higherKinds
+
+/**
+ * PureCodec builder functions
+ */
+object PureCodec {
```
RicoGit: Nice to have here an examples for creating 'PureCodec' with descriptions.

codec/core/src/main/scala/fluence/codec/FuncE.scala
```diff
@@ -0,0 +1,183 @@
...
+  def applyS[F[_]: Monad](a: A): EitherT[F, String, B] =
+    apply(a).leftMap(_.getMessage)
+
+  /**
+   * Wraps the function into concrete monad, lifting error side into MonadError effect.
+   * @param F Monad
+   * @tparam F MonadError instance to lift E into Throwable
+   * @return
+   */
+  def toKleisli[F[_]](implicit F: MonadError[F, Throwable]): Kleisli[F, A, B] =
```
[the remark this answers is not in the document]
alari: The common convention is to name the sole typeclass with the name of the type. `ME` doesn't tell you for which type it's defined. In case you notice that you use only applicative behaviour of `MonadError`, you may change the required typeclass seamlessly -- however, `F` name still sounds to it, while `ME` is not. 

Also it's a very rare case when you need to have several typeclasses of this kind defined for a single type. `Applicative` and `FlatMap`? It looks like `Monad`. And so on.

Another case is when you need something like Finally Tagless dependencies, or Free Monad's `Inject`. These cases are not about the way you compose effects, as it is with `Monad` or `Functor`, but about specific ways to describe particular domain-specific ones. Usually special names are used for typeclasses in these cases, like `I` for `Inject`, `P` for `Parallel`, and we use `BW` for `Bucket.WriteOps`. These types come not from category theory, and it's hard to use one domain specific typeclass for another. That's why `F` name is not to be used with them.

Finally, I'd suggest move away from `ME: MonadError`, `A: Applicative` and similar names in favor of the type name.

codec/core/src/main/scala/fluence/codec/FuncE.scala
```diff
@@ -0,0 +1,183 @@
...
+        self.apply(a).leftMap(f)
+    }
+
+}
+
+object FuncE {
+
+  /**
+   * Lift a pure function into FuncE.
+   * @param f The function to lift
```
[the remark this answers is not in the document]
alari: Is it possible to configure scala-fmt to do it?

codec/core/src/main/scala/fluence/codec/FuncE.scala
```diff
@@ -0,0 +1,183 @@
...
+   * @tparam A Input
+   * @tparam E Error type
+   * @tparam B Output
+   */
+  implicit def lift[E <: Throwable, A, B](f: A ⇒ B): FuncE[E, A, B] =
```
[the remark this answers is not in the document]
alari: `pure` moves a value to category. `lift` is used for arrows.

codec/core/src/main/scala/fluence/codec/FuncE.scala
```diff
@@ -0,0 +1,183 @@
...
+    apply(a).leftMap(_.getMessage)
+
+  /**
+   * Wraps the function into concrete monad, lifting error side into MonadError effect.
+   * @param F Monad
+   * @tparam F MonadError instance to lift E into Throwable
+   * @return
+   */
+  def toKleisli[F[_]](implicit F: MonadError[F, Throwable]): Kleisli[F, A, B] =
+    Kleisli(apply(_).value.flatMap(F.fromEither))
```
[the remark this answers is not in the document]
alari: `Kleisli` is intended to be composed. Now we use it in places where we still have `MonadError` way to handle failures. You're right, it worth adding `runF` shortcut instead.

client/core/src/main/scala/fluence/client/core/FluenceClient.scala
alari: It returns `EitherT`, while this for-comprehention is not yet ready for it?

codec/core/src/main/scala/fluence/codec/BifuncE.scala
alari: `EitherT[F: Monad, E <: Throwable, ?]` has `MonadError` instance in Cats.

You must provide `MonadError` to convert `PureCodec` to a `Codec`. This is to force you keep things pure and exceptions obvious for as long as you can.

I hope, we will deprecate `Codec` later.

codec/core/src/main/scala/fluence/codec/FuncE.scala
alari: Nope, it's a kind-projector syntax. But no worry, users of the library have no need to type it, so there should be no problem.

codec/core/src/main/scala/fluence/codec/PureCodecInstances.scala
```diff
@@ -0,0 +1,39 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package fluence.codec
+
+import fluence.codec.PureCodec.{lift, liftEither}
+import scodec.bits.{Bases, ByteVector}
+import scala.language.higherKinds
+
+private[codec] trait PureCodecInstances extends BifuncEInstances {
```
DieMyst: are we need BifuncEInstances here? Better to move it straight in BifuncE

codec/core/src/test/scala/fluence/codec/FuncEInstances.scala
```diff
@@ -0,0 +1,42 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package fluence.codec
+
+import cats.{Eq, Id}
+import cats.laws.discipline.eq._
+import org.scalacheck.{Arbitrary, Cogen, Gen}
+
+object FuncEInstances {
```
DieMyst: it's better to rename it with a reference to test environment: FuncETestInstances or something like that

kademlia/core/src/test/scala/fluence/kad/KeySpec.scala
```diff
@@ -33,13 +32,11 @@ class KeySpec extends WordSpec with Matchers {
...
-          val buffer = ByteBuffer.allocate(java.lang.Long.BYTES)
-          buffer.putLong(i)
-          buffer.array()
-        }))
-        .value
+      Key.fromBytes.unsafe(Array.concat(Array.ofDim[Byte](Key.Length - java.lang.Long.BYTES), {
+        val buffer = ByteBuffer.allocate(java.lang.Long.BYTES)
+        buffer.putLong(i)
+        buffer.array()
+      }))
```
RicoGit: It can be changed to `ByteVector.fromLong(newVer).toArray`.

kademlia/jvm/src/test/scala/fluence/kad/NetworkSimulationSpec.scala
```diff
@@ -125,22 +116,28 @@ class NetworkSimulationSpec extends WordSpec with Matchers with ScalaFutures wit
 
       servers.foreach { s ⇒
         logger.debug(Console.BLUE + s"Join: ${s.nodeId}" + Console.RESET)
-        s.join(Seq(firstContact, secondContact), 6).runAsync.futureValue
-        logger.debug(Console.BLUE + "Joined" + Console.RESET)
```
RicoGit: log was gone, it's ok?

kademlia/testkit/src/test/scala/fluence/kad/KademliaSimulationSpec.scala
```diff
@@ -35,13 +35,11 @@ import scala.util.Random
...
-        val buffer = ByteBuffer.allocate(java.lang.Long.BYTES)
-        buffer.putLong(i)
-        buffer.array()
-      }))
-      .value
+    Key.fromBytes.unsafe(Array.concat(Array.ofDim[Byte](Key.Length - java.lang.Long.BYTES), {
+      val buffer = ByteBuffer.allocate(java.lang.Long.BYTES)
+      buffer.putLong(i)
+      buffer.array()
+    }))
```
RicoGit: `ByteVector.fromLong(newVer).toArray`

PR #106, opened 2018-04-09
Http proxy

DieMyst:


transport/core/src/main/scala/fluence/transport/TransportServer.scala
```diff
@@ -17,5 +17,44 @@
...
+
+  protected def startServer: B ⇒ IO[S]
+  protected def shutdownServer: S ⇒ IO[Unit]
+
+  /**
+   * Launch server, grab ports, or fail
+   */
+  val start: IO[Unit] =
+    for {
+      _ ← if (serverRef.get() == null) IO.unit else shutdown
```
RicoGit: It looks like not thread safe. 2 threads can run 'shutdown' concurrently.
Lets make 'serverRef' abstract methods and each implementation will create its own concurrent code for manage state, method should return IO[S] or Task[S]

transport/core/src/main/scala/fluence/transport/TransportServer.scala
```diff
@@ -17,5 +17,44 @@
...
 
-// TODO: expose lifecycle events, allow subscribing on them
-trait TransportServer
+import java.util.concurrent.atomic.AtomicReference
+
+import cats.effect.IO
+
+trait TransportServer[B, S] extends slogging.LazyLogging {
+
+  protected val serverRef = new AtomicReference[S](null.asInstanceOf[S])
```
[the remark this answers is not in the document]
DieMyst: there is no monix in this module.

build.sbt
```diff
@@ -599,7 +613,7 @@ lazy val `contract-grpc` = project
     )
   )
   .enablePlugins(AutomateHeaderPlugin)
-  .dependsOn(`contract-core-jvm`, `transport-grpc-jvm`)
+  .dependsOn(`contract-core-jvm`, `transport-grpc-jvm`, `kademlia-grpc-jvm`)
```
alari: Why is it necessary?

kademlia/grpc/js/src/main/scala/Main.scala
```diff
@@ -19,12 +19,14 @@ import fluence.crypto.SignAlgo
...
 import fluence.kad.grpc.{KademliaGrpcService, KademliaNodeCodec}
+import fluence.kad.protocol.Key
+import scodec.bits.ByteVector
 import slogging.{LogLevel, LoggerConfig, PrintLoggerFactory}
 
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.scalajs.js.annotation.{JSExport, JSExportTopLevel}
 
-@JSExportTopLevel("Main")
+@JSExportTopLevel("SomeMain")
```
alari: Name should be changed to smth meaningful.

node/src/main/scala/fluence/node/FluenceNode.scala
```diff
@@ -142,7 +143,7 @@ object FluenceNode extends slogging.LazyLogging {
 
       contact ← Contact
         .buildOwn[IO](
-          addr = upnpContact.host.getOrElse(builder.address).getHostName,
+          addr = upnpContact.host.getOrElse(InetAddress.getLocalHost).getHostName,
```
alari: It should be reverted. If you want to bind to local address, add it to config.

transport/core/src/main/scala/fluence/transport/TransportServer.scala
```diff
@@ -17,5 +17,63 @@
...
+
+/**
+ * Trait to control server lifecycle.
+ *
+ * @tparam B Server builder.
+ * @tparam S Server implementation.
+ */
+trait TransportServer[B, S] {
+
+  private val serverRef = new AtomicReference[Option[S]](None)
```
alari: Not sure that it's a good design choice. I see two possibilities:
- Server as a shared mutable resource. In this case, you make a facade that hides current state of the resource and makes itself fill the gap between actual state and required for particular operations.
- Server as an immutable instance. Unless you launch a server, you can't have an instance to bind to its events. Once server is dead, the instance is obsolete, and you need to make another one in order to gain access to its events and methods again.

In current approach, we have a TransportServer which actually is not a server, but a stateful tool that wraps builder `B` that makes an instance of `S`, which itself _is_ an implementation of server, that is stored in AtomicReference. It makes things impure and complex to think about.

transport/core/src/main/scala/fluence/transport/TransportServer.scala
```diff
@@ -17,5 +17,63 @@
...
+
+      serverRef.set(Some(server))
+    }
+
+  /**
+   * Shut the server down, release ports
+   */
+  lazy val shutdown: IO[Unit] =
+    serverRef
+      .getAndSet(None)
```
alari: `serverRef.getAndSet` is called only once, for the first shutdown, when `IO[Unit]` is instantiated. Then the same value of `IO` is returned, -- the one that depends on the state of `serverRef` at the first call moment.

So this code is completely wrong, as effect is not noticed.

transport/core/src/main/scala/fluence/transport/TransportServer.scala
```diff
@@ -17,5 +17,63 @@
...
+   * Launch server, grab ports, or fail
+   */
+  val start: IO[Unit] =
+    for {
+      _ ← if (serverRef.get().isEmpty) IO.unit else shutdown
+      b ← builder
+      server ← startServer(b)
+      _ ← onStart
+    } yield {
+
```
alari: Blank line is meaningless.

transport/core/src/main/scala/fluence/transport/TransportServer.scala
```diff
@@ -17,5 +17,63 @@
...
+            _ ← shutdownServer(srv)
+          } yield {}
+      )
+
+  def getServer: IO[S] = {
+    for {
+      serverOp ← IO(serverRef.get)
+      server ← serverOp match {
+        case Some(s) ⇒ IO.pure(s)
+        case None ⇒ IO.raiseError(new RuntimeException("Server was shutdown."))
```
alari: That's also a design choice that goes from trying to fake server's lifecycle with a stateful object. Probably we should take a look on the second approach.

transport/grpc-proxy/src/main/protobuf/test.proto
```diff
@@ -0,0 +1,21 @@
+syntax = "proto3";
```
alari: It should be moved to test folder, shouldn't it?

transport/grpc-proxy/src/main/scala/fluence/grpc/proxy/GrpcWebsocketProxy.scala
```diff
@@ -0,0 +1,74 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package fluence.grpc.proxy
+
+import cats.effect._
+import cats.implicits._
```
alari: Let's never import everything, as it makes it obscure what's actually used.

transport/grpc-proxy/src/main/scala/fluence/grpc/proxy/GrpcWebsocketProxy.scala
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * @param proxyGrpc Proxy grpc API.
+ */
+class GrpcWebsocketProxy[F[_]](proxyGrpc: ProxyGrpc[F], port: Int = 8080)(implicit F: Effect[F])
+    extends StreamApp[F] with Http4sDsl[F] {
+
+  def route(scheduler: Scheduler): HttpService[F] = HttpService[F] {
+
+    case GET -> Root / "ws" ⇒
+      val queueF = async.unboundedQueue[F, WebSocketFrame]
```
alari: Is it safe to have unbound queue? Can we use backpressure here?

transport/grpc-proxy/src/main/scala/fluence/grpc/proxy/GrpcWebsocketProxy.scala
```diff
@@ -0,0 +1,74 @@
...
+      }
+
+      queueF.flatMap { queue ⇒
+        val d = queue.dequeue.through(echoReply)
+        val e = queue.enqueue
+        WebSocketBuilder[F].build(d, e)
+      }
+  }
+
+  def stream(args: List[String], requestShutdown: F[Unit]): Stream[F, ExitCode] =
```
alari: What is it? It looks unsafe -- not idempotent. Stream of what? What happens when it's called? What is `args`?

transport/grpc-proxy/src/main/scala/fluence/grpc/proxy/InProcessGrpc.scala
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+package fluence.grpc.proxy
+
+import io.grpc.{ManagedChannel, Server, ServerServiceDefinition}
+
+import scala.collection.JavaConverters._
+import scala.language.higherKinds
+
+final case class InProcessGrpc(server: Server, channel: ManagedChannel) {
```
alari: Docs?

transport/grpc-proxy/src/main/scala/fluence/grpc/proxy/InProcessGrpcServer.scala
```diff
@@ -0,0 +1,84 @@
...
+import fluence.transport.TransportServer
+import io.grpc.ServerServiceDefinition
+import io.grpc.inprocess.{InProcessChannelBuilder, InProcessServerBuilder}
+
+/**
+ * In process grpc server with lifecycle
+ *
+ */
+class InProcessGrpcServer private (
+  builderImpl: ⇒ IO[InProcessGrpcServer.Builder],
```
alari: Why do you need to get `IO` by-name? It should be just data object -- an instance of `IO` that does the same effect each time it's runned. What effect is hidden with `=>`? Why can't it be lifted into `IO`? Where is the scaladoc?

transport/grpc-proxy/src/main/scala/fluence/grpc/proxy/InProcessGrpcServer.scala
```diff
@@ -0,0 +1,84 @@
...
+
+package fluence.grpc.proxy
+
+import cats.effect.IO
+import fluence.transport.TransportServer
+import io.grpc.ServerServiceDefinition
+import io.grpc.inprocess.{InProcessChannelBuilder, InProcessServerBuilder}
+
+/**
+ * In process grpc server with lifecycle
```
alari: In what process? Why it exists?

transport/grpc-proxy/src/main/scala/fluence/grpc/proxy/GrpcWebsocketProxy.scala
DieMyst: This is overrided method from StreamApp. I add override and TODO for future changes. This App is  example for testing for now.

build.sbt
DieMyst: I move codecs for converting kademlia classes to protobuf classes from transport-grpc to kademlia-grpc, so some dependencies are changed

transport/grpc-proxy/src/main/protobuf/grpc.proto
```diff
@@ -0,0 +1,10 @@
+syntax = "proto3";
+
+package fluence.proxy;
+
+message WebsocketMessage {
+    string service = 1;
+    string method = 2;
+    bytes protoMessage = 3;
```
RicoGit: It is possible to rename this to 'payload' ? It's default name for stuff like this, but I do not insist.

transport/grpc-proxy/src/main/scala/fluence/grpc/proxy/GrpcWebsocketProxy.scala
```diff
@@ -0,0 +1,76 @@
...
+            response ← proxyGrpc.handleMessage(message.service, message.method, message.protoMessage.newInput())
+          } yield Binary(response): WebSocketFrame
+        case m ⇒
+          F.pure(Text(s"Unexpected message: $m"): WebSocketFrame)
+      }
+
+      queueF.flatMap { queue ⇒
+        val d = queue.dequeue.through(echoReply)
+        val e = queue.enqueue
+        WebSocketBuilder[F].build(d, e)
```
RicoGit: let's try to avoid one-letter variables, please!

transport/grpc-proxy/src/main/scala/fluence/grpc/proxy/ProxyGrpc.scala
```diff
@@ -0,0 +1,123 @@
...
+   * @tparam Req Request type.
+   * @tparam Resp Response type.
+   *
+   * @return Method descriptor or None, if there is no descriptor in registered services.
+   */
+  private def getMethodDescriptor[Req, Resp](service: String, method: String): Option[MethodDescriptor[Req, Resp]] = {
+    for {
+      serviceDescriptor ← inProcessGrpc.services.find(_.getServiceDescriptor.getName == service)
+      serverMethodDefinition ← Option(
+        serviceDescriptor.getMethod(service + "/" + method).asInstanceOf[ServerMethodDefinition[Req, Resp]]
```
RicoGit: Can we avoid cast? With pattern matching for example ?
```
serviceDescriptor.getMethod(service + "/" + method) match {
    case method: ServerMethodDefinition[Req, Resp] => 
          Some(method)
    case _ => 
        logger.error("we have a problem")
        None
}
```
or some thing like that

transport/grpc-proxy/src/main/scala/fluence/grpc/proxy/ProxyGrpc.scala
```diff
@@ -0,0 +1,123 @@
...
+   *
+   * @param service Name of grpc service (class name of service).
+   * @param method Name of grpc method (method name of service).
+   * @param request Input stream of bytes.
+   *
+   * @return Response as array of bytes.
+   */
+  def handleMessage(service: String, method: String, request: InputStream): F[Array[Byte]] = {
+    for {
+      methodDescriptor ← getMethodDescriptorF[Any, Any](service, method)
```
RicoGit: Why we use F with monadError instead EitherT[E, V] ? Let's discuss this all together

transport/grpc-proxy/src/main/scala/fluence/grpc/proxy/ProxyGrpc.scala
DieMyst: This can't be done because of the type erasure

RicoGit: Let's remove getMethodDescriptor and refactor getMethodDescriptorF like this: 
```
  private def getMethodDescriptorF(service: String, method: String): F[MethodDescriptor[Any, Any]] =
    F.catchNonFatal {
      inProcessGrpc.services
        .find(_.getServiceDescriptor.getName == service)
        .map(_.getMethod(service + "/" + method).getMethodDescriptor)
    } flatMap {
      case Some(md) ⇒ F.catchNonFatal(md.asInstanceOf)
      case None ⇒ F.raiseError(new IllegalArgumentException(s"There is no $service/$method method."))
    }
  }
```
Code will be a little more safety and simple, He will not allow the user to make a mistake with putting wrong generics

DieMyst: getMethod and getMethodDescriptor could return `null`.
I can rewrite ProxyGrpc with Any, if it will be better.

transport/grpc-proxy/src/test/scala/fluence/grpc/proxy/ProxyUnaryCallSpec.scala
```diff
@@ -0,0 +1,131 @@
...
+    ): WebsocketMessage = {
+      val splitted = descriptor.getFullMethodName.split("/").toList
+
+      WebsocketMessage(splitted(0), splitted(1), req.toByteString)
+    }
+
+    "work with unary calls" in {
+      val inProcessGrpc = InProcessGrpc.build("in-process", List(service)).unsafeRunSync()
+
+      val proxyGrpc = new ProxyGrpc[IO](inProcessGrpc)
```
alari: It looks like `IO` should be the sole effect that is exposed from `ProxyGrpc`.

PR #108, opened 2018-04-12
new KVStore draft

RicoGit:
let's discuss


storage/core/jvm/src/main/scala/fluence/storage/InMemoryKVStore.scala
```diff
@@ -0,0 +1,176 @@
...
+    new TrieMapKVStore[K, V] with InMemoryKVStoreRead[K, V] with InMemoryKVStoreWrite[K, V]
+    with ReadWriteKVStore[K, V, StoreError]
+
+  /**
+   * Create in memory [[ReadWriteKVStore]] with snapshot functionality.
+   *
+   * @tparam K A type of search key
+   * @tparam V A type of value
+   */
+  def withSnapshots[K, V]: ReadWriteKVStore[K, V, StoreError] with Snapshot[KVStoreRead[K, V, StoreError]] = {
```
alari: These lines are really frustrating :)

storage/core/jvm/src/main/scala/fluence/storage/InMemoryKVStore.scala
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    /**
+     * Returns lazy ''traverse'' representation (see [[Traverse]])
+     */
+    override def traverse: Traverse[K, V, StoreError] = new Traverse[K, V, StoreError] {
+
+      override def run[FS[_]: Monad](
+        implicit FS: MonadError[FS, StoreError],
+        liftIterator: ~>[Iterator, FS]
```
alari: It's convenient to write as `Iterator ~> FS`

storage/core/jvm/src/main/scala/fluence/storage/InMemoryKVStore.scala
```diff
@@ -0,0 +1,176 @@
...
+      ): FS[(K, V)] =
+        FS.fromEither {
+          Try(liftIterator(data.iterator)).toEither.left.map(err ⇒ StoreError.traverseError(Some(err)))
```
alari: Don't you loose the laziness/async execution here by using eager `Try` data type along with eager `fromEither` lifting?

storage/core/jvm/src/main/scala/fluence/storage/InMemoryKVStore.scala
```diff
@@ -0,0 +1,176 @@
...
+   */
+  private trait InMemoryKVStoreWrite[K, V] extends InMemoryKVStore[K, V] with KVStoreWrite[K, V, StoreError] { self ⇒
+
+    /**
+     * Returns lazy ''put'' representation (see [[Put]])
+     *
+     * @param key The specified key to be inserted
+     * @param value The value associated with the specified key
+     */
+    override def put(key: K, value: V): Put[K, V, StoreError] = new Put[K, V, StoreError] {
```
alari: Why do you need to have `V` in `Put`?

storage/core/src/main/scala/fluence/kvstore/KVStorage.scala
```diff
@@ -0,0 +1,87 @@
...
+import fluence.kvstore.ops.Put.KVStorePut
+import fluence.kvstore.ops.Remove.KVStoreRemove
+import fluence.kvstore.ops.Traverse.KVStoreTraverse
+
+import scala.language.higherKinds
+
+/**
+ * Top type for any key value storage.
+ */
+trait KVStorage
```
alari: Package is `store`, trait is `storage`, confusing a bit

storage/core/src/main/scala/fluence/kvstore/KVStorage.scala
```diff
@@ -0,0 +1,87 @@
...
+trait ReadWriteKVStore[K, V, E <: StoreError] extends KVStoreRead[K, V, E] with KVStoreWrite[K, V, E]
+
+/**
+ * Key-value storage api for getting storage snapshot.
+ *
+ * @tparam S The type of returned storage snapshot.
+ */
+trait Snapshot[S <: KVStorage] {
+
+  def createSnapshot[F[_]: Applicative](): F[S]
```
alari: Are you sure there's no effect, and any `Applicative` is a good fit for creating snapshots?

storage/core/src/main/scala/fluence/kvstore/StoreError.scala
```diff
@@ -0,0 +1,54 @@
...
+
+import scala.util.control.NoStackTrace
+
+/**
+ * Error for any key value store.
+ *
+ * @param message Error message
+ * @param causedBy Error cause for wrapped exceptions
+ */
+class StoreError(message: String, causedBy: Option[Throwable] = None) extends NoStackTrace {
```
alari: Why not to make it case class?

storage/core/src/main/scala/fluence/kvstore/ops/Operation.scala
```diff
@@ -0,0 +1,67 @@
...
+ */
+trait Operation[V, E <: StoreError] {
+
+  /**
+   * Runs operation using the user defined monad,
+   * returns EitherT-wrapped result.
+   *
+   * @tparam F User defined type of monad
+   */
+  def run[F[_]: Monad]: EitherT[F, E, V]
```
alari: Are you sure any monad could be used? is it possible that you have `IO` or concurrency here?

I bet it will be tough to use `.run` instead of `apply` for functional composition.

storage/core/src/main/scala/fluence/kvstore/ops/Put.scala
```diff
@@ -0,0 +1,55 @@
...
+import scala.language.higherKinds
+
+/**
+ * Lazy representation for putting key and value.
+ *
+ * @tparam K A type of search key
+ * @tparam V A type of value
+ * @tparam E A type for any storage errors
+ */
+trait Put[K, V, E <: StoreError] extends Operation[Unit, E]
```
alari: Why to have `V` here?

alari: Why to have `K` here?

storage/core/src/main/scala/fluence/kvstore/ops/Remove.scala
```diff
@@ -0,0 +1,52 @@
...
+   * @tparam E A type for any storage errors
+   */
+  trait KVStoreRemove[K, E <: StoreError] extends KVStorage {
+
+    /**
+     * Returns lazy ''remove'' representation (see [[Remove]])
+     *
+     * @param key The specified key to be inserted
+     */
+    def remove(key: K): Remove[K, E]
```
alari: Once you passed the key, you don't need to have `K` in types system.

RicoGit: yes, sure

storage/core/src/main/scala/fluence/kvstore/ops/Put.scala
RicoGit: need to remove, tnx

storage/core/src/main/scala/fluence/kvstore/ops/Operation.scala
RicoGit: "I bet it will be tough to use .run instead of apply for functional composition."
Why? I think with 'run' looks more clear:
```
 store.get(key1)[Id]     // apply
 store.get(key1).run[Id] // run
```
Don't you think so?

RicoGit: "Are you sure any monad could be used? is it possible that you have IO or concurrency here?"
I don't understand what are you meaning, let's talking about this by phone

storage/core/src/main/scala/fluence/kvstore/KVStorage.scala
RicoGit: definitely not, I did not think about it

storage/core/src/main/scala/fluence/kvstore/KVStorage.scala
RicoGit: I will planning to rename KVStorage to KVStore

storage/core/jvm/src/main/scala/fluence/storage/InMemoryKVStore.scala
RicoGit: ok

storage/core/jvm/src/main/scala/fluence/storage/InMemoryKVStore.scala
RicoGit: fixed

storage/core/jvm/src/main/scala/fluence/storage/InMemoryKVStore.scala
RicoGit: fixed

storage/core/jvm/src/main/scala/fluence/kvstore/InMemoryKVStore.scala
```diff
@@ -0,0 +1,165 @@
...
+   * @tparam K A type of search key
+   * @tparam V A type of value
+   */
+  def withSnapshots[K, V]: ReadWriteKVStore[K, V] with Snapshot[KVStoreRead[K, V]] = {
+    new TrieMapKVStore[K, V] with InMemoryKVStoreRead[K, V] with InMemoryKVStoreWrite[K, V] with ReadWriteKVStore[K, V]
+    with Snapshot[KVStoreRead[K, V]] {
+      override def createSnapshot[F[_]]()(
+        implicit F: ApplicativeError[F, StoreError]
+      ): F[KVStoreRead[K, V]] = {
+        try F.pure(new TrieMapKVStore(data.snapshot()) with InMemoryKVStoreRead[K, V])
```
alari: What could fail here? Can we avoid using `ApplicativeError` here -- in favor of EitherT or something similar? Don't snapshotting involve IO, freezing current state of the store?

storage/core/jvm/src/test/scala/fluence/kvstore/ObservableLiftIO.scala
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * [[LiftIO]] instance built for `monix.reactive.Observable`.
+ */
+trait ObservableLiftIO extends LiftIO[Observable] {
+  override def liftIO[A](ioa: IO[A]): Observable[A] = Observable.fromIO(ioa)
+}
+
+object ObservableLiftIO {
+
+  implicit def catsObservableLiftIO[V]: ObservableLiftIO = new ObservableLiftIO {}
```
alari: Why do you need to have a type argument here?

storage/core/src/main/scala/fluence/kvstore/KVStore.scala
```diff
@@ -0,0 +1,217 @@
...
+trait ReadWriteKVStore[K, V] extends KVStoreRead[K, V] with KVStoreWrite[K, V]
+
+/**
+ * Key-value storage api for getting storage snapshot.
+ *
+ * @tparam S The type of returned storage snapshot.
+ */
+trait Snapshot[S <: KVStore] {
+
+  def createSnapshot[F[_]]()(implicit F: ApplicativeError[F, StoreError]): F[S]
```
alari: Probably you need `LiftIO` here instead of `ApplicativeError`

storage/core/src/main/scala/fluence/kvstore/KVStore.scala
```diff
@@ -0,0 +1,217 @@
...
+
+}
+
+object KVStore {
+
+  // this MonadError is needed for travers and runF and getSnapshot operations
+  implicit def storeMonadError[F[_]](implicit F: MonadError[F, Throwable]): MonadError[F, StoreError] =
```
alari: Probably it's used only for `traverse`?

storage/core/src/main/scala/fluence/kvstore/KVStore.scala
```diff
@@ -0,0 +1,217 @@
...
+
+        override def run[F[_]: Monad: LiftIO]: EitherT[F, StoreError, Option[V1]] =
+          for {
+            k ← kCodec.direct[F](key).leftMap(ce ⇒ StoreError(ce))
+            v ← store.get(k).run
+            v1 ← v match {
+              case Some(v2) ⇒
+                vCodec.inverse(v2).map(Option(_)).leftMap(ce ⇒ StoreError(ce))
+              case None ⇒
+                EitherT.right[StoreError](Monad[F].pure[Option[V1]](None))
```
alari: You may use `rightT`

storage/core/src/main/scala/fluence/kvstore/KVStore.scala
```diff
@@ -0,0 +1,217 @@
...
+            }
+          } yield v1
+
+        override def runUnsafe(): Option[V1] = {
```
alari: Probably it's better to implement `runUnsafe` only once (`final def runUnsafe`) in terms of `runF`. Unsafe running doesn't have to be optimized. But it have to match the common behavior of the operation. If you make custom implementation for unsafe running, there's a risk that you'll use different logic during testing versus running in composition.

storage/core/src/main/scala/fluence/kvstore/KVStore.scala
```diff
@@ -0,0 +1,217 @@
...
+        }
+
+      }
+
+      /**
+       * Returns lazy ''traverse'' representation (see [[Traverse]])
+       */
+      override def traverse: Traverse[K1, V1] = new Traverse[K1, V1] {
+        override def run[FS[_]: Monad: LiftIO](
+          implicit FS: MonadError[FS, StoreError],
```
alari: If you have `MonadError` in scope, you don't need to require `Monad`

storage/core/src/main/scala/fluence/kvstore/KVStore.scala
```diff
@@ -0,0 +1,217 @@
...
+
+        override def run[F[_]: Monad: LiftIO]: EitherT[F, StoreError, Unit] =
+          for {
+            k ← kCodec.direct[F](key).leftMap(ce ⇒ StoreError(ce))
+            v ← vCodec.direct[F](value).leftMap(ce ⇒ StoreError(ce))
+          } yield store.put(k, v).run
+
+        override def runUnsafe(): Unit = {
+          val k = kCodec.direct.unsafe(key)
+          val v = vCodec.direct.unsafe(value)
```
alari: Here you will have different errors for `run` and `runUnsafe`. Let's implement `runUnsafe` in terms of `run` only.

storage/core/src/main/scala/fluence/kvstore/StoreError.scala
```diff
@@ -0,0 +1,54 @@
...
+
+}
+
+object StoreError {
+
+  def apply(message: String, causedBy: Option[Throwable] = None): StoreError = new StoreError(message, causedBy)
+
+  def apply(exception: Throwable): StoreError = new StoreError(exception.getMessage, Some(exception))
+
+  def getError[K](key: K, causedBy: Option[Throwable] = None): StoreError =
```
alari: These builders names are a bit confusing, as they sound as verbs. Maybe `errorGet` is better? or `forGet`? Looking on method, I expect to have a getter for some error from some container.

storage/core/src/main/scala/fluence/kvstore/ops/Get.scala
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+package fluence.kvstore.ops
+
+import scala.language.higherKinds
+
+/**
+ * Lazy representation for getting key and value.
+ */
+trait Get[V] extends Operation[Option[V]]
```
alari: Could `Operation` be a sealed trait and contain all its implementations in the same file -- or even in the same companion object?

Another question: why do you need to have subtypes of `Operation` at all? There's no pattern matching over them... Why not to simply return `Operation` itself?

storage/core/src/main/scala/fluence/kvstore/ops/Operation.scala
```diff
@@ -0,0 +1,72 @@
...
+   *
+   * @tparam F User defined type of monad
+   */
+  def run[F[_]: Monad: LiftIO]: EitherT[F, E, V]
+
+  /**
+   * Runs unsafe operation, '''throw the error if it happens'''.
+   * Intended to be used '''only in tests'''.
+   */
+  def runUnsafe(): V
```
alari: It should be marked final and implemented in terms of `run`.

storage/core/src/main/scala/fluence/kvstore/ops/Traverse.scala
```diff
@@ -0,0 +1,50 @@
...
+
+import scala.language.higherKinds
+
+/**
+ * Lazy representation for traversing all values.
+ *
+ * @tparam K A type of search key
+ * @tparam V A type of value
+ */
+trait Traverse[K, V] {
```
alari: There's a `Traverse` typeclass in Cats. Either this class should be made package private, or renamed to smth like `TraverseOp` to highlight that it's an operation.

storage/core/src/main/scala/fluence/kvstore/ops/Traverse.scala
```diff
@@ -0,0 +1,50 @@
...
+
+  /**
+   * Returns FS stream of all pairs in current key-value store.
+   *
+   * @param FS MonadError type class for user defined type FS
+   * @param liftIterator Creates FS stream from [[Iterator]]
+   *
+   * @tparam FS User defined type of stream with monadError
+   */
+  def run[FS[_]: Monad: LiftIO](implicit FS: MonadError[FS, StoreError], liftIterator: Iterator ~> FS): FS[(K, V)]
```
alari: There are too many typeclasses requested.

storage/core/src/main/scala/fluence/kvstore/ops/Traverse.scala
```diff
@@ -0,0 +1,50 @@
...
+
+  /**
+   * Returns [[Iterator]] with all key-value pairs for current KVStore,
+   * '''throw the error if it happens'''. Intended to be used '''only in tests'''.
+   */
+  def runUnsafe: Iterator[(K, V)]
```
alari: It should be marked `final` and implemented in terms of `run`.

RicoGit: Can you explain why it should be that? It difficult to implement in this case and potentially may be better implement 'unsafe' call independent from other methods. And may be better implement 'run' via 'runUnsafe', like in IO.runAsync? I would prefer to leave both methods for more flexibility.

storage/core/src/main/scala/fluence/kvstore/ops/Traverse.scala
RicoGit: I'll remove ME

storage/core/src/main/scala/fluence/kvstore/ops/Get.scala
RicoGit: "Why not to simply return Operation itself?" - make sense.

storage/core/src/main/scala/fluence/kvstore/KVStore.scala
RicoGit: tnx

storage/core/src/main/scala/fluence/kvstore/KVStore.scala
RicoGit: yes

storage/core/src/main/scala/fluence/kvstore/KVStore.scala
RicoGit: No, but it will possible to remove ME from everywhere where LiftIO required, I will do it

storage/core/src/main/scala/fluence/kvstore/StoreError.scala
RicoGit: ok

storage/core/jvm/src/main/scala/fluence/kvstore/InMemoryKVStore.scala
RicoGit: you're right, IO is more suitable be in this case

storage/core/jvm/src/test/scala/fluence/kvstore/ObservableLiftIO.scala
RicoGit: removed

storage/core/src/main/scala/fluence/kvstore/KVStore.scala
RicoGit: "different logic during testing versus running"
yes, it's an argument. I going to express 'rununsafe' via 'run' or vice versa, but final is not necessary, user should be able to override default behaviour

storage/core/src/main/scala/fluence/kvstore/KVStore.scala
RicoGit: MonadError isn't needed anymore.

PR #111, opened 2018-04-17
crypto-refactoring

alari:


crypto/core/src/main/scala/fluence/crypto/Crypto.scala
```diff
@@ -15,19 +15,20 @@
...
+  type KeyPairGenerator = Func[Option[Array[Byte]], KeyPair]
 
-}
+  // TODO: move it to MonadicalEitherArrow, make liftTry with try-catch, and easy conversions for other funcs and bijections
+  implicit val liftCodecErrorToCrypto: CodecError ⇒ CryptoError = err ⇒ CryptoError("Codec error", Some(err))
 
-object TestHasher {
-  def apply(hasher: CryptoHasher[Array[Byte], Array[Byte]] = TestCryptoHasher): TestHasher = new TestHasher(hasher)
+  implicit def codec[A, B](implicit codec: PureCodec[A, B]): Bijection[A, B] =
+    Bijection(fromOtherFunc(codec.direct), fromOtherFunc(codec.inverse))
```
RicoGit: What does this method do?

b-tree/client/src/main/scala/fluence/btree/client/MerkleBTreeClient.scala
```diff
@@ -295,11 +296,8 @@ class MerkleBTreeClient[K] private (
    */
   override def initRemove(key: K, version: Long): Task[RemoveState[Task]] = ???
 
-  private def binarySearch(key: K, keys: Array[Key]): Task[SearchResult] = {
-    import fluence.crypto.cipher.CryptoSearching._
-    implicit val decrypt: Key ⇒ Task[K] = keyCrypt.decrypt
-    search[Task, Key](keys).binarySearch(key)
-  }
+  private def binarySearch(key: K, keys: Array[Key]): Task[SearchResult] =
+    CipherSearch.binary(keys, keyCrypt.inverse).runF[Task](key)
```
RicoGit: The idea was create api for binarySearch the same as in scala STDLIB (scala.collection.Searching). Why you remove CryptoSearching?

crypto/core/src/main/scala/fluence/crypto/cipher/CipherSearch.scala
```diff
@@ -0,0 +1,51 @@
...
+import cats.Monad
+import cats.data.EitherT
+import fluence.crypto.{Crypto, CryptoError}
+
+import scala.collection.Searching.{Found, InsertionPoint, SearchResult}
+import scala.language.higherKinds
+
+object CipherSearch {
+
+  def binary[A, B](coll: IndexedSeq[A], decrypt: Crypto.Func[A, B])(
```
RicoGit: There was a good docs in a CryptoSearchImpl

crypto/core/src/test/scala/fluence/crypto/CryptoSearchingSpec.scala
```diff
@@ -17,33 +17,30 @@
...
+      val encryptedElements = plainTextElements.map(t ⇒ crypt.direct.unsafe(t))
 
-      import fluence.crypto.cipher.CryptoSearching._
-      implicit val decryptFn: Array[Byte] ⇒ Try[String] = crypt.decrypt
+      val search = CipherSearch.binary(encryptedElements, crypt.inverse)
 
-      encryptedElements.binarySearch("B").get shouldBe Found(1)
-      encryptedElements.binarySearch("D").get shouldBe Found(3)
-      encryptedElements.binarySearch("E").get shouldBe Found(4)
+      search.unsafe("B") shouldBe Found(1)
```
RicoGit: I don't like this Api and when 'unsafe' or 'runF' takes parameter,  it should be just a trigger. I would prefer api like this 
```
import CiperSearch._
// This api is understandable and familiar to everybody
encryptedElements.binarySearch("B").runUnsafe() 
```
crypt.inverse and ordering should be an implicit parameters for binarySearch method.

crypto/hashsign/js/src/test/scala/flyence/crypto/EcdsaSpec.scala
```diff
@@ -15,13 +15,13 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-package fluence.crypto
+package flyence.crypto
```
RicoGit: a typo!!!  fl**y**ence -> fluence

crypto/hashsign/jvm/src/test/scala/fluence/crypto/JvmHashSpec.scala
```diff
@@ -49,7 +49,7 @@ class JvmHashSpec extends WordSpec with Matchers {
 
       val hasher = JdkCryptoHasher.Sha1
 
-      ByteVector(hasher.hash(arr)).toBase64 shouldBe base64Check
+      ByteVector(hasher.unsafe(arr)).toBase64 shouldBe base64Check
```
RicoGit: Looks bad in my taste, I would prefer `hasher.hash(bytes).runUnsafe`

crypto/hashsign/jvm/src/test/scala/fluence/crypto/SignatureSpec.scala
```diff
@@ -82,18 +82,18 @@ class SignatureSpec extends WordSpec with Matchers {
 
     "throw an errors on invalid data" in {
       val algo = Ecdsa.signAlgo
-      val keys = algo.generateKeyPair().extract
+      val keys = algo.generateKeyPair.unsafe(None)
```
RicoGit: `algo.generateKeyPair(None).unsafe` looks better for me.
'generateKeyPair' should take parameter, this parameter(seed) is responsability of  'generateKeyPair', not a 'unsafe'. 

It's ok to choose an Effect in the end, but not parameter imho.
If 'unsafe', 'runF' will be parameterless, its would be easy to generalize, perhaps.

crypto/core/src/main/scala/fluence/crypto/Crypto.scala
alari: As the types say, it converts `PureCodec` into `Crypto.Bijection`, lifting codec errors to crypto errors. It's used to compose cipher bijection (encrypt/decrypt) with codecs and get cipher as a result.

build.sbt
```diff
@@ -437,15 +437,56 @@ lazy val `b-tree-server` = project
     `b-tree-client-jvm` % "compile->test"
   )
 
-lazy val `crypto` = crossProject(JVMPlatform, JSPlatform)
+//////////////////////////////////////////////////////////////
+//////////////         CRYPTO BEGIN        ///////////////////
+//////////////////////////////////////////////////////////////
+lazy val `crypto-core` = crossProject(JVMPlatform, JSPlatform)
```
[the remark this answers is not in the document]
alari: This pull request is the first step to moving all of `crypto` to separate repository, so that `build.sbt` will become smaller. For a big amount of sbt directives, like cross build ones, I found no way to split. 

Generally, it's out of this PR scope.

crypto/core/src/main/scala/fluence/crypto/cipher/CipherSearch.scala
alari: Yep, I'll bring them back, thanks

crypto/core/src/test/scala/fluence/crypto/CryptoSearchingSpec.scala
alari: I can't agree with this. For any element, searching is the same function, returning the same result. There's no reason to have intermediary object to have runners upon. I made the whole API sound the same, codec-fashioned way, that's why refactoring for CipherSearch is being made. I don't think it must conform scala's standard library conventions.

crypto/hashsign/jvm/src/test/scala/fluence/crypto/JvmHashSpec.scala
alari: Sorry, I see no reason to have an intermediate object instead of normal function: no functional composability, no meaning.

crypto/hashsign/jvm/src/test/scala/fluence/crypto/SignatureSpec.scala
alari: Still can't agree. It's an arrow from seed to keypair.

PR #113, opened 2018-04-23
Hide 'traverse' method behind a taking snapshot of KVStore 

RicoGit:
traverse without snapshot has non deterministic behaviour


storage/core/src/main/scala/fluence/kvstore/KVStore.scala
```diff
@@ -92,7 +112,7 @@ trait ReadWriteKVStore[K, V] extends KVStoreRead[K, V] with KVStoreWrite[K, V]
  *
  * @tparam S The type of returned storage snapshot.
  */
-trait Snapshot[S <: KVStore] {
+trait Snapshot[S <: KVStoreRead[_, _]] {
```
alari: Some ability description, like `Snapshotable`, fits better then noun. KVStore with ability to take snapshot is not a snapshot, is it?

PR #121, opened 2018-05-16
Getting GRPC stub as IO

alari:


contract/grpc/src/main/scala/fluence/contract/grpc/client/ContractsCacheClient.scala
```diff
@@ -53,13 +53,15 @@ class ContractsCacheClient[C: ContractValidate](stub: ContractsCacheStub)(
...
-      binContract ← IO.fromFuture(IO(stub.find(req)))
+      binContract ← IO.fromFuture(stub.map(_.find(req)))
       contract ← codec.decode(binContract)
       // contract from the outside required validation
       _ ← contract.validateME[IO]
     } yield Option(contract)).recover {
       case err ⇒
         logger.warn(s"Finding contract failed, cause=$err", err)
+        logger.warn(err.toString)
+        err.printStackTrace()
```
DieMyst: Print Stacktrace is not good here

dataset/node/src/main/scala/fluence/dataset/node/NodePut.scala
```diff
@@ -73,104 +73,106 @@ class NodePut[F[_]: Async](service: DatasetStorageRpc[F, Observable])(
...
-           */
-          override def nextChildIndex(keys: Array[Key], childsChecksums: Array[Hash]): F[Int] =
-            toF(
-              for {
-                _ ← pushServerAsk(
-                  PutCallback.Callback.NextChildIndex(
-                    AskNextChildIndex(
-                      keys = keys.map(k ⇒ ByteString.copyFrom(k.bytes)),
-                      childsChecksums = childsChecksums.map(c ⇒ ByteString.copyFrom(c.bytes))
+      oldValue ← service
```
DieMyst: is this changes important?

alari: The only important change here is `.to[F]` at the end of the method. However, it forces scalafmt to reformat the whole.

RicoGit: so sad

PR #135, opened 2018-06-25
Rocks db kv store

RicoGit:
Added documentation for KVStore usage. And made possibility to create storage with codecs and snapshot.

### DIFF
--- a/storage/core/jvm/src/main/scala/fluence/kvstore/InMemoryKVStore.scala
+++ b/storage/core/jvm/src/main/scala/fluence/kvstore/InMemoryKVStore.scala
@@ -53,6 +53,8 @@ object InMemoryKVStore {
 
     protected def data: TrieMap[K, V]
 
+    override def close(): IO[Unit] = IO.unit
+
   }
 
   /**
--- a/storage/core/jvm/src/main/scala/fluence/kvstore/InMemoryKVStore.scala
+++ b/storage/core/jvm/src/main/scala/fluence/kvstore/InMemoryKVStore.scala
@@ -150,12 +150,12 @@ object InMemoryKVStore {
    * @tparam K A type of search key
    * @tparam V A type of value
    */
-  def withSnapshots[K, V]: InMemoryKVStore[K, V] with Snapshotable[InMemoryKVStoreRead[K, V]] = {
+  def withSnapshots[K, V]: InMemoryKVStore[K, V] with Snapshotable[K, V] = {
 
     /** Allows to create a point-in-time view of a storage. */
-    new InMemoryKVStore[K, V] with Snapshotable[InMemoryKVStoreRead[K, V]] {
-      override def createSnapshot[F[+ _]: Monad: LiftIO]: F[InMemoryKVStoreRead[K, V]] =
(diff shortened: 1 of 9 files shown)


storage/core/jvm/src/main/scala/fluence/kvstore/InMemoryKVStore.scala
```diff
@@ -152,7 +154,7 @@ object InMemoryKVStore {
 
     /** Allows to create a point-in-time view of a storage. */
     new InMemoryKVStore[K, V] with Snapshotable[InMemoryKVStoreRead[K, V]] {
-      override def createSnapshot[F[_]: LiftIO]: F[InMemoryKVStoreRead[K, V]] =
+      override def createSnapshot[F[+ _]: Monad: LiftIO]: F[InMemoryKVStoreRead[K, V]] =
```
alari: Why do you need to enable variance here? In other case will algorithm work or not?

RicoGit: Covariant here is necessary. Store with snapshot and codecs didn't work without that.

alari: Could you explain why?

RicoGit:
```
trait Snapshotable[+S <: KVStoreRead[_, _]] {
  def createSnapshot[F[+ _]: Monad: LiftIO]: F[S]
}
```
Because +S is covariant, because Snapshotable[KVStore] should be parent for Snapshotable[RocksDbStore].
Method `createSnapshot` return type F[S] where S is covariant and can't be placed into F[_] (invariant position) but can be placed in F[+_]

alari: It means that everywhere we use KVStore, we will have to change `F[_]: Monad` into `F[+_]: Monad`, forcing variance where we don't intend it. That's why type constructor variance is avoided whereas possible.
Let's investigate this case and try to find out a way to keep things non-variant, if we ever have a chance to.

RicoGit: hmm, I didn't think about it. Good solution seems to change [F[_] : Monad : LIftIO] to IO[S]
IO has covariant type parameter and IO is  suitable effect for 'createSnapshot'
```
trait Snapshotable[+S <: KVStoreRead[_, _]] {
  def createSnapshot: IO[S]
}
```
Anyway in any place when we need KVStore, 'F' require 'liftIO' and F will be able to wrap IO to F. What do you think?

alari: Why to return a specific subtype of `KVStoreRead`? Maybe it's better to put key and value types as arguments to `Snapshotable`, and return a `KVStoreRead`, hiding the implementation?

```scala
trait Snapshotable[K, V] {
  def createSnapshot[F[_]: Monad: LiftIO]: F[KVStoreRead[K, V]]
}
```

It makes types definition much more simple to understand, not adding unnecessary constraints.

RicoGit: I Did as you propose.

PR #136, opened 2018-07-04
Connection refactoring

DieMyst:


kademlia/grpc/src/main/scala/fluence/kad/grpc/client/KademliaClient.scala
```diff
@@ -98,10 +93,17 @@ class KademliaWebsocketClient(connection: IO[WebsocketPipe[WebsocketMessage, Web
...
+      res ← nodeContactCodec.runF[IO](responseBytes)
     } yield res
   }
 }
+
+object KademliaClient {
+  def apply(streamHandler: Connection)(
+    implicit
+    codec: PureCodec[Node[Contact], protobuf.Node],
+    ec: ExecutionContext
```
RicoGit: Looks like ExecutionContext isn't used in KademliaClient and might be removed.

project/plugins.sbt
```diff
@@ -16,4 +16,6 @@ addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.9.0")
 
 addSbtPlugin("com.lihaoyi" % "workbench" % "0.4.0")
 
+addSbtPlugin("ch.epfl.scala" % "sbt-bloop" % "1.0.0-M11")
```
RicoGit: Isn't it break sbt or idea for everyone(without nightly build and so on)? What about scalaJs?

scala-multistream/grpc/jvm/src/main/scala/fluence/grpc/ServiceManager.scala
```diff
@@ -0,0 +1,71 @@
...
+package fluence.grpc
+
+import cats.MonadError
+import cats.syntax.flatMap._
+import io.grpc._
+
+import scala.collection.JavaConverters._
+import scala.language.higherKinds
+
+class ServiceManager(services: Map[String, MethodDescriptor[Any, Any]]) {
```
RicoGit: would be nice to have a docs for this class

transport/websocket-js/src/main/scala/fluence/transport/websocket/WebsocketConnection.scala
```diff
@@ -0,0 +1,87 @@
...
+    WebsocketMessage(service, method, requestId, WebsocketMessage.Response.Payload(ByteString.copyFrom(payload)))
+  }
+
+  override def handle(
+    service: String,
+    method: String,
+    requests: Observable[Array[Byte]]
+  ): IO[Observable[Array[Byte]]] = {
+    connection.map { websocketClient ⇒
+      val requestId = Random.nextLong()
```
RicoGit: There is common practice to use UUID for requiest id. 
`UUID.randomUUID().getMostSignificantBits()` returns long value.
Up to you.

kademlia/grpc/src/main/scala/fluence/kad/grpc/client/KademliaClient.scala
DieMyst: you are right

scala-multistream/grpc/jvm/src/main/scala/fluence/grpc/ServiceManager.scala
DieMyst: no, it is only for generation helper classes by sbt command

transport/websocket-js/src/main/scala/fluence/transport/websocket/WebsocketConnection.scala
DieMyst: there is no UUID in js

PR #148, opened 2018-09-03
Swarm init

DieMyst:
https://github.com/fluencelabs/fluence/issues/170


externalstorage/src/main/scala/fluence/swarm/SwarmClient.scala
```diff
@@ -0,0 +1,319 @@
...
+   */
+  def downloadRaw[T](target: String): EitherT[F, SwarmError, RawResponse] = {
+    val downloadURI = uri(BzzRaw, target)
+    logger.info(s"Download manifest request. Target: $target")
+    sttp
+      .response(asJson[RawResponse])
+      .get(downloadURI)
+      .send()
+      .toEitherT(er => SwarmError(s"Error on downloading manifest from $downloadURI. $er"))
+      .flatMap(
```
folex: do you need `flatMap` here? or just `map`?

folex: maybe you can avoid `EitherT[Either]` at all

externalstorage/src/main/scala/fluence/swarm/SwarmClient.scala
```diff
@@ -0,0 +1,330 @@
...
+   *                  Setting it in the future will prevent nodes from finding content until the clock hits startTime.
+   *                  Setting it in the past allows you to create a history for the resource retroactively.
+   * @param ownerAddr Swarm address (Ethereum wallet address)
+   * @param data content the Mutable Resource will be initialized with.
+   * @param multiHash is a flag indicating whether the data field should be interpreted as a raw data or a multihash.
+   *                  TODO There is no implementation of multiHashed data for now.
+   * @param signer signature algorithm. Must be ECDSA for real Swarm node.
+   * @return hash of metafile. This is the address of mutable resource.
+   */
+  def initializeMRU(
```
alari: I'm feeling like there should be separate `SwarmMRU` for a mutable resource, that takes at least `name` and `frequency` as an instance arguments and simplifies the MRU lifecycle.

Not necessary to do in this PR.

externalstorage/src/main/scala/fluence/swarm/SwarmClient.scala
```diff
@@ -0,0 +1,330 @@
...
+    } yield resp
+  }
+
+  /**
+   * Upload a metafile for future use.
+   * @see https://swarm-guide.readthedocs.io/en/latest/usage.html#creating-a-mutable-resource
+   *
+   * @param name optional resource name. You can use any name.
+   * @param frequency expected time interval between updates, in seconds
+   * @param startTime time the resource is valid from, in Unix time (seconds). Set to the current epoch
```
alari: Probably `startTime` should be also bound inside a hypothetical `SwarmMRU` instance.

externalstorage/src/main/scala/fluence/swarm/SwarmClient.scala
```diff
@@ -0,0 +1,330 @@
...
+        logger.info(s"A metafile of a mutable resource has been uploaded. Hash: $r.")
+        r
+      }
+  }
+
+  /**
+   * Update a mutable resource.
+   * @see https://swarm-guide.readthedocs.io/en/latest/usage.html#updating-a-mutable-resource
+   *
+   * @param name optional resource name. You can use any name.
```
alari: How could it be used? May I retrieve the resource with its name later?

You have a huge copy-paste for arguments there. It complicates the work with MRU, as you need to pass around the arguments everywhere. Maybe DRY?

externalstorage/src/main/scala/fluence/swarm/SwarmConstants.scala
```diff
@@ -0,0 +1,73 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package fluence.swarm
+
+object SwarmConstants {
+  val keyLength: Int = 32
```
alari: Usually, constants start with an uppercase letter in Scala. That's due to the existence of case objects, which are the constant instances.

externalstorage/src/main/scala/fluence/swarm/SwarmConstants.scala
```diff
@@ -0,0 +1,73 @@
...
+  val updateLookupLength: Int = 4 + 4 + keyLength
+  val metaHashLength: Int = keyLength
+
+  // 64 bytes ECDSA size + 1 byte of recovery id
```
alari: Shouldn't it be removed?

externalstorage/src/main/scala/fluence/swarm/SwarmError.scala
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+
+package fluence.swarm
+
+import scala.language.higherKinds
+import scala.util.control.NoStackTrace
+
+// TODO change this error to errors with hierarchy
+case class SwarmError(message: String, causedBy: Option[Throwable] = None) extends NoStackTrace {
```
alari: Do you have an issue about it?

externalstorage/src/main/scala/fluence/swarm/responses/RawResponse.scala
```diff
@@ -0,0 +1,45 @@
...
+import io.circe.generic.semiauto.{deriveDecoder, deriveEncoder}
+
+/**
+ * One file in the manifest with additional information.
+ *
+ * @param hash address of file in swarm
+ * @param contentType type of stored data (tar, multipart, etc)
+ * @param mod_time time of modification
+ */
+case class Entry(hash: String, contentType: String, mod_time: String)
```
alari: Inconsistent naming.

There should be only camelCase names.

If it's used to reflect JSON structure, it should be mentioned explicitly, maybe even in the class name.

externalstorage/src/main/scala/fluence/swarm/responses/RawResponse.scala
```diff
@@ -0,0 +1,45 @@
...
+  implicit val entryDecoder: Decoder[Entry] = deriveDecoder
+  implicit val entryEncoder: Encoder[Entry] = deriveEncoder
+}
+
+/**
+ * Representation of raw manifest with entries.
+ *
+ * @param entries list of files under manifest
+ */
+case class RawResponse(entries: List[Entry])
```
alari: Why it's named `RawResponse` when you know that it's a manifest? What stands for `Raw`? Why not name it `ManifestResponse`?

externalstorage/src/main/scala/fluence/swarm/SwarmConstants.scala
DieMyst: They will be used later when functionality will be extended. It is a little bit complicated to search these constants in Swarm source, so, I decided to save these constants for future usage.

externalstorage/src/main/scala/fluence/swarm/responses/RawResponse.scala
DieMyst: Ok, this can be achieved by Circe itself.

PR #183, opened 2018-09-20
Remake Vm errors hierarchy 

RicoGit:
Allows type-safe working with produced errors. For example:
  ```
  val vm: fluence.vm.WasmVm = ???
     vm.invoke[IO](None, "fnName", Seq()).leftMap {
       case InvalidArgError(message, cause) ⇒  ???
       case NoSuchFnError(message, cause) ⇒  ???
       case TrapError(message, cause) ⇒ ???
       // case InternalVmError(message, cause) ⇒ ???
     }
```
   
Compiler checks that all of possible error types will be handled. If some of error types will not be handled compiler produce warning like this:
```  
     Warning:(103, 50) match may not be exhaustive.
     It would fail on the following input: InternalVmError(_, _)
     vm.invoke[IO](None, "fnName", Seq()).leftMap {
```
   This mean that {@link InternalVmError} will not be handled.

### DIFF
--- a/vm/src/main/scala/fluence/vm/package.scala
+++ b/vm/src/main/scala/fluence/vm/package.scala
@@ -15,8 +15,9 @@
  */
 
 package fluence
-import cats.Applicative
+import cats.{Applicative, Functor}
 import cats.data.EitherT
+import cats.effect.IO
 
 import scala.language.higherKinds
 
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: vm/src/main/scala/fluence/vm/package.scala [hunk 1]
```diff
@@ -34,4 +35,13 @@ package object vm {
     EitherT.fromEither[F](either)
   }
 
+  implicit class VmErrorMapper[F[_]: Functor, E <: VmError, T](eitherT: EitherT[F, E, T]) {
+
+    def toVmError: EitherT[F, VmError, T] = {
+      eitherT.leftMap { e: VmError ⇒
```
alari: There should be a method in EitherT that lifts error type to the supertype.

alari: Yet I haven't found one :(

Issue #168, opened 2018-09-10
State machine session data cleanup


DmiSergeev:
### Motivation
Currently, the State machine stores all the session metadata and results without any limits on the size of these data, leading to high storage usage – this may cause a performance degradation.

### Proposition
To implement automatic cleanup procedures in the State machine they must be triggered by client session termination (see #167).

Also this 'triggering' should ideally be tunable somehow, making it possible to define some 'period' (measured in time or number of transactions or blocks) when the cleanup is actually performed, possibly depending on how the termination occured (whether it was normally closed, failed, or expired).

Data that should be cleaned up [include](../blob/master/statemachine/src/main/scala/fluence/statemachine/tree/StorageKeys.scala):
* queued payloads
* statuses
* results of successful invocation and error messages
* client/session/tx hierarchy keys

alari:
Closed as obsolete

Issue #187, opened 2018-09-21
Vm error strong hierarchy


RicoGit:
### Motivation

Virtual Machine (VM) produces a lot of different error kinds. Right now VM's methods signatures don't provide a knowledge about produced errors. Each VM method returns a general type of error and there is no way to know, without reading source code, which errors might be produced and which errors will never appear. This is an error-prone and unfriendly API. When new error type will be created a compiler will not make a tip about that and new error might be unhandled.

### Proposition 

* Error hierarchy should allow a compiler to make a tip when we forgot to handle some error kind. 
* Method signatures must reflect possible errors. For example, `Wasm.apply` should produce `InitializationError`, `TrapError`, `InternalVmError` and we need to know this from the signature of the method `apply`. 

#### Possible solution

Create sealed traits for each method in a VM, for `invoke` - `InvokationError`, for `getVmState` - `GetVmStateError` and so on. `InvokationError` sealed trait should become a parent for each error which might be produced in the `invoke` method. It allows the compiler to detect unhandled errors in pattern-matching through exhaustiveness check.
For example:
`sealed trait InternalVmError extends ApplyError with InvokeError with GetVmStateError` means that InternalVmError can be produced by methods apply, invoke, getVmState

If there are more suitable solutions with the less boilerplate code we should consider them too.

alari:
Closed as obsolete

Issue #155, opened 2018-09-07
Error hierarchy for State machine


DmiSergeev:
### Motivation
Currently, a significant amount of State machine code uses `String` as left part of `Either`/`EitherT` – for handling various errors and exceptional cases. It includes both creation of new error messages for some [edge cases](../blob/master/statemachine/src/main/scala/fluence/statemachine/state/TendermintStateHolder.scala#L54) and wrapping [caught exceptions](../blob/master/statemachine/src/main/scala/fluence/statemachine/tx/TxParser.scala#L46). It makes the current code not expressive enough and not clean.

### Proposition
A carefully designed error class/object hierarchy is required. [StateMachineError](../blob/master/statemachine/src/main/scala/fluence/statemachine/error/StateMachineError.scala) as the root class has already applied for the [transaction invocation](../blob/master/statemachine/src/main/scala/fluence/statemachine/tx/VmOperationInvoker.scala) and other statemachine part. VM project [error hierarchy](https://github.com/fluencelabs/fluence/issues/187) can be used as a reference.

alari:
Closed as obsolete

PR #188, opened 2018-09-22
Session termination and Python enhancements

### LINKED ISSUE #158: Python client enhancements
### Motivation
Current Data engine Python client has some drawbacks:
* It is implemented in 2.7 version. Despite it is still active, it's adoption is decreasing and it might be treated as outdated.
* There are some style issues, like long lines, inconsistent naming, etc.
* There are some vague and messy function [examples](../blob/master/client/misc_utils.py#L22).
* License headers are not generated for Python files.

Those drawbacks affect project quality and obstruct further maintenance and enhancements.

### Proposed change
* Rewrite client to Python 3.
* Resolve style issues (long lines, naming etc.)
* Resolve some vague and messy implementation issues.
* Generate absent license headers.

### LINKED ISSUE #167: Client session termination management
### Motivation
Currently, there is no concept of the _closed session_: clients come and go, create sessions, send transactions, and forget about them. This concept is important, without it we cannot:
* Understand that the session data can be cleaned up (#168).
* Provide the client an information that some session was aborted during the execution of some of its transactions.

### Proposition
First, we need to introduce the session **state**: it might be _active_ or _terminated_. Then, we define 3 termination ways:
* Explicit termination by the client (normal case): the client needs to do it after it finishes sending transactions.
* Aborted (or failed) termination: if an unrecoverable error happened during transaction processing.
* Expiration: the session is terminated automatically after some 'period' (measured in time or number of transactions or blocks) without transactions from the client.

Whichever the session termination type was, if the client sends any transaction in the terminated session, it fails and the client is notified about it. It should also cause all in-flight requests to fail, so if the client is still waiting on the response (e. g. on promise or awaitable) or retrieves it later, it will receive an error.

This proposal implies some mechanism for closed session's data cleanup, it is to be implemented in #168.

DmiSergeev:
* Session termination (closes #167) – the client sessions are closed by 1 of 3 reasons: explicitly by the client, due to an error, being expired after inactivity. A related 'Session data cleanup' issue (#168) not included in this issue and not yet implemented. 
* Python cross-version client (2 & 3) and refactoring (closes #158).


statemachine/src/main/scala/fluence/statemachine/tx/FunctionCallDescription.scala
```diff
@@ -0,0 +1,85 @@
...
+case class FunctionCallDescription(module: Option[String], functionName: String, argList: List[String])
+
+object FunctionCallDescription {
+
+  /**
+   * Description for reserved non-VM function call that explicitly closes sessions by the client.
+   */
+  val CloseSession = FunctionCallDescription(None, "@closeSession", Nil)
+
+  private val FunctionWithModuleAndArgListPattern: Regex = "(\\w+)\\.(\\w+)\\(([\\w,]*)\\)".r
```
DieMyst: small examples here will be good I think

statemachine/src/main/scala/fluence/statemachine/tx/SessionSummary.scala
```diff
@@ -0,0 +1,85 @@
...
+  /**
+   * Deserializes the summary from the given [[StoreValue]].
+   *
+   * @param value serialized summary
+   */
+  def fromStoreValue(value: StoreValue): Option[SessionSummary] =
+    (for {
+      parsedJson <- parseJson(value)
+      summary <- parsedJson.as[SessionSummary]
+    } yield summary).toOption
```
DieMyst: when you use toOption, you lose errors. You can use either `instead` or write errors in logs.
It is good to do this in all places, like in `TxParser.parseTx`
something like `HexCodec.hexToString(...)` writes useful information in errors and this should be in logs at least.

statemachine/src/main/scala/fluence/statemachine/tx/TxProcessor.scala
```diff
@@ -60,16 +64,18 @@ class TxProcessor[F[_]](
    */
   def processNewTx(tx: Transaction): F[Unit] =
     for {
+      _ <- incrementTxCounter()
```
DieMyst: what if we raise an error somewhere? Will it be added to the blockchain guaranteed?

DieMyst: in `receivedDeliverTx` if verifying will be wrong, the client will increment counter and the server will not.

statemachine/src/main/scala/fluence/statemachine/tx/TxProcessor.scala
```diff
@@ -42,10 +44,12 @@ import scala.language.higherKinds
...
  * @param mutableConsensusState Consensus state, affected by every newly appended transaction
  * @param vmInvoker object used to perform immediate calls to VM
+ * @param config settings which tx processing logic depends on
  */
 class TxProcessor[F[_]](
   private val mutableConsensusState: MutableStateTree[F],
-  private val vmInvoker: VmOperationInvoker[F]
+  private val vmInvoker: VmOperationInvoker[F],
+  private val config: StateMachineConfig
 )(implicit F: Monad[F])
```
DieMyst: Are there no exceptions? Is it good and safe to use Monad under TxProcessor?

statemachine/src/main/scala/fluence/statemachine/tx/TxStateDependentChecker.scala
```diff
@@ -0,0 +1,57 @@
...
+  def check(uncheckedTx: Transaction): EitherT[F, String, Transaction] =
+    for {
+      currentState <- EitherT.right(state)
+      deduplicatedTx <- EitherT.cond(
+        !currentState
+          .hasValue(StoragePaths.txStatusPath(uncheckedTx.header)),
+        uncheckedTx,
+        ClientInfoMessages.DuplicatedTransaction
+      )
+      checkedTx <- EitherT.cond(
```
folex: And maybe return `Unit` here too? You have `deduplicatedTx` and `checkedTx`, but they're the same as `uncheckedTx`, and you even using `uncheckedTx.header` below. It's kinda confusing.

statemachine/src/main/scala/fluence/statemachine/tx/TxStateDependentChecker.scala
```diff
@@ -0,0 +1,57 @@
...
+  /**
+   * Checks whether given `tx` is unique and its state is active against provided [[state]].
+   *
+   * @param uncheckedTx parsed transaction
+   * @return either checked transaction or error message
+   */
+  def check(uncheckedTx: Transaction): EitherT[F, String, Transaction] =
+    for {
+      currentState <- EitherT.right(state)
+      deduplicatedTx <- EitherT.cond(
```
folex: maybe return `Unit` instead of `deduplicatedTx` here? Code looks like you're changing tx, but you're not.

statemachine/src/main/scala/fluence/statemachine/tx/TxStateDependentChecker.scala
folex: Some kind of `EitherT.assert` would be useful if there was such a method :)

statemachine/src/main/scala/fluence/statemachine/tx/VmOperationInvoker.scala
```diff
@@ -82,24 +58,6 @@ object VmOperationInvoker {
    *
    * @param vmError error returned from VM
    */
-  // todo handle different error types separately
-  private def convertToStateMachineError(vmError: VmError): StateMachineError =
+  def convertToStateMachineError(vmError: VmError): StateMachineError =
```
folex: Maybe add some kind of logging here? So we know that VM threw an error, and StateMachine took control of it

statemachine/src/main/scala/fluence/statemachine/tx/TxProcessor.scala
DmiSergeev: Currently, all errors occurred during VM invocations are written to the State tree.
However, you are right, there are some errors that would better to pass beyond `TxProcessor` (examples might be fatal errors and some internal errors), rather than write to the State tree.
There exists #155 (Error hierarchy for State machine) that is intended to fix the problem you mentioned. I was not going to include it to this pull request, because it depended on #187, that wasn't merged until last Friday.

statemachine/src/main/scala/fluence/statemachine/tx/TxProcessor.scala
```diff
@@ -87,7 +93,7 @@ class TxProcessor[F[_]](
       // Preparing txs for application. It includes invocation and dequeuing
         .map(tx => {
           for {
```
folex: maybe move inner `for {}` to a separate function?

statemachine/src/main/scala/fluence/statemachine/tx/TxProcessor.scala
DmiSergeev: > in receivedDeliverTx if verifying will be wrong, the client will increment counter and the server will not.
Looks like I need to add more comments about `txCounter`.
Actually, here you are talking about different counters. `txCounter` is global (counting every tx in every session), while the client session has the local session counter that the server also maintains, but implicitly.

DmiSergeev: > what if we raise an error somewhere? Will it be added to the blockchain guaranteed?

Actually, it's assumed that we don't raise the errors during tx processing.

statemachine/src/main/scala/fluence/statemachine/tx/SessionSummary.scala
DmiSergeev: OK, I will review this and similar cases.

statemachine/src/main/scala/fluence/statemachine/ServerRunner.scala
```diff
@@ -66,46 +77,48 @@ object ServerRunner extends IOApp with LazyLogging {
...
    */
-  private[statemachine] def buildAbciHandler(): IO[AbciHandler] =
+  private[statemachine] def buildAbciHandler(): EitherT[IO, StateMachineError, AbciHandler] =
     for {
-      initialState <- MVar[IO].of(TendermintState.initial)
-      stateHolder = new TendermintStateHolder[IO](initialState)
-      mutableConsensusState = new MutableStateTree(stateHolder)
+      config <- EitherT
+        .fromEither[IO](pureconfig.loadConfig[StateMachineConfig])
+        .leftMap[StateMachineError](
```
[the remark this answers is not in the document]
DmiSergeev: Thx! Left this here accidentally.

client/verify.py
```diff
@@ -1,19 +1,37 @@
...
-	return read_json(tmaddress + "/genesis")["result"]["genesis"]
+from misc_utils import parse_utc_unix_ns
+from codec import to_uvarint, l_endian_4b, b64_decode, hex_decode, hex_decode_bytes, hex_encode, hex_encode_bytes, ints_to_bytes
 
 # https://github.com/tendermint/tendermint/blob/master/types/vote.go
 # https://github.com/tendermint/tendermint/blob/master/types/vote_test.go
 def sign_bytes(vote, chain_id):
-	tmpl = '{"@chain_id":"%s","@type":"vote","block_id":{"hash":"%s","parts":{"hash":"%s","total":"%s"}},"height":"%s","round":"%s","timestamp":"%s","type":2}'
-	return str(tmpl \
+	tmpl = '{"@chain_id":"%s","@type":"vote","block_id":{"hash":"%s","parts":{"hash":"%s","total":"%s"}}' + \
```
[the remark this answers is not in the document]
DmiSergeev: It's TMPL – short for 'template'.
Here, we need the order preserving space-less json formation. Unfortunately, it is not very straightforward task in Python. At the time I coded it I thought it would better to write this structure explicitly.
I would try to make it clearer.

client/misc_utils.py
```diff
@@ -1,37 +1,38 @@
...
+    from urllib import urlopen
+else:
+    from urllib.request import urlopen
+
+def parse_utc_unix_ns(timestamp_txt):
+	dt64 = numpy.datetime64(timestamp_txt).astype("datetime64[ns]").astype("int")
+	(t_unix, t_ns) = (dt64 // 1000000000, dt64 % 1000000000)
 	return (t_unix, t_ns)
 
 def read_json(url):
```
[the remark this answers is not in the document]
DmiSergeev: OK, agree.
I will add docs here and in other places.

statemachine/src/main/scala/fluence/statemachine/AbciHandler.scala
```diff
@@ -81,10 +81,11 @@ class AbciHandler(
    * @return `CheckTx` response data
    */
   override def requestCheckTx(req: RequestCheckTx): ResponseCheckTx = {
-    val responseData = validateTx(req.getTx, txParser, checkTxDuplicateChecker).unsafeRunSync()
+    val responseData = validateTx(req.getTx, txParser, checkTxStateChecker).unsafeRunSync()
```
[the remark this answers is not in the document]
DmiSergeev: We are dealing with external Java library here, avoiding `unsafe` might be trickier here.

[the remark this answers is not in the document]
DieMyst: I'm not sure if it is ok if an error will be thrown. If you will use `.timeout().attempt().{maybe some code about error's handling}.unsafeRunSync()` it will be better.
And some docs about errors too.

statemachine/src/main/scala/fluence/statemachine/tx/TxProcessor.scala
```diff
@@ -87,7 +93,7 @@ class TxProcessor[F[_]](
...
-            invoked <- invokeTx(tx).isRight
+            invoked <- invokeTx(tx).map(_.allowDependentTxInvocation)
```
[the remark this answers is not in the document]
DmiSergeev: Good point for moving check to `yield` block, agree.
Why you suggest an explicit comparison with `TransactionStatus.Success`? Maybe `allowDependentTxInvocation` is not the best name, but it won't be better to encapsulate the check from the particular status value, would it?

statemachine/src/main/scala/fluence/statemachine/tx/TxStateDependentChecker.scala
DmiSergeev: > And maybe return Unit here too? You have deduplicatedTx and checkedTx, but they're the same as uncheckedTx, and you even using uncheckedTx.header below. It's kinda confusing.

> maybe return Unit instead of deduplicatedTx here? Code looks like you're changing tx, but you're not.

Maybe I'm wrong but I think that it's better to emphasize the `right` branch of Either-chains. And the `right` branch is `uncheckedTx` -> `deduplicatedTx` -> `checkedTx` where the names express the current 'states' of the tx (despite it remains the same). Otherwise, these transformations are not emphasized and we need to explore the `left` branches to guess what the `right` is.

statemachine/src/main/scala/fluence/statemachine/tx/FunctionCallDescription.scala
```diff
@@ -0,0 +1,85 @@
...
+  private val NonEmptyArgListPattern: Regex = "\\w+(,\\w+)*".r
+
+  /**
+   * Parses text payload in `[moduleName].functionName(arg1, ..., argN)` format to a typed function call description.
+   *
+   * @param payload text representation of the function invocation
+   */
+  def parse[F[_]](payload: String)(implicit F: Monad[F]): EitherT[F, StateMachineError, FunctionCallDescription] =
+    for {
+      parsedPayload <- payload match {
```
alari: It seems like this and the next block could be simplified with plain Either's Left/Right, wrapped with EitherT only once.

PR #205, opened 2018-10-02
Db on pointers

DieMyst:


js-client/src/examples/DbClient.ts
```diff
@@ -0,0 +1,154 @@
...
+                arr.filter((v) => v.nonEmpty)
+                    .map((v) => v.get)
+            );
+    }
+
+
+    private async getRows0<T>(fetch: number, fields: number, parser: (arr: Result[]) => T, acc: Promise<Option<T>>[]): Promise<Promise<Option<T>>[]> {
+        var fetched: [boolean, Promise<Option<T>>[]] = await this.fetchRows(fetch, fields, parser);
+
+        var cont: boolean = fetched[0];
```
DmiSergeev: Using something like `hasMoreRows` here and adding a comment to `fetchRows` return value would be simpler to understand.

js-client/src/examples/DbClient.ts
```diff
@@ -0,0 +1,154 @@
...
+        let lastRow: Promise<Result> = Promise.resolve(empty);
+        for(var _i = 0; _i < fetch; _i++) {
+            let row = this.session.invoke("next_row").result();
+            let res = this.getRow(fields, parser);
+            listOfRows.push(res);
+            lastRow = row;
+        }
+
+        let lastRowResult: Result = await lastRow;
+        if (isValue(lastRowResult) && parseInt(lastRowResult.value) === -1) {
```
DmiSergeev: All promises are ready at the end of the method (because the last one is ready). So here we can return an array of result instead of an array of promises.
Am I right that it's a common practice in JS to work with promises as long as possible?

js-client/src/examples/QuotationClient.ts
```diff
@@ -0,0 +1,155 @@
...
+
+    constructor(host: string, port: number) {
+        this.dbClient = new DbClient(host, port)
+    }
+
+    /**
+     * select avg(field) from Quotations;
+     * ex: select avg(price) from Quotations;
+     */
+    async setAverage(field: Field, fetch: number = 2): Promise<number> {
```
DmiSergeev: `set` prefixes look strange here. Could we get rid of them?

js-client/src/examples/DbClient.ts
DieMyst: `lastRowResult` is not the last one promise. We were request fields after the last row too.

js-client/src/examples/DbClient.ts
DieMyst: ok

PR #206, opened 2018-10-03
String parameters support

michailvoronov:
In current PR:
- added support of finding and using allocate/deallocate WASM internals methods;
- added additional functions for preprocessing supplied arguments that inject a string into Wasm memory and split each string into two arguments: pointer (i32) to the injected string and the string size (i32);
- introduced new error class VmMemoryError that represents all memory errors;
- added some additional logic and additional checks for existed methods to support string passing;
- improved some comments and local variable naming;
- added new test cases for this new functionality.

Also, note that it is assumed that each string parameters wrapped out with double quotes.


vm/src/main/scala/fluence/vm/WasmVm.scala
```diff
@@ -135,21 +135,23 @@ object WasmVm {
       new WasmVmImpl(
```
alari: Probably it's not just a random implementation, but it's an `AsmbleWasmVm`? Class name with `Impl` suffix says nothing about the nature of the implementation, while in this case, it's special.

vm/src/main/scala/fluence/vm/WasmVm.scala
```diff
@@ -135,21 +135,23 @@ object WasmVm {
...
+   * Compiles WASM modules to JVM bytecode and registering derived classes
    * in the Asmble engine. Every WASM module compiles to exactly one JVM class
    */
   private def prepareContext(
     inFiles: Seq[String],
     config: VmConfig
   ): ScriptContext = {
     val invoke = new Invoke()
-    // todo in future should be used common logger for this project
+    // todo in future common logger for this project should be used
```
alari: Let's use `slogger`? It is used in other submodules, and it's very simple to mix in.

vm/src/main/scala/fluence/vm/WasmVm.scala
```diff
@@ -167,8 +169,8 @@ object WasmVm {
    * This method initializes every module and builds a total index for each
    * function of every module. The index is actually a map where the key is a
    * string "Some(moduleName), fnName)" and value is a [[WasmFunction]] instance.
-   * Module name can be "None" if the module name wasn't specified, in this case,
-   * there aren't to be 2 modules without names that contain functions with the
+   * Module name can be "None" if the module name wasn't specified. In this case,
+   * there aren't to be two modules without names that contain functions with the
```
alari: Could you check the language here, please?

vm/src/main/scala/fluence/vm/WasmVmImpl.scala
```diff
@@ -96,22 +103,92 @@ class WasmVmImpl(
       }
       .map(ByteVector(_))
 
+  // TODO : In the future, it should be rewritten with cats.effect.resource
+  def allocate[F[_]: LiftIO: Monad](size: Int): EitherT[F, InvokeError, AnyRef] = {
```
alari: Some docs needed, especially if the function is public. The same holds for `deallocate`.

vm/src/main/scala/fluence/vm/WasmVmImpl.scala
```diff
@@ -96,22 +103,92 @@ class WasmVmImpl(
...
+    allocateFunction match {
+      case Some(fn) => fn(size.asInstanceOf[AnyRef] :: Nil)
+      case _ =>
+        EitherT.leftT(
+          NoSuchFnError(s"Unable to find the function for memory deallocation with the name=$allocateFunctionName")
+        )
+    }
+  }: EitherT[F, InvokeError, AnyRef]
```
alari: Probably if you hint the types in `leftT` call, you'll be able to remove this conversion. I'd better not leave it as it is, as it hides the exact place where the types don't match.

vm/src/main/scala/fluence/vm/WasmVmImpl.scala
```diff
@@ -96,22 +103,92 @@ class WasmVmImpl(
...
+    */
+  private def injectStringIntoWasmModule[F[_]: LiftIO: Monad](
+    injectedString: String,
+    moduleInstance: ModuleInstance
+  ): EitherT[F, InvokeError, AnyRef] = {
+    for {
+      address <- allocate(injectedString.length)
+      _ <- EitherT.rightT(
+        injectedString.getBytes("UTF-8").zipWithIndex.map{
+          case(byte, byteId) => moduleInstance.memory.get.put(address.toString.toInt + byteId, byte)}
```
alari: - Missing space after the `case` keyword
- Closing brace should be on the next line
- For each byte of the injected string, `address.toString.toInt` is called; why not to call it once and reuse the result? BTW: is it safe (can never throw)? Why? Maybe it worth commenting
- With `zipWithIndex`, the list is traversed, a new one is being made, with a counter and a tuple allocated for each element. Probably you need smth like `scan` that traverses the list only once and does not need to allocate anything except a limited intermediary data (and `address.toString.toInt` could be used as a seed)

vm/src/main/scala/fluence/vm/WasmVmImpl.scala
```diff
@@ -149,7 +226,7 @@ object WasmVmImpl {
     Try(castFn(arg).asInstanceOf[AnyRef]).toEither.left
       .map(e ⇒ InvalidArgError(errorMsg, Some(e)))
 
-  /** Safe array concatenation. Prevents  array overflow or OOM. */
+  /** Safe array concatenation. Prevents array overflow or OOM. */
   private def safeConcat[F[_]: LiftIO: Monad](
```
alari: Looks weird. Could you please add a TODO here, in order to check and refactor it later?

vm/src/main/scala/fluence/vm/WasmVmImpl.scala
```diff
@@ -36,49 +36,56 @@ import scala.util.Try
...
 
-      arguments ← parseArguments(fnArgs, wasmFn)
+      preprocessedArguments <- preprocessArguments(fnArgs, wasmFunction.module)
+      arguments ← parseArguments(preprocessedArguments, wasmFunction)
 
       // invoke the function
-      result ← wasmFn[F](arguments)
-    } yield if (wasmFn.javaMethod.getReturnType == Void.TYPE) None else Option(result)
+      result ← wasmFunction[F](arguments)
+      // TODO : In the current version it is expected that callee clean memory by itself, but
```
alari: Could you add some details here, please?

vm/src/main/scala/fluence/vm/WasmVm.scala
michailvoronov: Seems resonable, @RicoGit what do you think?

vm/src/main/scala/fluence/vm/WasmVmImpl.scala
RicoGit: Could you give a detailed explanation?

vm/src/main/scala/fluence/vm/WasmVm.scala
RicoGit: Makes sense.

vm/src/main/scala/fluence/vm/WasmVm.scala
RicoGit: Asmble has own hardcoded logger type. There is no simple solution to change this.

vm/src/main/scala/fluence/vm/WasmVmImpl.scala
```diff
@@ -179,10 +256,14 @@ object WasmVmImpl {
...
-            case cl if cl == classOf[Float] ⇒ cast(arg, _.toFloat, s"Arg $index of '$arg' not a float")
-            case cl if cl == classOf[Double] ⇒ cast(arg, _.toDouble, s"Arg $index of '$arg' not a double")
+            case cl if cl == classOf[Int] ⇒ cast(arg, _.toInt,
+              s"Arg $index of '$arg' not an int (or passed string argument isn't match corresponding argument in wasm function)")
+            case cl if cl == classOf[Long] ⇒ cast(arg, _.toLong,
+              s"Arg $index of '$arg' not a long (or passed string argument isn't match corresponding argument in wasm function)")
+            case cl if cl == classOf[Float] ⇒ cast(arg, _.toFloat,
+              s"Arg $index of '$arg' not a float (or passed string argument isn't match corresponding argument in wasm function)")
+            case cl if cl == classOf[Double] ⇒ cast(arg, _.toDouble,
+              s"Arg $index of '$arg' not a double (or passed string argument isn't match corresponding argument in wasm function)")
```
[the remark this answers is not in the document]
michailvoronov: Lets consider the next situation when user supplied (string, int, float) arguments to Wasm function that actually receives (int, int, float). preprocessArguments transform this arguments to (int, int, int, float) and this discrepancy will be revealed in the float case.

vm/src/main/scala/fluence/vm/WasmVmImpl.scala
```diff
@@ -96,22 +103,92 @@ class WasmVmImpl(
...
+    * @param moduleInstance module instance used for injecting string arguments to the WASM memory
+    * @tparam F a monad with an ability to absorb 'IO'
+    */
+  private def preprocessArguments[F[_]: LiftIO: Monad](
+     functionArguments: Seq[String],
+     moduleInstance: ModuleInstance
+   ) : EitherT[F, InvokeError, List[String]] = {
+    val ret: Seq[EitherT[F, InvokeError, List[String]]] =
+      functionArguments.map {
+        case stringArgument if stringArgument.startsWith("\"")
```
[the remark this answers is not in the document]
michailvoronov: I've thought about it, but in my opinion, it's could led to some potential error that can be difficult to detect. F.e. if someone could use quotes inside a string and forget start quote this string would be valid but error potential challenging to identify. I think that this strict rule for this question is more suitable than flexible.

vm/src/test/scala/fluence/vm/WasmVmImplSpec.scala
```diff
@@ -228,6 +228,24 @@ class WasmVmImplSpec extends WordSpec with Matchers {
...
+          state ← vm.getVmState[IO].toVmError
+        } yield {
+          value1 shouldBe Some(90)
+          value2 shouldBe Some(0)
+          value3 shouldBe Some('X'.toInt)
+        }
+
+        res.success()
+      }
+
```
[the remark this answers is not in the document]
michailvoronov: done

vm/src/main/scala/fluence/vm/WasmVm.scala
michailvoronov: done

vm/src/test/scala/fluence/vm/AsmleWasmVmSpec.scala
```diff
@@ -228,6 +228,88 @@ class WasmVmImplSpec extends WordSpec with Matchers {
...
+            val error = res.failed()
+            error.getMessage shouldBe
+              "Invalid number of arguments, expected=2, actually=1 for fn='<no-name>.circular_xor' " +
+              "(or passed string argument is incorrect or isn't matched the corresponding argument in Wasm function)"
+          }
+
+        }
+      }
+
+      "string passing" should {
```
RicoGit: Use nested declaration
```
 "string passing" should {
   "raise an error" when {
      "trying to pass incorrect string" in { .. }
      "trying to use Wasm memory when getMemory function isn't defined" in { .. }
      "Wasm allocate function returns an incorrect value" in { .. }
      "Wasm allocate function returns an incorrect value" in { .. }
 } 
}
```
Thank you for tests, mind you!

vm/src/main/scala/fluence/vm/WasmVm.scala
alari: `there aren't to be two modules` -- it still looks incorrect.

PR #217, opened 2018-10-08
Fix getting Wasm memory

RicoGit:


vm/src/main/scala/fluence/vm/ModuleInstance.scala
```diff
@@ -57,14 +57,16 @@ case class ModuleInstance(
...
                 // Duplicate is required for reaching idempotent reading ByteBuffer(BB).
                 // ''ByteBuffer.get'' change inner BB state, for preventing this
                 // we create a thin copy of this BB. The new buffer's capacity,
                 // limit, position, and mark values will be identical to those of
                 // original buffer, but the content(bytes) will be shared (bytes won't be copied).
                 // After reading all bytes, duplicate will be collected by GC
-                mem.duplicate().get(arr, 0, arr.length)
+                val view = mem.duplicate()
+                view.clear()
+                val arr = new Array[Byte](view.capacity())
```
michailvoronov: Why not limit?

vm/src/main/scala/fluence/vm/ModuleInstance.scala
```diff
@@ -57,14 +57,16 @@ case class ModuleInstance(
...
+                view.get(arr, 0, arr.length)
```
michailvoronov: view.get(arr) is equals to view.get(arr, 0, arr.length) but shorter and more readable because of the second and third params can be perceived as an offset in ByteBuffer.

RicoGit: Yes, will use 'get(arr)', thanks!

vm/src/main/scala/fluence/vm/ModuleInstance.scala
RicoGit: Let's discuss this topic, it's the difficult question, 'limit' might be wrong in some cases, but 'capacity' is correct in general, but might be inefficient.

PR #221, opened 2018-10-09
Fix vm counter

RicoGit:
* fix vm-counter example
* remove vm-sqldb example


vm/examples/counter/src/main/rust/memory_manager.rs
```diff
@@ -0,0 +1,45 @@
...
+
+    let result_ptr = alloc(result.len()).unwrap().as_ptr();
+
+    // writes bytes into memory byte-by-byte. Address of first byte will be == `ptr`
+    for (idx, byte) in result.iter().enumerate() {
+        std::ptr::write(result_ptr.offset(idx as isize), *byte);
+    }
+
+    result_ptr
+}
```
michailvoronov: EOF

vm/examples/counter/src/main/rust/lib.rs
```diff
@@ -10,6 +22,19 @@ pub unsafe fn inc() {
...
+pub unsafe fn allocate(size: usize) -> NonNull<u8> {
+    alloc(size)
+        .expect(format!("[Error] Allocation of {} bytes failed.", size).as_str())
 }
+
+#[no_mangle]
+pub unsafe fn deallocate(ptr: NonNull<u8>, size: usize) -> () {
+    dealloc(ptr, size)
+        .expect(format!("[Error] Deallocate failed for prt={:?} size={}.", ptr, size).as_str())
+}
```
michailvoronov: EOF

vm/examples/counter/src/main/rust/memory_manager.rs
```diff
@@ -0,0 +1,45 @@
...
+use std::ptr::NonNull;
+use std::mem;
+use std::io::Write;
+
+pub const STR_LEN_BYTES: usize = 4;
+
+/// Result for all possible Error types.
+pub type GenResult<T> = Result<T, Box<Error>>;
+
+/// Allocates memory area of specified size and returns address of the first
```
michailvoronov: It seems that "... and returns its address" sounds less complicated because of what other address can return from the function with name alloc and that has such comment other than that?

Issue #195, opened 2018-09-26
Tendermint / State machine Process manager


DmiSergeev:
### Motivation
Fluence network is to be combined from many clusters. The key feature we lack now is to enable nodes to take part in different clusters. In order to support this feature, the process management is required: it includes cluster nodes’ creation, monitoring, and shutdown. The Process manager would allow launching a new Tendermints / State machines and tracking their lifecycle.

### Current state
Currently, a single local cluster setup is only supported, it is performed through several bash scripts, was developed and used for testing purposes only.

### Proposition
By _Solver worker_ a single `TM Core instance + Fluence State machine` combination is meant. Those processes should be run in Docker containers with all required settings being configurable.

The Process manager must provide the following functionality:
* to launch Solver workers by command
* to monitor launching Solver workers: track the processes’ health, access their metrics; to expose the monitoring data outside
* to shut Solver worker down by command

Launching the Solver worker implies starting a single Docker image that should:
* take the prearranged configuration settings/files and VM module files passed as instance(s)' parameters
* create necessary home directories with the prearranged files put there
* start TM Core process on a requested port
* start ABCI State machine process with required parameters

PR #222, opened 2018-10-09
Run Docker container with healthchecks

alari:
Going toward #195:
- Master node process
- Launch docker container, `rm -f` when it's finished
- Check container's liveness periodically
- Ping container with HTTP for health checking, if it's alive


node/src/main/scala/fluence/node/FluenceNode.scala
```diff
@@ -17,7 +17,44 @@
...
+      .run[IO]("-p 9393:80 nginx")
+      .through(
+        // Check that container is running every 3 seconds
+        DockerIO.check[IO](3.seconds)
+      )
+      .evalMap[IO, (FiniteDuration, Either[String, Unit])] {
+        case (d, true) ⇒
+          // As container is running, perform a custom healthcheck: request a HTTP endpoint inside the container
+          sttp
+            .get(uri"http://localhost:9393")
```
DieMyst: is hard-coded port and host ok?

alari: It's as good as using `nginx` image is ;)

Checks should become more sophisticated during our integration with statemachine containers. This code is just to highlight the possible flow: get checks for the container being run, then call any domain-specific IO-wrapped check.

One of the problems this design aims to tackle is that it's not that easy to understand what the "health" is for the statemachine process. At least we have three levels:
- Is container running?
- Is the tendermint process available?
- Is there progress in the cluster?

The first check could be solved once for all containers. The others are unclear both in terms of checking and the right reaction.

PR #224, opened 2018-10-10
Scripts for node master

DmiSergeev:
* `docker/master-*.sh` scripts implement master process logic:
  * creating reusable (long-term) node keys
  * extracting public keys from long-term keys directory
  * running a node with all configuration provided
* `docker/contract-combine-nodes.sh` implements a logic to combine multiple nodes' keys to a single JSON with `genesis` and `persistent_peers` components
* `docker/sim-*.sh` scripts simulate an integration between Master and Contract

Examples run
* `docker/example/*` scripts provide running example clusters – to run a cluster, one needs:
  * to change directory to `docker/example`
  * to generate keys (if this was not done before) via `sim-init-long-term-keys.sh`
  * to simulate launching multi-node cluster via `sim-cluster-*.sh`


client/example/dataengine-example.py
```diff
@@ -56,23 +56,23 @@ def demo_many_queries(addr, genesis):
...
 
-tmport = sys.argv[1] if len(sys.argv) >= 2 else "46157"
+tmport = sys.argv[1] if len(sys.argv) >= 2 else "25057"
 tm = TendermintRPC("http://localhost:" + tmport)
 genesis = tm.get_genesis()
 height = tm.get_max_height()
 
 # 1st session: correct, but not explicitly closed
-session1_id = id_generator()
+session1_id = "157A0E"#id_generator()
```
DieMyst: is it debug code?

client/example/dataengine-example.py
```diff
@@ -56,23 +56,23 @@ def demo_many_queries(addr, genesis):
...
 # 3rd session: correct and explicitly closed
 # 1st session expires during 3rd session processing
-demo_queries(tm, genesis, False, True)
+#demo_queries(tm, genesis, False, True)
 
 # 4th session: same as 1st - transactions declined as duplicated
-demo_queries(tm, genesis, False, False, session1_id)
+#demo_queries(tm, genesis, False, False, session1_id)
 
 #demo_many_queries(tm, genesis)
```
DieMyst: eof

docker/contract-combine-nodes.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+# param
+# $1 cluster_name
+# $2, $3, $4, $5 nodes' keys
+
```
folex: it is better to always check number of arguments passed. e.g.
```if [ "$#" -ne 1 ]; then
    echo "Illegal number of parameters"
fi```

File: docker/contract-combine-nodes.sh [hunk 4]
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
```
folex: It is a good style to use `#!/bin/bash -e`, so script stops on the very first non-zero exit code

docker/contract-combine-nodes.sh
```diff
@@ -0,0 +1,25 @@
...
+# param
+# $1 cluster_name
+# $2, $3, $4, $5 nodes' keys
+
+# iterate through given node public key JSONs, combine genesis info and persistent peers
+for ((i = 2; i <= $#; i++)); do
+    node_name="node"$(($i-2))
+
+    validators=$validators"{\"pub_key\":$(echo ${!i} | jq .validator),\"power\":\"1\",\"name\":\"$node_name\"},"
```
folex: what does `${!i} | jq .validator` mean?

folex: and can you please move that to a separate variable? complex inline expressions are hard to read and reason about

folex: Also, you can use heredoc to escape escaping hell (pun intended). E.g.:
```
new_validator=$(cat <<EOF
{
    "pub_key": "$(echo ${!i} | jq .validator)",
    "power": "1",
    "name": "$node_name"
},
EOF
)
```

docker/statemachine/run-node.sh
```diff
@@ -1,11 +1,18 @@
...
+# $3 state_machine_target_path
 
 # uncomment line below if attaching to container is needed before node running
 #sleep 8
 
 # copy predefined Tendermint config
 cp -f $2/config.toml $1/config/config.toml
 
+# set advertised address
+sed -i -e "s#external_address.*=.*\$#external_address = \"$(cat $1/config/external_addr.txt)\"#g" $1/config/config.toml
```
folex: - use `^` and `$` in regex to match whole line
- extract inline expression `cat $1/config/external_addr.txt` to a separate variable
- `.*` is greedy, it can much more than just a single space and behave surprisingly. Use non greedy `.*?` or even more specific ` *`

docker/statemachine/run-node.sh
```diff
@@ -1,20 +1,26 @@
...
 # $3 state_machine_target_path
 
+if [ "$#" -ne 3 ]; then
+    echo "Illegal number of parameters: 3 required"
+    exit 1
+fi
+
 # uncomment line below if attaching to container is needed before node running
-#sleep 8
+sleep 8
```
folex: That should always be uncommented? Maybe add a comment on it mentioning that it is intended to be enabled not just for debugging purposes

PR #227, opened 2018-10-13
Switch to byte array parameter

michailvoronov:
Switch to byte array parameter:
  - changed invoke signature from Seq[String] to Option[Array[Byte]];
  - removed some logic for argument parsing;
  - corrected some tests to correspond new signature;
  - added some new tests;
  - changed argument parsing on statemachine part.


statemachine/src/main/scala/fluence/statemachine/util/HexCodec.scala
```diff
@@ -47,4 +49,12 @@ object HexCodec {
...
   def hexToString(hex: String): Either[String, String] =
     ByteVector.fromHexDescriptive(hex, Bases.Alphabets.HexUppercase).map(x => new String(x.toArray))
+
+  /**
+   * Decodes a string from hex representation to array of bytes.
+   *
+   * @param hex string in hex representation
+   * @throws NumberFormatException if incorrect hex string supplied
+   */
+  def hexToArray(hex: String): Array[Byte] = new BigInteger(hex, 16).toByteArray
```
DmiSergeev: Implementing it as `ByteVector.fromHexDescriptive(hex, Bases.Alphabets.HexUppercase).map(x => x.toArray)` thus returning `Either` would allow to avoid using `Try` and instantiating a `BigInteger` object.

statemachine/src/main/scala/fluence/statemachine/tx/FunctionCallDescription.scala
```diff
@@ -30,14 +31,14 @@ import scala.util.matching.Regex
  * @param functionName name of the invoked function
  * @param argList ordered sequence of function arguments
  */
-case class FunctionCallDescription(module: Option[String], functionName: String, argList: List[String])
+case class FunctionCallDescription(module: Option[String], functionName: String, argList: Option[Array[Byte]])
```
DmiSergeev: Why do we need `Option` here? What the difference between `None` and `Some(Array())`?

vm/src/main/scala/fluence/vm/AsmleWasmVm.scala
```diff
@@ -162,52 +167,35 @@ class AsmleWasmVm(
...
-              charset
-            )
-          } yield List(address.toString, (stringArgument.length - 2).toString)
-        case arg ⇒ EitherT.rightT[F, InvokeError](List(arg))
-      }
-
-    import cats.instances.list._
-    Traverse[List].flatSequence(result.toList)
-  }
+  ): EitherT[F, InvokeError, List[AnyRef]] =
```
DmiSergeev: Maybe for the Either's right side we can replace `List[AnyRef]` with a stricter type?
It might be something like `Option[(Int, Int)]`.

statemachine/src/main/scala/fluence/statemachine/util/HexCodec.scala
michailvoronov: Yes, thank you! Your variant is more faster. Done.

statemachine/src/main/scala/fluence/statemachine/tx/FunctionCallDescription.scala
michailvoronov: Since it is planned to switch to "byte array pipe" between client and server I changes the sigmature of Wasm invoke method from Option[Array[Byte]] to Array[Byte]. So now all public for clients Wasm methods expect pointer to injected byte array and its size.
Done.

vm/src/main/scala/fluence/vm/AsmleWasmVm.scala
michailvoronov: It is an interesting question because of invoke method receives List[AnyRef]. So it needs to additionally convert from Tuple(Int, Int) to List[AnyRef]. And I would prefer case class with two fields (address and size) for code readability. 
But now invoke method is used only for exported and allocate/deallocate functions that all receive Ints parameters. However, in the future, this could be changed (f.e. after switching to 64-bit address space both Int and Long can be passed). 
Furthermore, this function is private, and I suggest at the moment to leave it as is. But it is possible to refactor it with allocate/deallocate after the verification game process has been wholly researched (it seems that it could be refactored with cats.effect.resource).

statemachine/src/main/scala/fluence/statemachine/tx/FunctionCallDescription.scala
```diff
@@ -19,25 +19,25 @@ import cats.Monad
...
-case class FunctionCallDescription(module: Option[String], functionName: String, argList: List[String])
+case class FunctionCallDescription(module: Option[String], functionName: String, arg: Array[Byte])
 
 object FunctionCallDescription {
 
   /**
    * Description for reserved non-VM function call that explicitly closes sessions by the client.
    */
-  val CloseSession = FunctionCallDescription(None, "@closeSession", Nil)
+  val CloseSession = FunctionCallDescription(None, "@closeSession", Array[Byte]())
```
RicoGit: `Array.emptyByteArray()` returns singleton instance without any allocations.

statemachine/src/main/scala/fluence/statemachine/tx/FunctionCallDescription.scala
```diff
@@ -60,18 +57,13 @@ object FunctionCallDescription {
...
-          }
-        }
+      (module, functionName, unparsedArg) = parsedPayload
 
-      }
-    } yield FunctionCallDescription(module, functionName, parsedArgList))
+      // each public Wasm function receives byte array - so returns an empty array in case of empty argument
+      parsedArg <- if (unparsedArg.isEmpty) Either.right(Array[Byte]())
+      else hexToArray(unparsedArg).left.map(_ => wrongPayloadArgument(unparsedArg))
+
```
RicoGit: Are you apply auto-formatting? This expression looks bad.

statemachine/src/main/scala/fluence/statemachine/tx/FunctionCallDescription.scala
```diff
@@ -85,8 +77,8 @@ object FunctionCallDescription {
    * Produces [[StateMachineError]] corresponding to payload's argument list that cannot be parsed to
    * correct function arguments.
    *
-   * @param unparsedArgList wrong payload argument list
+   * @param unparsedArg wrong payload argument
    */
-  private def wrongPayloadArgumentListFormatError(unparsedArgList: String): StateMachineError =
-    PayloadParseError("WrongPayloadArgumentListFormat", s"Wrong payload arguments: $unparsedArgList")
+  private def wrongPayloadArgument(unparsedArg: String): StateMachineError =
```
RicoGit: Name of this method tells nothing about its function. What is it actually doing? This can be understood only by reading the body of the method. Lets add some verb to method name, for example:
- wrapToError(unpasedArg)
- createPayloadParseError(unpasedArg)
- toParseError(unpasedArg)
and so on

statemachine/src/main/scala/fluence/statemachine/util/HexCodec.scala
```diff
@@ -47,4 +47,12 @@ object HexCodec {
...
   def hexToString(hex: String): Either[String, String] =
     ByteVector.fromHexDescriptive(hex, Bases.Alphabets.HexUppercase).map(x => new String(x.toArray))
+
+  /**
+   * Decodes a string from hex representation to array of bytes.
+   *
+   * @param hex string in hex representation
+   */
+  def hexToArray(hex: String): Either[String, Array[Byte]] =
+    ByteVector.fromHexDescriptive(hex, Bases.Alphabets.HexUppercase).map(x => x.toArray)
```
RicoGit: `map(x => x.toArray)` >>> `map(_.toArray)`
less noise, IMHO. But up to you.

vm/src/main/scala/fluence/vm/AsmleWasmVm.scala
```diff
@@ -162,52 +153,31 @@ class AsmleWasmVm(
...
+    for {
+      offset <- injectArrayIntoWasmModule(fnArgument, moduleInstance)
+    } yield offset.asInstanceOf[AnyRef] :: fnArgument.length.asInstanceOf[AnyRef] :: Nil
```
RicoGit: This method only calls another method and looks redundant. Its name 'preprocessFnArgument' is very generic and hide under the hood the method name 'injectArrayIntoWasmModule' which is more important(it's the essence of what is happening). I would prefer to delete 'preprocessFnArgument' and call direct 'injectArrayIntoWasmModule'.
What do you think?

vm/src/main/scala/fluence/vm/AsmleWasmVm.scala
```diff
@@ -316,114 +286,28 @@ object AsmleWasmVm {
...
-                  s"Invalid type for param $index: $paramType, expected [i32|i64|f32|f64]"
-                )
-              )
-          }
-      }
-
-      result ← list2Either[F, InvalidArgError, AnyRef](args.toList)
-
-    } yield result
-
```
RicoGit: So now we should implement the stuff like this on the Wasm side)

vm/src/main/scala/fluence/vm/WasmVm.scala
```diff
@@ -53,17 +53,15 @@ trait WasmVm {
...
+   * @param fnName a Function name to invocation
+   * @param fnArgument a Function arguments
    * @tparam F a monad with an ability to absorb 'IO'
    */
   def invoke[F[_]: LiftIO: Monad](
     module: Option[String],
-    function: String,
-    fnArgs: Seq[String] = Nil
+    fnName: String,
+    fnArgument: Array[Byte] = Array[Byte]()
```
RicoGit: Array.emptyByteArray()

statemachine/src/main/scala/fluence/statemachine/tx/FunctionCallDescription.scala
michailvoronov: Thank!

statemachine/src/main/scala/fluence/statemachine/tx/FunctionCallDescription.scala
michailvoronov: I think that wrongPayloadArgumentFormatError (+ FormatError to the current variant) is more near to existing error style in statemachine.

statemachine/src/main/scala/fluence/statemachine/util/HexCodec.scala
michailvoronov: Done.

vm/src/main/scala/fluence/vm/AsmleWasmVm.scala
michailvoronov: Yes, this method comes from the previous case when invoke receives Option[Array[Byte]], the readability of invoke method was too bad and i put away matching of this option.

vm/src/main/scala/fluence/vm/WasmVm.scala
michailvoronov: Done.

vm/src/main/scala/fluence/vm/AsmleWasmVm.scala
michailvoronov: Yes, there is a lot of work on "byte array pipe" on the Wasm side:).

statemachine/src/main/scala/fluence/statemachine/tx/FunctionCallDescription.scala
RicoGit: @DmiSergeev What do you think?

DmiSergeev: @RicoGit I will refactor/rename statemachine errors in the nearest future, during working on #155 .

PR #234, opened 2018-10-16
Cluster scripts: llamadb launcher, jq erasing, port fixes

DmiSergeev:
* llamaDB launcher
* no `jq` need for running local clusters
* port fixes


statemachine/docker/examples/sim-cluster-sqldb.sh
```diff
@@ -3,3 +3,4 @@ pushd .
 cd ..
 ./sim-cluster.sh sqldbnet "$PWD/examples/vmcode-sqldb" 27057 $HOME/.fluence/long-term-keys
 popd
+docker attach sqldbnet_node3
```
alari: Shouldn't we remove the old sqldb on pointers completely?

statemachine/docker/master-show-node-keys.sh
```diff
@@ -11,4 +11,4 @@ node_id=$(./master-run-tm-utility.sh tm-show-node-id "$1")
 validator=$(./master-run-tm-utility.sh tm-show-validator "$1")
 
 # create a JSON document with public keys
```
alari: Looks like it's not a JSON anymore.

statemachine/docker/sim-cluster.sh
```diff
@@ -2,7 +2,7 @@
 # param
 # $1 docker_network_name / cluster_name
 # $2 vm_code_directory
-# $3 host_rpc_port
+# $3 host_base_rpc_port
```
alari: Is it possible to add any docs for such parameters? How it affects the launch? How should I choose this port?

statemachine/docker/examples/sim-cluster-llamadb.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash -e
+pushd .
+cd ..
+./sim-cluster.sh llamadbnet "$PWD/examples/vmcode-llamadb" 29057 $HOME/.fluence/long-term-keys
+popd
+docker attach llamadbnet_node3
```
DieMyst: it is better to use `docker logs -f` because with attaching if you try to exit from docker with `Ctrl+C` the container will be stopped.

DmiSergeev: Great idea, thanks!

PR #238, opened 2018-10-17
Llamadb tests 

RicoGit:
Added test cases and usage examples for DELETE and TRUNCATE statements.

### DIFF
--- a/vm/examples/llamadb/src/main/rust/lib.rs
+++ b/vm/examples/llamadb/src/main/rust/lib.rs
@@ -94,9 +94,8 @@ unsafe fn deref_str(ptr: *mut u8, len: usize) -> String {
 
 /// Acquires lock, does query, releases lock, returns query result
 fn run_query(sql_query: &str) -> GenResult<String> {
-    let statement = llamadb::sqlsyntax::parse_statement(sql_query)?;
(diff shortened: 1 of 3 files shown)


vm/examples/llamadb/src/main/rust/lib.rs
```diff
@@ -135,16 +137,14 @@ fn statement_to_string(statement: ExecuteStatementResponse) -> String {
 unsafe fn put_to_mem(str: String) -> *mut u8 {
     // converting string size to bytes in little-endian order
     let len_as_bytes: [u8; STR_LEN_BYTES] = mem::transmute((str.len() as u32).to_le());
-    let mut result: Vec<u8> = Vec::with_capacity(STR_LEN_BYTES + str.len());
+    let total_len = STR_LEN_BYTES + str.len();
```
michailvoronov: It seems that it is possible for to do int overflow here in some special condition: minimum 4Gb memory is needed and some vector handling logic has to be changed (f.e. suppling not total_len to std::ptr::copy but str.len() or so on). So now it is impossible to doing something bad with this but these conditions can be satisfy, so i suggest to add some comment about it (to prevent such situations in future) or to add some overflow checks.

PR #239, opened 2018-10-17
Todo list with delete

DieMyst:
Adds a delete functionality to the `todo` app.


DieMyst:
@alari all is updated for now

PR #248, opened 2018-10-23
Cluster metric collection, state machine tests restore, Tendermint 0.25

DmiSergeev:
* Tendermint updated to 0.25.0
* Timestamp added to tx json to measure latencies
* Metric collection for State machine with Prometheus-compatible metrics feed
* Metrics for query times and latencies
* Docker-compose for launching Prometheus and Grafana with State machine monitoring dashboard
* Python client restored and moved to `py-client`
* State machine tests restored (but they are still too coupled with VM)


py-client/dataengine.py
```diff
@@ -87,7 +127,7 @@ def result(self, requests_per_sec = 2, response_timeout_sec = 10, wait_before_re
 		"""
 		tm = self.session.engine.tm
 		path = self.target_key + "/result"
-		print("querying " + path)
+		#print("querying " + path)
```
DieMyst: is it ok?

py-client/dataengine.py
```diff
@@ -131,21 +171,23 @@ def __init__(self, engine, client, signing_key, session):
 		self.summary_key = "@meta/%s/%s/@sessionSummary" % (self.client, self.session)
 		self.counter = 0
 
-	def submit(self, command, *params):
+	def submit(self, command, arg):
```
DieMyst: arg => args?

statemachine/src/main/scala/fluence/statemachine/tx/TxProcessor.scala
```diff
@@ -117,7 +117,7 @@ class TxProcessor[F[_]](
       result <- payloadOp match {
         case None => F.pure(Nil)
         case Some(payload) =>
-          txAndQueuedDependencies(txHeader.dependentTxHeader).map(Transaction(txHeader, payload) :: _)
+          txAndQueuedDependencies(txHeader.dependentTxHeader).map(Transaction(txHeader, payload, "") :: _)
```
DieMyst: as I see - `timestamp` argument is not used here

statemachine/src/main/scala/fluence/statemachine/AbciHandler.scala
```diff
@@ -128,18 +142,35 @@ class AbciHandler(
...
+      validated <- (for {
+        parsedTx <- txParser.parseTx(txBytes)
+        checkedTx <- txStateChecker.check(parsedTx)
+      } yield checkedTx).value.map {
+        case Left(message) => TxResponseData(None, CodeType.BAD, message)
+        case Right(tx) => TxResponseData(Some(tx), CodeType.OK, ClientInfoMessages.SuccessfulTxResponse)
+      }
+      validationDuration = validationTimeMeter.millisElapsed
+      latency = validated.validatedTx
+        .map(tx => TimeMeter.millisFromPastToNow(tx.timestamp.toLong) - validationDuration)
```
DieMyst: `toLong` can throw an error here, it is better to use `Long` in the `Transaction` class

statemachine/src/main/scala/fluence/statemachine/util/Crypto.scala
```diff
@@ -50,20 +50,47 @@ object Crypto extends slogging.LazyLogging {
...
-      engine2 = new EdDSAEngine()
-      _ <- Try(engine.initVerify(key))
-      _ <- Try(engine2.initVerify(key))
-    } yield engine.verifyOneShot(hashed, signatureBytes) || engine2.verifyOneShot(dataBytes, signatureBytes)
+      _ <- Try(engine.initVerify(publicKey))
+      verified <- Try(engine.verifyOneShot(hashed, signatureBytes))
+    } yield verified
     verificationPassed.failed.foreach { e =>
-      logger.error("An error on verifying signature: " + e.getMessage)
+      logger.error("An error on verifying signature: {}", e.getMessage)
```
DieMyst: maybe we will use string interpolator 
```
logger.error(s"An error on verifying signature: ${e.getMessage}")

```

statemachine/src/main/scala/fluence/statemachine/util/Crypto.scala
```diff
@@ -50,20 +50,47 @@ object Crypto extends slogging.LazyLogging {
...
   }
 
+  /**
+   * Signs `data` using a given EdDSA-25519 `privateKeyBase64`.
+   *
+   * @param data text input
+   * @return text Base-64 signature
+   */
+  def sign(data: String, privateKeyBase64: String): String = {
+    val signatureTry = for {
```
DieMyst: You can wrap the whole function into Try {...}, it will be clearer

tools/monitoring/grafana/provisioning/datasources/prometheus.json
```diff
@@ -0,0 +1,19 @@
...
+  "user": "",
+  "database": "",
+  "basicAuth": false,
+  "isDefault": true,
+  "jsonData": {
+    "httpMethod": "GET",
+    "keepCookies": []
+  },
+  "readOnly": false
+}
```
DieMyst: eof

statemachine/docker/examples/vmcode-counter/mul.wast
```diff
@@ -1,7 +1,149 @@
-;; equivalent to `int mul(int a, int b) { return a * b; }` in C++
+;; this example simply returns product of two integers
```
michailvoronov: Maybe it is better to copy counter.wast and mul.wat files in this directory (with corresponding .gitignore update) by appropriate bash script before docker running?

statemachine/src/test/scala/fluence/statemachine/IntegrationSpec.scala
```diff
@@ -177,43 +179,41 @@ class IntegrationSpec extends WordSpec with Matchers with OneInstancePerTest {
...
       sendCommit()
-      latestAppHash shouldBe "E6FC72DA9F8296F9549105711EF10F15C598BD8162976577BA00B3E1FB3AA758"
 
       val txWithWrongSignature = tx(client, session, 0, "inc()", "bad_signature")
       sendCheckTx(txWithWrongSignature) shouldBe (CodeType.BAD, ClientInfoMessages.InvalidSignature)
       sendDeliverTx(txWithWrongSignature) shouldBe (CodeType.BAD, ClientInfoMessages.InvalidSignature)
 
       sendCommit()
-      latestAppHash shouldBe "E6FC72DA9F8296F9549105711EF10F15C598BD8162976577BA00B3E1FB3AA758"
+      latestAppHash shouldBe "CED687708BB077AFFF670943B503E30035E1FA4B56535B77A204BF80F5E51C3B"
```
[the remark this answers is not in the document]
DmiSergeev: Yes, there is a problem with all these hashes. Therefore I made a TODO note in the same test.
Currently, this test is the only test covering State machine logic. It is quite useful when someone is making refactoring/improvements in the State machine code without changing the logic. Failing test might indicate the presence of bugs in this situation, so refactoring procedure becomes safer.
However, I agree that maintainers of VM logic might suffer when even minor changes cause failing tests here.
That's why we need to decompose current tests into 2 groups:
* tests for State machine: they might use some Stub VM and fix such hashes (probably they even do not need to assert all hashes – at least some of them)
* tests for SM - VM integration: just few tests, looks like we don't need hash asserts here, just logic asserts.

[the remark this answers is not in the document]
michailvoronov: Yes, sounds reasonable. So what about variants with 1-2 test cases with TODO both for SM and VM maintainers? (this should affect the VM part in the nearest future with adding incremental hashing, shadow stack and so on)
Also I suggest to add additional comment to the defaultMaxMemPages fied in vm/src/main/resources/reference.conf because of each changes of memory pages count lead to the app hash changes.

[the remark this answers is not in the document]
michailvoronov: @rico, what do you think?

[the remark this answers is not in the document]
RicoGit: I think we should leave the assertion for the hash only with a stub VM.

py-client/dataengine.py
DmiSergeev: fixed

PR #252, opened 2018-10-23
Add Drop statement and TEXT type into examples

RicoGit:


vm/examples/llamadb/src/main/scala/fluence/vm/examples/LlamaDbRunner.scala
```diff
@@ -19,98 +19,83 @@ package fluence.vm.examples
...
+        "UPDATE Roles r SET r.role = 'Professor' WHERE r.user_id = " +
+          "(SELECT id FROM Users WHERE name = 'Sara')"
+      )
 
-      truncateQuery = "TRUNCATE TABLE Users"
-      truncateQueryRes ← vm.invoke[IO](None, "do_query", truncateQuery.getBytes())
-      truncateQueryState ← vm.getVmState[IO]
+      truncateQuery ← executeSql(vm, "TRUNCATE TABLE Users")
+
+      dropTable ← executeSql(vm, "DROP TABLE Users")
```
michailvoronov: There are two types of naming here:
- ends with "Query": deleteQuery, UpdateQuery, ...
- doesn't end with "Query"
So maybe it is better to lead one of this variant? For my opinion the first variant is better but It is too minor and up to you.

PR #270, opened 2018-10-31
Avoid calling allocation function from Wasm with null size

michailvoronov:


vm/src/main/scala/fluence/vm/AsmleWasmVm.scala
```diff
@@ -152,6 +152,26 @@ class AsmleWasmVm(
     }
   }
 
+  /**
+   * Preprocesses a Wasm function argument: injects it into Wasm module memory (through injectArrayIntoWasmModule)
+   * and replace with pointer to it in the Wasm module and size. This functions returns 0 :: 0 :: Nil
```
RicoGit: replace -> replaces

vm/src/main/scala/fluence/vm/AsmleWasmVm.scala
```diff
@@ -152,6 +152,26 @@ class AsmleWasmVm(
...
+   * if supplied fnArgument was empty.
```
RicoGit: Let's note in the doc that allocation will be skipped for an empty array.

PR #275, opened 2018-11-02
[WIP] Range merkle proof implementation

folex:


docs/protocol/merkle_proof.go
```diff
@@ -0,0 +1,327 @@
...
+  Hash     Digest
+  Children []Node
+  Parent   *Node
+}
+
+// Merkle Proof for a range of chunks
+type RangeMerkleProof struct {
+  Chunks     []Chunk
+  Hashes     [][2]Digest
+  StartChunk int32
```
xdralex: Don't need these in the proof: start chunk index can be derived as a binary encoding of `Hashes`. Use 1 when the `Hashes[k][0]` is present, 0 when it doesn't so for `Hashes = {{h1, ...}, {h2, ...}, {nil, ...}, {h3, ...}}` the start chunk index will be `1011b = 11`. Stop chunk index can be derived as `StartChunk + len(Chunks) – 1`.

docs/protocol/merkle_proof.go
```diff
@@ -0,0 +1,327 @@
...
+  "math"
+)
+
+// Merkle Tree representation
+type MerkleTree struct {
+  Root   Node
+  Height uint32
+}
+
+func (t MerkleTree) String() string {
```
xdralex: We probably don't need string dumps or other util stuff in the protocol core to avoid too much clutter (after all, it's meant to be read as a pseudocode).

docs/protocol/merkle_proof.go
```diff
@@ -0,0 +1,327 @@
...
+type Node struct {
+  Hash     Digest
+  Children []Node
+  Parent   *Node
+}
+
+// Merkle Proof for a range of chunks
+type RangeMerkleProof struct {
+  Chunks     []Chunk
+  Hashes     [][2]Digest
```
xdralex: Need to be able to specify the digest which is not present. 

Probably by using `Hashes [][2]*Digest` or something like 
```go
type DigestOpt *Digest
...
Hashes[][]DigestOpt
```

docs/protocol/merkle_proof.go
```diff
@@ -0,0 +1,327 @@
...
+  StartChunk int32
+  StopChunk  int32
+}
+
+func (p RangeMerkleProof) String() string {
```
xdralex: Ditto unneeded string dumps

docs/protocol/merkle_proof.go
```diff
@@ -0,0 +1,327 @@
...
+}
+
+type ByteRange struct {
+  Offset    int32
+  Length    int32
+  ChunkSize int32
+}
+
+// Fluence Merkle Tree chunk size
+const FlChunkSize int32 = 4
```
xdralex: I would rather concentrate on the algorithm than on specific Fluence vs Swarm implementation.
You could showcase the algorithm on the generic Fluence example and say that Swarm usage could be similar.

docs/protocol/merkle_proof.go
```diff
@@ -0,0 +1,327 @@
...
+
+// Swarm Merkle Tree chunk size
+const SwChunkSize int32 = 4000
+
+func Max(a int32, b int32) int32 {
```
xdralex: Would be better to move this to `package.go` I think :)

docs/protocol/merkle_proof.go
```diff
@@ -0,0 +1,327 @@
...
+    description.WriteString(" | ")
+    description.WriteString(hex.EncodeToString(h[1][:4]))
+  }
+  description.WriteString("\n")
+
+  return description.String()
+}
+
+type ByteRange struct {
+  Offset    int32
```
xdralex: Let's use uint64 instead of int32 and support byte arrays longer than ~4Gb~ 2Gb.

docs/protocol/merkle_proof.go
```diff
@@ -0,0 +1,327 @@
...
+  if a < b {
+    return a
+  }
+
+  return b
+}
+
+// split `data` in chunks of size `chunk`
+func Split(data []byte, chunk int32) []Chunk {
+  var dataLen = int32(len(data))
```
xdralex: uint64 :)

docs/protocol/merkle_proof.go
```diff
@@ -0,0 +1,327 @@
...
+  var chunksLen = dataLen / chunk
+  if (dataLen % chunk) != int32(0) {
```
xdralex: No need to cast 0 to int32 I guess (?)

docs/protocol/merkle_proof.go
```diff
@@ -0,0 +1,327 @@
...
+func Min(a int32, b int32) int32 {
+  if a < b {
+    return a
+  }
+
+  return b
+}
+
+// split `data` in chunks of size `chunk`
+func Split(data []byte, chunk int32) []Chunk {
```
xdralex: Maybe chunk => chunkSize to avoid confusion?

docs/protocol/merkle_proof.go
```diff
@@ -0,0 +1,327 @@
...
+    description.WriteString(hex.EncodeToString(h[0][:4]))
+    description.WriteString(" | ")
+    description.WriteString(hex.EncodeToString(h[1][:4]))
+  }
+  description.WriteString("\n")
+
+  return description.String()
+}
+
+type ByteRange struct {
```
xdralex: Should probably be called `ChunkedRange` (usually you don't expect ChunkSize in the byte range).

docs/protocol/merkle_proof.go
```diff
@@ -0,0 +1,327 @@
...
+  var stop = byteRange.StartChunk() + byteRange.Length/byteRange.ChunkSize
+  if (byteRange.Length % byteRange.ChunkSize) == int32(0) {
+    stop--
+  }
+
+  return stop
+}
+
+// Merkle Proof for an inclusion of a range of chunks in Fluence Merkle Tree
+func FlRangeMerkleProof(data []byte, offset int32, length int32) RangeMerkleProof {
```
xdralex: Ditto one core function instead of fluence / swarm specific ones.

docs/protocol/merkle_proof.go
```diff
@@ -0,0 +1,327 @@
...
+  var log = math.Log2(float64(n))
+  var pow = math.Floor(log)
+  if math.Remainder(log, 1.0) != 0.0 {
+    pow++
+  }
+
+  return int32(math.Pow(2, pow)), uint32(pow)
+}
+
+func hash(data []byte) Digest {
```
xdralex: Should probably use `Hash()` from `core.go`... implement it there if you'd like :)

docs/protocol/merkle_proof.go
```diff
@@ -0,0 +1,327 @@
...
+func SwRangeMerkleProof(data []byte, offset int32, length int32) RangeMerkleProof {
+  var byteRange = ByteRange{
+    Offset:    offset,
+    Length:    length,
+    ChunkSize: SwChunkSize,
+  }
+  return BuildRangeMerkleProof(data, byteRange, SwarmHash)
+}
+
+func nextPower2(n int32) (int32, uint32) {
```
xdralex: By reading just the declaration it is not clear what does this function return...
Yeah, it's some tuple, but what the elements are?

xdralex: Also, I don't think you really need this. In fact, you use this function only once – when building the Merkle Tree. However, there you use it to expand the array of chunks to the nearest 2<sup>n</sup> length. I would argue that's not really needed if you are not materializing the Merkle Tree and building the proof by layers.

docs/protocol/merkle_proof.go
```diff
@@ -0,0 +1,327 @@
...
+  return int32(math.Pow(2, pow)), uint32(pow)
+}
+
+func hash(data []byte) Digest {
+  var result = sha256.Sum256(data)
+  return result
+}
+
+// builds a Merkle Tree out of chunks applying hashFn as a hash function
+func BuildMerkleTree(chunks []Chunk, hashFn HashFunc) MerkleTree {
```
xdralex: Per our discussion, I think there is no need to materialize the Merkle tree :) Let's build the proof by layers.

xdralex: Also, let's not pass the HashFunc. First of all, it's needed only to support Fluence vs Swarm cases. Second, it's not used here anyways :)

docs/protocol/merkle_proof.go
xdralex: Also, I'm really, really, really skeptical about this function implementation and not really sure if it will be always working as desired given the possibility of floating point errors. 

I would search for the index of the most significant bit (e.g. for `00110100b = 52` byte this index will be 3, which gives us `8 – 3 + 1 = 6` for the power _n_ and `2^6 = 64` for the closest 2<sup>n</sup>). This can be done naively by shifting and iterations (in `O(n)` time) or by using clever multiplications – see [De Bruijn sequence](https://en.wikipedia.org/wiki/De_Bruijn_sequence#Uses).

docs/protocol/merkle_proof.go
```diff
@@ -0,0 +1,327 @@
...
+  var bottom = make([]Node, nodesCount)
+
+  var i int32
+  for ; i < chunksLen; i++ {
+    bottom[i] = Node{
+      Hash: hash(chunks[i]),
+    }
+  }
+
+  var empty = hash(make([]byte, len(chunks[0])))
```
xdralex: Let's say our chunk size is 4. Because `make([]byte, 4)` creates the `[0, 0, 0, 0]` array, this construction doesn't distinguish between `[1, 2, 3, 4]` and `[1, 2, 3, 4, 5, 6, 7, 8, 0, 0, 0, 0]` input data. For some cases it's fine, for some cases it's not. 

When we are using a binary Merkle tree, we compute the next level hash as `h_a-b = H(h_a, h_b)` where `h_a = H(a)` and `h_b = H(b)`. If `h_b` doesn't exist because chunks count is not an exact 2<sup>n</sup>, why you don't omit it altogether? In this case it will result in `h_a_b = H(h_a) = H(H(a))`.

cc @michailvoronov

docs/protocol/merkle_proof.go
```diff
@@ -0,0 +1,327 @@
...
+  return MerkleTree{
+    Root:   bottom[0],
+    Height: power + 1, // power is calculated just from the number of bottom tree nodes, so +1 is mandatory
+  }
+}
+
+// returns Merkle Tree nodes that form a Range Merkle Proof for chunks in the interval [start, stop]
+func RangeProof(tree MerkleTree, start int32, stop int32) [][2]Digest {
+  // -2 comes from: -1 since we already have 0x01, and -1 since we're starting at the level 1
+  var mask int32 = 0x1 << (tree.Height - 2)
```
xdralex: I don't really think you need bit shifts here... Integer division and multiplication by 2 as well as simple modular arithmetics should be enough and will look much simpler :)

docs/protocol/merkle_proof.go
```diff
@@ -0,0 +1,327 @@
...
+
+  // the resulting proof
+  // first [] stands for the level of the tree, starting at 1. So proof[0] contains hashes for the level 1
+  // the second [] stands for the left or right branch of proof. So proof[0][0] contains leftmost needed hash at the level 1.
+  var proof = make([][2]Digest, (tree.Height - 1))
```
xdralex: Maybe it will make things more clear to read if you replace [2]Digest with 

```go
type LayerExpansion struct {
  Left  DigestOpt
  Right DigestOpt
}
```

docs/protocol/merkle_proof.go
```diff
@@ -0,0 +1,327 @@
...
+
+    level++
+    mask = mask >> 1
+  }
+
+  return proof
+}
+
+// builds Merkle Proof for a range of `length` bytes starting at `offset` by splitting them in chunks of size `chunk`
+func BuildRangeMerkleProof(data []byte, byteRange ByteRange, hashFn HashFunc) RangeMerkleProof {
```
xdralex: ditto unneeded HashFunc

docs/protocol/merkle_proof.go
```diff
@@ -0,0 +1,327 @@
...
+  var chunks = Split(data, byteRange.ChunkSize)
```
xdralex: Honestly, I am having a hard time wrapping my head around the ByteRange and its fields and methods. Maybe you could simply pass an offset, length and hardcode the chunk side inside. Also compute the start/stop chunk inside as well instead of having that defined as methods on the ByteRange.

docs/protocol/merkle_proof.go
```diff
@@ -0,0 +1,327 @@
...
+  var startChunk = byteRange.StartChunk()
```
xdralex: I think from/to instead of start/stop sounds better... I feel like start/stop defines a process (like timing the actions with a stopwatch) whereas from/to defines a range.

docs/protocol/merkle_proof.go
```diff
@@ -0,0 +1,327 @@
...
+  var stopChunk = byteRange.StopChunk()
+
+  var bottomChunks = chunks[startChunk : stopChunk+1]
+
+  var tree = BuildMerkleTree(chunks, hashFn)
+  fmt.Printf("%v\n", tree)
```
xdralex: Let's get rid of those printlns :)

docs/protocol/merkle_proof.go
xdralex: I think the Merkle proof should be built for chunks first, and only then extended to byte ranges.

docs/protocol/merkle_proof.go
```diff
@@ -0,0 +1,327 @@
...
+
+  return RangeMerkleProof{
+    Chunks:     bottomChunks,
+    Hashes:     proof,
+    StartChunk: startChunk,
+    StopChunk:  stopChunk,
+  }
+}
+
+func CheckProof(root Digest, proof RangeMerkleProof) bool {
```
xdralex: Won't hurt to have some comments here :)

xdralex: Also, if you remove Start/StopChunk from the RangeMerkleProof, it would be nice for this function to compute those values and return them.

docs/protocol/merkle_proof.go
```diff
@@ -0,0 +1,327 @@
...
+
+    for i := 0; i < len(bottom)-1; i += 2 {
+      tree[i/2] = hash(append(bottom[i][:], bottom[i+1][:]...))
+    }
+
+    bottom = tree
+    mask = mask << 1
+  }
+
+  fmt.Println("tree[0]\t", len(tree[0]), tree[0])
```
xdralex: Ditto printlns

docs/protocol/merkle_proof.go
```diff
@@ -0,0 +1,327 @@
...
+    }
+
+    tree = make([]Digest, len(bottom)/2)
+
+    for i := 0; i < len(bottom)-1; i += 2 {
+      tree[i/2] = hash(append(bottom[i][:], bottom[i+1][:]...))
+    }
+
+    bottom = tree
+    mask = mask << 1
```
xdralex: Ditto bitwise operations replacement with integer division/etc

docs/protocol/merkle_proof.go
```diff
@@ -0,0 +1,327 @@
...
+  }
+
+  fmt.Println("tree[0]\t", len(tree[0]), tree[0])
+  fmt.Println("root\t", len(root), root)
+  return tree[0] == root
+}
+
+func printArray(label string, arr [][32]byte) {
```
xdralex: ditto ditto

Issue #280, opened 2018-11-08
Integration tests for Fluence Publisher CLI


DieMyst:
## Motivation

If the contract is updated, unhandled errors may occur in the Fluence Publisher CLI. Releases will be inconsistent, and users will notice it only during runtime. Solving this issue will allow us to avoid unexpected problems in the future.

## Proposition

Create an integration test for interaction between the Fluence CLI, the smart contract running on Ethereum blockchain and the Swarm. In particular, cover the publishing functionality of CLI.

PR #281, opened 2018-11-08
Publisher integration tests

DieMyst:
#280


bootstrap/scripts/compile_sol.js
```diff
@@ -0,0 +1,43 @@
...
+var output = solc.compile({ sources: input }, 1)
+
+if (!output) {
+    abort('No output from compiler');
+} else if (output.errors) {
+    for (var error in output.errors) {
+            var message = output.errors[error]
+            console.error(message)
+    }
+}
```
folex: shouldn't it exit here? There is no output, so no reason in creating directories and trying to write bin and abi.

DieMyst: yes, sure

publisher/README.md
```diff
@@ -15,7 +15,7 @@ Current code deployment process to the Fluence network looks like this:
 
 ## Installation
 
-Requirements: [`rust`](https://www.rust-lang.org/install.html) and [`cargo`](https://doc.rust-lang.org/cargo/getting-started/installation.html) are needed.
+Requirements: [`rust`](https://www.rust-lang.org/install.html), [`cargo`](https://doc.rust-lang.org/cargo/getting-started/installation.html) and [`npm`](https://www.npmjs.com/get-npm) are needed.
```
RicoGit: Cargo installation will be enough. There is no need to install `rust` before that. Let's left only Cargo installation link.
```
Requirements: [`cargo`](https://doc.rust-lang.org/cargo/getting-started/installation.html) - The Rust package manager and [`npm`](https://www.npmjs.com/get-npm) are needed.
```

publisher/build.rs
```diff
@@ -0,0 +1,17 @@
+use std::process::Command;
```
RicoGit: I saw that some others Rust files have file header. Let's make this consistent.

publisher/build.rs
```diff
@@ -0,0 +1,17 @@
...
+    install_cmd.args(&["install"]);
+    install_cmd.status().unwrap();
+
+    println!("Compiling sol files...");
+
+    let mut gen_cmd = Command::new("npm");
+    gen_cmd.current_dir("../bootstrap/");
+    gen_cmd.args(&["run", "compile-sol"]);
+    gen_cmd.status().unwrap();
+}
```
RicoGit: eof

publisher/scripts/swarm-simulation.js
```diff
@@ -0,0 +1,8 @@
+var http = require('http');
+
+process.title = process.title = "swarm-simulation";
+
+http.createServer(function (request, response) {
+    response.writeHead(200);
+    response.end("d1f25a870a7bb7e5d526a7623338e4e9b8399e76df8b634020d11d969594f24a", 'utf-8');
+}).listen(8500);
```
RicoGit: eof

publisher/src/main.rs
```diff
@@ -25,59 +25,99 @@ mod publisher;
...
-    bar.finish_with_message("Code submitted.");
+    let transaction = publish(publisher, true);
 
     let formatted_finish_msg = style("Code published. Submitted transaction").blue();
     let formatted_tx = style(transaction.unwrap()).red().bold();
 
     println!("{}: {:?}", formatted_finish_msg, formatted_tx);
 }
 
+pub fn with_progress<U, F: FnOnce() -> U>(msg: &str, prefix: &str, finish: &str, f: F) -> U {
```
RicoGit: Let's avoid one letter names and add docs for the public.

publisher/src/main.rs
```diff
@@ -25,59 +25,99 @@ mod publisher;
...
+    let bar = create_progress_bar(prefix, msg);
+    let result = f();
+    bar.finish_with_message(finish);
+    result
+}
+
+fn publish(publisher: Publisher, show_progress: bool) -> Result<H256, Box<Error>> {
+    // uploading code to swarm
```
RicoGit: If `upload_code_to_swarm` method will be created, then there will not need to write this comment and code will be self-describable.

publisher/src/main.rs
```diff
@@ -133,3 +175,159 @@ fn create_progress_bar(prefix: &str, msg: &str) -> ProgressBar {
...
+mod tests {
+    use super::{get_contract, publish};
+    use publisher::app::Publisher;
+    use std::error::Error;
+    use web3;
+    use web3::contract::Options;
+    use web3::futures::Future;
+    use web3::types::*;
+
+    #[cfg(test)]
```
RicoGit: Is it this attribute really required?

publisher/src/main.rs
```diff
@@ -133,3 +175,159 @@ fn create_progress_bar(prefix: &str, msg: &str) -> ProgressBar {
...
+                "addAddressToWhitelist",
+                account,
+                OWNER.parse()?,
+                Options::default(),
+            )
+            .wait()?)
+    }
+
+    #[test]
+    fn publish_wrong_password() -> Result<(), Box<Error>> {
```
RicoGit: I think it is redundant for test return Result.
If the test passed it just don't 'panic'.
For what reason did you do that?

DieMyst: it is more comfortable to use `?` instead of `unwrap`

publisher/src/main.rs
DieMyst: yes, or compilator will indicate unused methods

publisher/src/main.rs
RicoGit: A lot of boilerplate code in this case.

publisher/src/main.rs
RicoGit: Could you post compile error message?

DieMyst:
```
warning: struct is never used: `SemanticDirection`
 --> src/main.rs:1:1
  |
1 | struct SemanticDirection;
  | ^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: #[warn(dead_code)] on by default
```

RicoGit: I add #[allow(dead_code)] like this 
```
#[cfg(test)]
#[allow(dead_code)]
mod tests {
...
}
```
and no one 'warning' appears.

PR #283, opened 2018-11-08
Contiguous memory in VMState instead of chunks

folex:
- integrate Range Merkle Proofs in code and in protocol description
- migrate `VMState` chunks to `[]byte` memory
- rename 
     - `MerkleRoot` -> `MerkleHash`
     - `SwarmHash` -> `SwarmMerkleHash`
     - `TmMerkleRoot` -> `TmMerkleHash`
- migrate `QueryResults` chunks to `MemoryRegion` of type `[]byte`
- add `VerifySwarmProof`
- be more specific about calculating Merkle Root from VM memory

Open question:
Should I add `CreateSwarmProof`? It's not required anywhere, but I feel like it's missing since there is a `VerifySwarmProof`


folex:
Will be merged as a part of #285

PR #285, opened 2018-11-09
Use memory region for hash mismatch dispute & query results

folex:
Changes:
- integrate Range Merkle Proofs in code and in protocol description
- migrate `VMState` chunks to `[]byte` memory
- rename 
     - `MerkleRoot` -> `MerkleHash`
     - `SwarmHash` -> `SwarmMerkleHash`
     - `TmMerkleRoot` -> `TmMerkleHash`
- migrate `QueryResults` chunks to `MemoryRegion` of type `[]byte`
- add `VerifySwarmProof`
- be more specific about calculating Merkle Root from VM memory
- introduce `MemoryRegion` struct
- change hash mismatch dispute from finding different chunks to finding different byte ranges
- change `QueryResults` to use `MemoryRegion` instead of `[]byte`

Questions:
- should I add `CreateSwarmProof`? It's not required anywhere, but I feel like it's missing since there is a `VerifySwarmProof`
- maybe `ByteRegion` is a better name than `MemoryRegion`?
- is `int32` ok or should it be `uint32` or `[u]int64`?
- should I describe procedure of finding <code>state[u:v]<sub>k</sub><sup>A</sup> != state[u:v]<sub>k</sub><sup>M</sup></code>?


docs/protocol.md
```diff
@@ -409,7 +409,7 @@ Tendermint uses Merkle trees to compute the Merkle root of certain pieces of dat
 // listed Tendermint functions carry the same meaning and arguments as core functions
 func TmSign(publicKey PublicKey, privateKey PrivateKey, digest Digest) Seal {}
 func TmVerify(seal Seal, digest Digest) bool {}
-func TmMerkleRoot(chunks []Chunk) Digest {}
+func TmMerkleHash(chunks []Chunk) Digest {}
```
xdralex: You have never described what the ordinary `MerkleHash` is.

docs/protocol.md
```diff
@@ -451,13 +451,13 @@ Note we haven't specified here how the application state hash (`Block.Header.App
 
 ### Block processing
 
-Once the block has passed through Tendermint consensus, it is delivered to the state machine. State machine passes block transactions to the WebAssembly VM causing the latter to change state. The virtual machine state is essentially a block of memory split into chunks which can be used to compute the virtual machine state hash. We can say that the virtual machine state `k + 1` is derived by applying transactions in the block `k` to the virtual machine state `k`.
+Once the block has passed through Tendermint consensus, it is delivered to the state machine. State machine passes block transactions to the WebAssembly VM causing the latter to change state. The virtual machine state is essentially a block of memory that can be split into chunks to compute the virtual machine Merkle Root. We can say that the virtual machine state `k + 1` is derived by applying transactions in the block `k` to the virtual machine state `k`.
```
xdralex: Lowercase "root" in Merkle root please :)

docs/protocol.md
```diff
@@ -451,13 +451,13 @@ Note we haven't specified here how the application state hash (`Block.Header.App
...
 
 ```go
 type WasmCode = []Chunk
 
 type VMState struct {
-  Chunks []Chunk // virtual machine memory chunks
```
xdralex: Question: should we use 
```
type VMState struct {
  Memory[] byte
}
```

or just `type VMState = []byte`? Seems like virtual machine state here **is** just an array of bytes.

docs/protocol.md
```diff
@@ -469,7 +469,7 @@ Once the block is processed by the WebAssembly VM, it has to be stored in Swarm
 ```go
 type Manifest struct {
   Header              Header       // block header
-  VMStateHash         Digest       // hash of the VM state derived by applying the block
+  VMStateHash         Digest       // Merkle Root of the VM state derived by applying the block
```
xdralex: Merkle root :)

docs/protocol.md
```diff
@@ -513,28 +513,25 @@ Let's assume the transaction sent by the client was included into the block `k`.
 
 ```go
 type QueryResponse struct {
-  Chunks    map[int]Chunk       // selected virtual machine state chunks
-  Proofs    map[int]MerkleProof // Merkle proofs: chunks belong to the virtual machine state
-  Manifests [3]Manifest         // block manifests
+  MemoryRegion MemoryRegion // region of the virtual machine memory containing query result
```
xdralex: **the** query result

docs/protocol.md
```diff
@@ -513,28 +513,25 @@ Let's assume the transaction sent by the client was included into the block `k`.
...
+  Proof        MerkleProof  // Merkle Proof for `Memory` belonging to the whole VM memory
```
xdralex: "Merkle proof" :) What is `Memory`? 
I'd rather say: "Merkle proof that the memory region belongs to the whole VM memory"

docs/protocol.md
```diff
@@ -513,28 +513,25 @@ Let's assume the transaction sent by the client was included into the block `k`.
...
+  Manifests    [3]Manifest  // block manifests
 }
 ```
 
-Results obtained by invoking the function are stored as a part of the virtual machine state. This way, the node can return selected chunks of the virtual machine memory and supply Merkle proofs confirming these chunks indeed correspond to the correct virtual machine state Merkle root.
+Results obtained by invoking the function are stored as a part of the virtual machine state. This way, the node can return selected region of the virtual machine memory and supply a Merkle proof confirming the region indeed correspond to the correct virtual machine state's Merkle root.
```
xdralex: correspondS

docs/protocol.md
```diff
@@ -513,28 +513,25 @@ Let's assume the transaction sent by the client was included into the block `k`.
...
-  var proofs = make(map[int]MerkleProof)
-
-  for _, index := range chunksIndices {
-    var chunk = vmState.Chunks[index]
-
-    chunks[index] = chunk
-    proofs[index] = CreateMerkleProof(vmState.Chunks, int32(index))
+// prepares the query response containing memory region with results
+func MakeQueryResponse(manifests [3]Manifest, vmState VMState, offset int32, length int32) QueryResponse {
+  var proof = CreateMerkleProof(vmState.Memory, offset, length)
```
xdralex: CreateMerkleProof is never defined in the protocol

docs/protocol.md
```diff
@@ -513,28 +513,25 @@ Let's assume the transaction sent by the client was included into the block `k`.
...
+  var memoryRegion = MakeMemoryRegion(vmState.Memory, offset, length)
```
xdralex: MakeMemoryRegion is never defined in the protocol

docs/protocol.md
```diff
@@ -732,8 +728,8 @@ The validation contract carries endorsements of the snapshot by batch validators
 ```go
 // uploads the VM state to Swarm if needed and endorses it in the validation smart contract
 func (validator BatchValidator) Endorse(contract ValidationFluenceContract, height int64, state VMState) {
-  var swarmHash = SwarmHash(pack(state.Chunks))
-  var vmStateHash = MerkleRoot(state.Chunks)
+  var swarmHash = SwarmHash(state.Memory)
+  var vmStateHash = MerkleRoot(state.Memory)
```
xdralex: MerkleHash?

docs/protocol.md
```diff
@@ -766,8 +762,8 @@ func (validator BatchValidator) LoadSnapshot(contract ValidationFluenceContract,
   var meta = confirmation.SnapshotMeta
 
   var state = VMStateUnpack(SwarmDownload(meta.SnapshotReceipt))
-  var correct = meta.VMStateHash == MerkleRoot(state.Chunks)
-  
+  var correct = meta.VMStateHash == MerkleRoot(state.Memory)
```
xdralex: MerkleHash?

docs/protocol.md
```diff
@@ -785,7 +781,7 @@ For each block the batch validator verifies that BFT consensus was reached by th
 func VerifyVMStateConsensus(contract BasicFluenceContract, manifests [3]Manifest) []PublicKey {}
 ```
 
-The batch validator applies blocks one by one to the snapshot and computes the virtual machine state hash after each block application. If the calculated hash doesn't match the hash stored in the manifest, either the batch validator or the real-time cluster has performed an incorrect state advance. This condition should be resolved via the verification game which is described later in this document.
+The batch validator applies blocks one by one to the snapshot and computes a Merkle Root of the virtual machine state after each block application. If the calculated Merkle Root doesn't match the `VmStateHash` stored in the manifest, either the batch validator or the real-time cluster has performed an incorrect state advance. This condition should be resolved via the verification game which is described later in this document.
```
xdralex: Merkle root

docs/protocol.md
```diff
@@ -802,23 +798,25 @@ func (validator BatchValidator) Validate(
   var snapshot, ok = validator.LoadSnapshot(validationContract, height - sideContract.CheckpointInterval)
 
   if ok {
+    var nextSnapshot VMState
```
xdralex: Why the need for another variable? Doesn't look like the original `snapshot` is going to be used anywhere...

docs/protocol.md
```diff
@@ -847,35 +845,35 @@ Now, two options are possible:
...
 
- It's obvious that for any chunk index **i** in the state snapshot, **A** can generate a Merkle proof proving this chunk data <code>state[i]<sub>k</sub><sup>A</sup></code> corresponds to the Merkle root <code>vmHash<sub>k</sub><sup>A</sup> == MerkleRoot(state<sub>k</sub><sup>A</sup>)</code> and at the same time another Merkle proof that it corresponds to the Merkle root <code>swHash<sub>k</sub><sup>A</sup> == SwarmHash(state<sub>k</sub><sup>A</sup>)</code>. 
+ It's obvious that for any byte range `[u, v]` in the state snapshot, **A** can generate a Merkle proof proving this data <code>state[u:v]<sub>k</sub><sup>A</sup></code> corresponds to the Merkle root <code>vmHash<sub>k</sub><sup>A</sup> == MerkleRoot(state<sub>k</sub><sup>A</sup>)</code> and at the same time another Merkle proof that it corresponds to the Merkle root <code>swHash<sub>k</sub><sup>A</sup> == SwarmHash(state<sub>k</sub><sup>A</sup>)</code>.
```
xdralex: `i:j` please :) too many variables already haha

xdralex: MerkleRoot => MerkleHash here? And not only here, but throughout the doc

docs/protocol.md
```diff
@@ -847,35 +845,35 @@ Now, two options are possible:
...
 
- Now, all **A** needs to do is to find a chunk index **p** which points to the mismatched content in computed states: <code>state[p]<sub>k</sub><sup>A</sup> != state[p]<sub>k</sub><sup>M</sup></code>, and submit a dispute to Ethereum smart contract requesting **M** to provide chunk **p** data and Merkle proofs proving its belonging to <code>vmHash<sub>k</sub><sup>M</sup></code> and <code>swHash<sub>k</sub><sup>M</sup></code>.
+ Now, all **A** needs to do is to find a byte range `[u, v]` which differs in computed states: <code>state[u:v]<sub>k</sub><sup>A</sup> != state[u:v]<sub>k</sub><sup>M</sup></code>, and submit a dispute to Ethereum smart contract requesting **M** to submit the byte range `[u, v]` and Merkle proofs proving its belonging to <code>vmHash<sub>k</sub><sup>M</sup></code> and <code>swHash<sub>k</sub><sup>M</sup></code>.
```
xdralex: ditto `i:j`

docs/protocol.md
```diff
@@ -847,35 +845,35 @@ Now, two options are possible:
...
 
  Because those hashes were produced from different snapshots in the first place, **M** will not be able to do that. If the contract finds one of two proofs invalid or **M** times out, **M** is considered as the side which has lost the dispute and will have it's deposit slashed.
 
 ```go
 // opens a new snapshot hash mismatch dispute
-func (contract ValidationFluenceContract) OpenSnapshotDispute(height int64, chunkIndex int) SnapshotDispute {
+func (contract ValidationFluenceContract) OpenSnapshotDispute(height int64, offset int32, length int32) SnapshotDispute {
```
xdralex: Honestly, i think `offset` and `length` should be something like `uint64` (like `size_t` haha). We might use >4GB memory and also don't really need negative values for these variables

docs/protocol.md
```diff
@@ -847,35 +845,35 @@ Now, two options are possible:
...
   return SnapshotDispute{
     SnapshotMeta: contract.Confirmations[height].SnapshotMeta,
-    ChunkIndex:   chunkIndex,
+    Offset:       offset,
+    Length:       length,
   }
 }
 
+// requests to submit specified byte range along with the Merkle Proof for it
```
xdralex: Not sure if I understand this...

xdralex: Could you also please reformat this struct?

docs/protocol.md
xdralex: This line is also too long I think so it doesn't fit GitHub page...

docs/protocol.md
```diff
@@ -847,35 +845,35 @@ Now, two options are possible:
...
 }
 
 // returns whether the supplied Merkle proofs have passed an audite
-func (dispute SnapshotDispute) Audit(chunk Chunk, vmProof MerkleProof, swarmProof MerkleProof) bool {
-  // TODO: check chunk index in the proof
-  // TODO: use Swarm-based Merkle proof verification
-
-  return VerifyMerkleProof(chunk, vmProof, dispute.SnapshotMeta.VMStateHash) &&
-      VerifyMerkleProof(chunk, swarmProof, dispute.SnapshotMeta.SnapshotReceipt.ContentHash)
+func (dispute SnapshotDispute) Audit(memoryRegion MemoryRegion, vmProof MerkleProof, swarmProof MerkleProof) bool {
```
xdralex: This line is also too long...

docs/protocol.md
```diff
@@ -896,13 +894,13 @@ As a reminder, for now we have been treating the virtual machine as a black box:
 func NextVMState(code WasmCode, vmState VMState, txs Transactions) VMState {}
 ```
 
-The client receives a segment of the virtual machine state along with the proof that the virtual machine state hash was stored in Swarm for future verification. The client also receives a proof that returned segment indeed corresponds to the state hash.
+The client receives a region of the virtual machine memory along with the proof that the Merkle Root of the virtual machine state was stored in Swarm for future verification. The client also receives a proof that returned segment indeed corresponds to the state Merkle Root.
```
xdralex: ditto Merkle root

docs/protocol.md
```diff
@@ -896,13 +894,13 @@ As a reminder, for now we have been treating the virtual machine as a black box:
...
 
 ```go
 type QueryResponse struct {
```
xdralex: I think I have already commented on this somewhere earlier

docs/protocol.md
```diff
@@ -949,6 +947,19 @@ func ProcessBlock(code WasmCode, block Block, prevVMState VMState, ...) {
   }  
   SwarmUpload(pack(manifest))
 }
+
+func ProcessBlock(code WasmCode, block Block, prevVMState VMState, ...) {
```
xdralex: Don't you want to remove the old version of this code block?

docs/protocol.md
```diff
@@ -732,8 +728,8 @@ The validation contract carries endorsements of the snapshot by batch validators
 ```go
 // uploads the VM state to Swarm if needed and endorses it in the validation smart contract
 func (validator BatchValidator) Endorse(contract ValidationFluenceContract, height int64, state VMState) {
-  var swarmHash = SwarmHash(pack(state.Chunks))
-  var vmStateHash = MerkleRoot(state.Chunks)
+  var swarmHash = SwarmHash(state.Memory)
```
xdralex: Seems like inconsistency – sometimes it's SwarmHash, sometimes SwarmMerkleHash in this doc

docs/protocol/memory_region.go
```diff
@@ -0,0 +1,32 @@
+package protocol
+
+// default chunk size for Merkle Proofs over bytes
+const FlChunkSize int32 = 4096
```
xdralex: Let's just pass the chunk size to the method as an argument

docs/protocol/memory_region.go
```diff
@@ -0,0 +1,32 @@
...
+  if (rangeEnd % FlChunkSize) != int32(0) {
+    to += FlChunkSize
+  }
+
+  return MemoryRegion {
+    ExtendedRegion: data[from:to],
+    offset: offset,
+    length: length,
+  }
+}
```
xdralex: newline please

docs/protocol/memory_region.go
```diff
@@ -0,0 +1,32 @@
+package protocol
+
+// default chunk size for Merkle Proofs over bytes
+const FlChunkSize int32 = 4096
+
+// represents a byte range selected from memory to construct a Merkle Proof over it
```
xdralex: Merkle proof

docs/protocol/memory_region.go
```diff
@@ -0,0 +1,32 @@
...
+// if selected range is not aligned to chunk size, then it's extended to be aligned
```
xdralex: "with chunks boundaries, it is extended to get aligned"
(not only chunk size matters: for 4KB chunk, 2KB:6KB is perfectly sized but is still not suitable for the Merkle proof)

docs/protocol/memory_region.go
```diff
@@ -0,0 +1,32 @@
...
+  var from = offset - offset % FlChunkSize
+  var rangeEnd = offset + length
+  var to = rangeEnd - rangeEnd % FlChunkSize
+  if (rangeEnd % FlChunkSize) != int32(0) {
+    to += FlChunkSize
+  }
+
+  return MemoryRegion {
+    ExtendedRegion: data[from:to],
+    offset: offset,
```
xdralex: This offset is computed incorrectly

docs/protocol/memory_region.go
```diff
@@ -0,0 +1,32 @@
+package protocol
+
+// default chunk size for Merkle Proofs over bytes
+const FlChunkSize int32 = 4096
+
+// represents a byte range selected from memory to construct a Merkle Proof over it
+// if selected range is not aligned to chunk size, then it's extended to be aligned
+type MemoryRegion struct {
+  ExtendedRegion []byte // selected region of the memory; extended to be aligned to chunks
```
xdralex: Let's just call this `BackingArray` :)

docs/protocol/memory_region.go
```diff
@@ -0,0 +1,32 @@
...
+  offset int32          // start of the byte range in `Region`
```
xdralex: What is `Region`?

docs/protocol/memory_region.go
```diff
@@ -0,0 +1,32 @@
...
+  length int32          // length of the byte range
+}
+
+// returns original unextended byte range
+func (region MemoryRegion) ByteRange() []byte {
```
xdralex: Let's just call this Data :)

docs/protocol/memory_region.go
```diff
@@ -0,0 +1,32 @@
...
+
+// represents a byte range selected from memory to construct a Merkle Proof over it
+// if selected range is not aligned to chunk size, then it's extended to be aligned
+type MemoryRegion struct {
+  ExtendedRegion []byte // selected region of the memory; extended to be aligned to chunks
+  offset int32          // start of the byte range in `Region`
+  length int32          // length of the byte range
+}
+
+// returns original unextended byte range
```
xdralex: the

docs/protocol/merkle_proof.go
```diff
@@ -12,5 +12,16 @@ type MerkleProof struct {
 // splits the byte sequence into chunks of specific size
 func Split(data []byte, chunkSize int32) []Chunk { panic("") }
 
-// build Range Merkle Proof for the range of chunks
+// splits `data` in default-sized chunks and calculates Merkle Root out of them
+func MerkleHash(data []byte) Digest { panic("") }
+
+// builds Range Merkle Proof for the range of chunks
```
xdralex: Do we really need this function?

docs/protocol/merkle_proof.go
```diff
@@ -12,5 +12,16 @@ type MerkleProof struct {
...
 func BuildMerkleProof(chunks []Chunk, from int32, to int32) MerkleProof { panic("") }
+
+// splits `data` in default-sized chunks and calculates proof for the specified range
+func CreateMerkleProof(data []byte, offset int32, length int32) MerkleProof { panic("") }
+
+// checks merkle proof for the range of default-sized chunks
+func VerifyMerkleProof(data []byte, proof MerkleProof, merkleRoot Digest) bool { panic("") }
+
+func VerifySwarmProof(data []byte, proof MerkleProof, merkleRoot Digest) bool { panic("") }
```
xdralex: This should be in swarm.go probably

xdralex: also newline

docs/protocol/merkle_proof.go
```diff
@@ -12,5 +12,16 @@ type MerkleProof struct {
...
 
-// build Range Merkle Proof for the range of chunks
+// splits `data` in default-sized chunks and calculates Merkle Root out of them
+func MerkleHash(data []byte) Digest { panic("") }
+
+// builds Range Merkle Proof for the range of chunks
 func BuildMerkleProof(chunks []Chunk, from int32, to int32) MerkleProof { panic("") }
+
+// splits `data` in default-sized chunks and calculates proof for the specified range
+func CreateMerkleProof(data []byte, offset int32, length int32) MerkleProof { panic("") }
```
xdralex: This functions should be presented in the protocol

docs/protocol.md
folex: I'd go for `type VMState = []byte`

xdralex:
Re: questions

>should I add CreateSwarmProof? It's not required anywhere, but I feel like it's missing since there is a VerifySwarmProof

I think it might be needed in the snapshot dispute...

>maybe ByteRegion is a better name than MemoryRegion?
Agreed

>is int32 ok or should it be uint32 or [u]int64?
depends :) for offset/length I'd say uint64 is better

>should I describe procedure of finding state[u:v]kA != state[u:v]kM?
Um? Let's discuss

docs/protocol/merkle_proof.go
folex: No, it's not used yet. I'll delete it.

docs/protocol.md
folex: typo in field name, I mean `Merkle Proof for \`MemoryRegion\``

docs/protocol/memory_region.go
folex: It shouldn't be aligned to cover all needed siblings, that's the work for proof generation. Here it is enough to have bytes to be whole chunks, so everyone can be sure exactly what bytes the proof will be for. Proof itself will contain hashes for all required sibling chunks.

docs/protocol.md
folex: I think it's more readable that way. I was first confused about endorsing an old snapshot:
```
validator.Endorse(validationContract, height, snapshot)
```

docs/protocol.md
folex: %) wow

PR #287, opened 2018-11-19
Wasm logger

RicoGit:
- Applies the last version of 'asemle' (fixed problem with merge strategy for asm)
- Creates SDK logger module.
- Applies SDK logger to Llama_db wrapper

Today later will be added examples for using this logger.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -111,6 +111,12 @@ lazy val statemachine = (project in file("statemachine"))
       "net.i2p.crypto"         % "eddsa"          % "0.3.0",
       scalaTest
     ),
+    assemblyMergeStrategy in assembly := {
+      case PathList("module-info.class", xs @ _*) => MergeStrategy.first
+      case x =>
+        val oldStrategy = (assemblyMergeStrategy in assembly).value
+        oldStrategy(x)
+    },
     imageNames in docker := Seq(ImageName("fluencelabs/solver")),
     dockerfile in docker := {
       // Run `sbt docker` to create image
--- a/build.sbt
+++ b/build.sbt
@@ -41,6 +41,7 @@ lazy val `vm-counter` = (project in file("vm/examples/counter"))
     // only with system ClassLoader.
     assemblyJarName in assembly := "counter.jar",
     assemblyMergeStrategy in assembly := {
+      // a module definition fails compilation for java 8, just skip it
       case PathList("module-info.class", xs @ _*) => MergeStrategy.first
       case x =>
         val oldStrategy = (assemblyMergeStrategy in assembly).value
@@ -68,6 +69,7 @@ lazy val `vm-llamadb` = (project in file("vm/examples/llamadb"))
     commons,
     assemblyJarName in assembly := "llamadb.jar",
     assemblyMergeStrategy in assembly := {
+      // a module definition fails compilation for java 8, just skip it
       case PathList("module-info.class", xs @ _*) => MergeStrategy.first
       case x =>
         val oldStrategy = (assemblyMergeStrategy in assembly).value
@@ -112,6 +114,7 @@ lazy val statemachine = (project in file("statemachine"))
       scalaTest
     ),
     assemblyMergeStrategy in assembly := {
+      // a module definition fails compilation for java 8, just skip it
       case PathList("module-info.class", xs @ _*) => MergeStrategy.first
       case x =>
         val oldStrategy = (assemblyMergeStrategy in assembly).value
--- a/build.sbt
+++ b/build.sbt
@@ -21,7 +21,7 @@ lazy val vm = (project in file("vm"))
   .settings(
     commons,
     libraryDependencies ++= Seq(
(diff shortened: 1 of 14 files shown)


build.sbt
```diff
@@ -40,6 +40,12 @@ lazy val `vm-counter` = (project in file("vm/examples/counter"))
     // 'Asmble' code required for loading some classes (like RuntimeHelpers)
     // only with system ClassLoader.
     assemblyJarName in assembly := "counter.jar",
+    assemblyMergeStrategy in assembly := {
+      case PathList("module-info.class", xs @ _*) => MergeStrategy.first
```
alari: Could you add some comments to this block of code?

sdk/rust/src/logger/mod.rs
```diff
@@ -0,0 +1,189 @@
+//! Module for writes log mesages from Wasm code.
+//!
+//! This module is client for Logger Wasm Module in Fluence WasmVm. In couple
```
alari: Could you check the language in the module docs please?

sdk/rust/src/logger/mod.rs
```diff
@@ -0,0 +1,189 @@
...
+        let log_msg = format!(
+            "{:<5} [{}] {}\n",
+            record.level().to_string(),
+            record.module_path().unwrap_or_default(),
+            record.args()
+        );
+
+        unsafe {
+            for byte in log_msg.as_bytes() {
+                write(i32::from(*byte))
```
alari: Could you add a comment please about where it writes the byte?
Is it possible to use all 32 bits to write 4 bytes at once?

vm/src/main/scala/fluence/vm/config/VmConfig.scala
```diff
@@ -26,6 +26,8 @@ import scala.util.control.NoStackTrace
  *
  * @param defaultMaxMemPages the maximum count of memory pages when a module doesn't say
  * @param specTestRegister if true, registers the spec test harness as 'spectest'.
+ * @param loggerRegister if > 0, registers the logger Wasm module as 'logger'
```
alari: Where is it actually used?

build.sbt
RicoGit: ok

vm/src/main/scala/fluence/vm/config/VmConfig.scala
RicoGit: this flag used inside `asmble` for initialization [WasmLogger](https://github.com/fluencelabs/asmble/blob/master/compiler/src/main/kotlin/asmble/run/jvm/LoggerModule.kt#L11
)

sdk/rust/src/logger/mod.rs
```diff
@@ -0,0 +1,189 @@
...
+//!         } else {
+//!             simple_logger::init_with_level(log::Level::Info).unwrap();
+//!         }
+//!
+//!         error!("This message will be logged.");
+//!         trace!("This message will not be logged.");
+//!     }
+//!
+//! ```
+//! This example provides method for initialization [`WasmLogger`] only for Wasm
```
michaelvoronov: provide methods

vm/examples/llamadb/src/main/rust/lib.rs
```diff
@@ -28,6 +31,23 @@ type GenResult<T> = Result<T, Box<Error>>;
 // Public functions for work with Llamadb.
 //
 
+/// Initializes `WasmLogger` instance and returns result as string.
+/// Enabled only for Wasm target.
+#[no_mangle]
+#[cfg(target_arch = "wasm32")]
+pub unsafe fn init_logger(_: *mut u8, _: usize) -> NonNull<u8> {
```
michaelvoronov: From my point of view, it is better to return raw error codes instead of string consists of the constant message and  error code in the end.

sdk/rust/src/logger/mod.rs
RicoGit: WebAssebmly doesn't support type byte, so there we deside to pass i32. This is a common situation when byte passed as int. Any way machine word is more than 32 bit for commom hardware.
Yes, it is possible to use all 32 or 64 bits at once, but it requires a lot of unsafe code and maintaining VM and SDK consistency about byte order.  And I suggest it won't more performant, but it will be more complex. Cost of one call of WasmLogger in `asmble` less than converting bytes to long and back.

vm/examples/llamadb/src/main/rust/lib.rs
RicoGit: I don't think so.) The idea is that all public wasm methods return bytes in our predefined format.
For example json like this:
```js
{
    code: 0,
    errorMsg: 'Something goes wrong'
}
```
For now it's just a String. If I return just an error code how will you know what is actually wrong?

sdk/rust/src/logger/mod.rs
RicoGit: done

PR #291, opened 2018-11-23
Master nodes integration test

DmiSergeev:
This test contains a single test method that checks:
* MasterNode connectivity with ganache-hosted Deployer smart contract
* MasterNode ability to load previous node clusters and subscribe to new clusters
* Successful cluster formation and starting blocks creation

### DIFF
--- a/node/src/main/scala/fluence/node/MasterNode.scala
+++ b/node/src/main/scala/fluence/node/MasterNode.scala
@@ -50,7 +50,7 @@ case class MasterNode(
           clusterData,
           solverTendermintPath.toString,
           // TODO fetch (from swarm) & cache
-          Paths.get("./statemachine/docker/examples/vmcode-" + clusterData.code).toAbsolutePath.toString
+          Paths.get("../statemachine/docker/examples/vmcode-" + clusterData.code).toAbsolutePath.toString
         )
     }
 
--- a/node/src/main/scala/fluence/node/MasterNode.scala
+++ b/node/src/main/scala/fluence/node/MasterNode.scala
@@ -50,7 +50,10 @@ case class MasterNode(
           clusterData,
           solverTendermintPath.toString,
           // TODO fetch (from swarm) & cache
-          Paths.get("../statemachine/docker/examples/vmcode-" + clusterData.code).toAbsolutePath.toString
+          Paths
+            .get(getClass.getClassLoader.getResource("examples").getPath + "/vmcode-" + clusterData.code)
+            .toAbsolutePath
+            .toString
         )
     }
 
--- a/node/src/main/scala/fluence/node/tendermint/KeysPath.scala
+++ b/node/src/main/scala/fluence/node/tendermint/KeysPath.scala
@@ -22,6 +22,8 @@ import cats.effect.IO
 import fluence.node.docker.DockerParams
 import io.circe.parser.parse
 
+import scala.sys.process.Process
+
 /**
  * Wraps tendermint directory for the Master process (contains shared tendermint keys).
  *
@@ -103,7 +105,13 @@ case class KeysPath(masterTendermintPath: String) extends slogging.LazyLogging {
    * @param solverTendermintPath Solver's tendermint path
    */
   def copyKeysToSolver(solverTendermintPath: Path): IO[Unit] = path.map { p =>
-    Files.copy(
+    Process(
+      s"docker run --rm -i -v $p:/keys -v $solverTendermintPath:/solver --entrypoint cp fluencelabs/solver:latest -f /keys/config/node_key.json /solver/config/node_key.json"
+    ).!!
+    Process(
+      s"docker run --rm -i -v $p:/keys -v $solverTendermintPath:/solver --entrypoint cp fluencelabs/solver:latest -f /keys/config/priv_validator.json /solver/config/priv_validator.json"
+    ).!!
+  /*Files.copy(
       p.resolve("config").resolve("node_key.json"),
       solverTendermintPath.resolve("config").resolve("node_key.json"),
       REPLACE_EXISTING
@@ -113,6 +121,6 @@ case class KeysPath(masterTendermintPath: String) extends slogging.LazyLogging {
       p.resolve("config").resolve("priv_validator.json"),
       solverTendermintPath.resolve("config").resolve("priv_validator.json"),
       REPLACE_EXISTING
-    )
+    )*/
   }
 }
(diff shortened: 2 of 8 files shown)


node/src/main/scala/fluence/node/MasterNode.scala
```diff
@@ -50,7 +50,10 @@ case class MasterNode(
           clusterData,
           solverTendermintPath.toString,
           // TODO fetch (from swarm) & cache
-          Paths.get("./statemachine/docker/examples/vmcode-" + clusterData.code).toAbsolutePath.toString
+          Paths
+            .get(getClass.getClassLoader.getResource("examples").getPath + "/vmcode-" + clusterData.code)
```
folex: This way of resolving path doesn't work when running from sbt locally

node/src/main/scala/fluence/node/NodeConfig.scala
```diff
@@ -65,22 +67,21 @@ object NodeConfig extends slogging.LazyLogging {
...
-      }
+      _ <- checkIp(ip)
 
       startPort <- IO(startPortString.toShort)
       endPort <- IO(endPortString.toShort)
       _ ← checkPorts(startPort, endPort)
     } yield NodeConfig(ip, startPort, endPort, validatorKey, nodeAddress)
 
+  private def checkIp(ip: String): IO[Unit] =
+    Try(InetAddress.getByName(ip)) match {
```
alari: `Try` is eager, while `IO` is lazy. It's better to use `IO(InetAddress.getByName).attempt.flatMap{...}`, to keep the computation lazy as `IO` promises to be.

node/src/main/scala/fluence/node/tendermint/KeysPath.scala
```diff
@@ -103,16 +105,19 @@ case class KeysPath(masterTendermintPath: String) extends slogging.LazyLogging {
...
-      solverTendermintPath.resolve("config").resolve("node_key.json"),
-      REPLACE_EXISTING
-    )
-
-    Files.copy(
-      p.resolve("config").resolve("priv_validator.json"),
-      solverTendermintPath.resolve("config").resolve("priv_validator.json"),
-      REPLACE_EXISTING
-    )
+    DockerParams
```
alari: Why this way to copy files is better? Seems like an overkill.

node/src/test/scala/fluence/node/MasterNodeIntegrationSpec.scala
```diff
@@ -0,0 +1,267 @@
...
+ */
+class MasterNodeIntegrationSpec extends FlatSpec with LazyLogging with Matchers with BeforeAndAfterAll {
+
+  implicit private val ioTimer: Timer[IO] = IO.timer(global)
+  implicit private val ioShift: ContextShift[IO] = IO.contextShift(global)
+
+  private val url = sys.props.get("ethereum.url")
+  private val client = EthClient.makeHttpResource[IO](url)
+
+  val dir = new File("../bootstrap")
```
alari: Should this dependency be documented shomewhere with big red letters?

Or maybe we could keep all the resources required for testing in test resources?

node/src/test/scala/fluence/node/MasterNodeIntegrationSpec.scala
```diff
@@ -0,0 +1,267 @@
...
+            _ <- contract.addNode[IO](nodeConfig1)
+
+            // adding code – this should cause event, but MasterNodes not launched yet, so they wouldn't catch it
+            _ <- contract.addCode[IO](clusterSize = 2)
+
+            // sending useless tx - just to switch to a new block
+            _ <- contract.addAddressToWhitelist[IO](owner)
+
+            // launching MasterNodes - they should take existing cluster info via getNodeClusters
+            _ = new Thread(() => node0.run.unsafeRunSync()).start()
```
alari: This could be done with functional concurrency primitives, not sure if it matters in the test?

node/src/test/scala/fluence/node/MasterNodeIntegrationSpec.scala
folex: I created a ticket on reorganizing bootstrap directory & making this dependency more explicit.

https://github.com/orgs/fluencelabs/projects/1#card-15213789

node/src/main/scala/fluence/node/NodeConfig.scala
```diff
@@ -65,21 +67,21 @@ object NodeConfig extends slogging.LazyLogging {
...
-      }
+      _ <- checkIp(ip)
 
       startPort <- IO(startPortString.toShort)
       endPort <- IO(endPortString.toShort)
       _ ← checkPorts(startPort, endPort)
     } yield NodeConfig(ip, startPort, endPort, validatorKey, nodeAddress)
+  }
+
+  private def checkIp(ip: String): IO[Unit] =
```
alari: BTW: this could be any MonadError, as there's no IO being performed.

node/src/main/scala/fluence/node/docker/DockerParams.scala
```diff
@@ -79,6 +86,6 @@ object DockerParams {
     DockerParams(Queue("docker", "run", "-d"))
 
   def run(executable: String, params: String*): DockerParams =
-    DockerParams(Queue("docker", "run", "--rm", "-i", "--entrypoint", executable), params.toList)
+    DockerParams(Queue("docker", "run", "--user", "", "--rm", "-i", "--entrypoint", executable), params.toList)
```
alari: In this case, user is always set to empty string. Why? What's the difference with passing no `--user` option? What's the behaviour of docker command when `--user` option is set twice?

node/src/main/scala/fluence/node/tendermint/KeysPath.scala
```diff
@@ -15,19 +15,19 @@
...
+import fluence.node.docker.{DockerIO, DockerParams}
 import io.circe.parser.parse
 
 /**
  * Wraps tendermint directory for the Master process (contains shared tendermint keys).
  *
  * @param masterTendermintPath Tendermint's home directory
  */
-case class KeysPath(masterTendermintPath: String) extends slogging.LazyLogging {
+case class KeysPath(masterTendermintPath: String)(implicit ec: ContextShift[IO]) extends slogging.LazyLogging {
```
alari: Probably it's better to pass ContextShift in methods where it's required, not in constructor.

node/src/main/scala/fluence/node/tendermint/NodeInfo.scala
```diff
@@ -32,13 +32,13 @@ import io.circe.generic.semiauto._
 case class NodeInfo(cluster: Cluster, node_index: String) {
   def clusterName: String = cluster.genesis.chain_id
 
-  def writeTo(tendermintPath: Path): IO[Unit] = IO {
-    val config = tendermintPath.resolve("config")
-    config.toFile.mkdirs()
-    Files.write(config.resolve("node_info.json"), NodeInfo.nodeInfoEncoder(this).spaces2.getBytes)
+  def writeTo[F[_]](tendermintPath: Path)(implicit F: Sync[F]): F[Unit] = F.delay {
```
alari: IO operation is actually performed here, I'm not sure that `Sync.delay` is enough.

node/src/main/scala/fluence/node/docker/DockerParams.scala
folex: Ooops, bad typo.

Each next `--user $user` will override the previous one.

PR #293, opened 2018-11-27
Download code from swarm

DieMyst:


node/src/main/scala/fluence/node/MasterNode.scala
```diff
@@ -42,16 +43,54 @@ case class MasterNode(
   implicit ce: ConcurrentEffect[IO]
 ) extends slogging.LazyLogging {
 
+  /**
+    * Downloads file from the Swarm and store it on a disk.
+    * @param swarmPath a code address and a Swarm URL address
+    * @param filePath a path to code to store
+    */
+  private def downloadFromSwarmToFile(swarmPath: SwarmPath, filePath: Path): IO[Unit] = {
```
alari: This machinery should be factored out, so that MasterNode is responsible for calling some method that downloads the code, but doesn't implement it. MasterNode's responsibility is just to connect everything together.

node/src/main/scala/fluence/node/MasterNode.scala
```diff
@@ -42,16 +43,54 @@ case class MasterNode(
...
+
+  /**
+    * Checks if there is no code already then download a file from the Swarm and store it to a disk.
+    * @param solverTendermintPath a path to solver's directory
+    * @param swarmPath a code address and a Swarm URL address
+    * @return a path to a code
+    */
+  private def downloadAndWriteCodeToFile(solverTendermintPath: Path, swarmPath: SwarmPath): IO[String] =
+    for {
+      dirPath <- IO(solverTendermintPath.resolve("vmcode-" + swarmPath.hexAddress))
```
alari: Why should the path for solver's VM code be different? There's only one VM code folder per solver.

BTW, why VM code is stored into Tendermint path? Why Swarm-related code knows about Tendermint at all?

node/src/main/scala/fluence/node/MasterNode.scala
```diff
@@ -42,16 +43,54 @@ case class MasterNode(
...
+      _ <- if (dirPath.toFile.exists()) IO.unit else IO(Files.createDirectory(dirPath))
+      filePath <- IO(dirPath.resolve(swarmPath.hexAddress + ".wasm"))
+      _ <- if (filePath.toFile.exists()) IO.unit
```
alari: Is it safe? Could it throw?

node/src/main/scala/fluence/node/MasterNode.scala
```diff
@@ -42,16 +43,54 @@ case class MasterNode(
...
+    * Checks if there is no code already then download a file from the Swarm and store it to a disk.
+    * @param solverTendermintPath a path to solver's directory
+    * @param swarmPath a code address and a Swarm URL address
+    * @return a path to a code
+    */
+  private def downloadAndWriteCodeToFile(solverTendermintPath: Path, swarmPath: SwarmPath): IO[String] =
+    for {
+      dirPath <- IO(solverTendermintPath.resolve("vmcode-" + swarmPath.hexAddress))
+      _ <- if (dirPath.toFile.exists()) IO.unit else IO(Files.createDirectory(dirPath))
+      filePath <- IO(dirPath.resolve(swarmPath.hexAddress + ".wasm"))
```
alari: Here we have the unique code-related hash, no need to use it in the folder's name as well.

BTW, shouldn't we check that this file's Swarm hash corresponds to the address?

node/src/main/scala/fluence/node/MasterNode.scala
```diff
@@ -42,16 +43,54 @@ case class MasterNode(
...
+      _ <- if (filePath.toFile.exists()) IO.unit
+      else
+        IO(Files.createFile(filePath))
```
alari: Shouldn't it be handled inside `downloadFromSwarmToFile`?

node/src/main/scala/fluence/node/MasterNode.scala
```diff
@@ -42,16 +43,54 @@ case class MasterNode(
...
-        SolverParams(
-          clusterData,
-          solverTendermintPath.toString,
-          // TODO fetch (from swarm) & cache
-          Paths.get("./statemachine/docker/examples/vmcode-" + clusterData.code).toAbsolutePath.toString
-        )
+        for {
+          path <- clusterData.code match {
+            case LocalPath(localPath) =>
+              IO.pure(Paths.get("./statemachine/docker/examples/vmcode-" + localPath).toAbsolutePath.toString)
```
alari: It's not pure, as it could throw.

node/src/main/scala/fluence/node/NodeConfig.scala
```diff
@@ -35,7 +37,8 @@ case class NodeConfig(
   startPort: Short,
   endPort: Short,
   validatorKey: ValidatorKey,
-  nodeAddress: String
+  nodeAddress: String,
+  swarmAddress: Option[URL]
```
alari: - It's not documented
- Swarm address of what?

node/src/main/scala/fluence/node/NodeConfig.scala
```diff
@@ -56,11 +59,16 @@ object NodeConfig extends slogging.LazyLogging {
...
+        case a1 :: a2 :: a3 :: tail ⇒ IO.pure(a1, a2, a3, tail)
         case _ ⇒ IO.raiseError(new IllegalArgumentException("4 program arguments expected"))
       }
 
-      (ip, startPortString, endPortString) = argsTuple
+      (ip, startPortString, endPortString, tail) = argsTuple
+
+      swarmAddress <- tail.headOption.map(s => if (!s.startsWith("http")) "http://" + s else s) match {
+        case None => IO.pure(None)
+        case Some(addr) => IO(Some(new URL(addr)))
```
alari: Why not to use `.map` here?

node/src/main/scala/fluence/node/NodeConfig.scala
```diff
@@ -56,11 +59,16 @@ object NodeConfig extends slogging.LazyLogging {
   def fromArgs(keysPath: KeysPath, args: List[String]): IO[NodeConfig] =
     for {
       argsTuple ← args match {
-        case a1 :: a2 :: a3 :: Nil ⇒ IO.pure(a1, a2, a3)
+        case a1 :: a2 :: a3 :: tail ⇒ IO.pure(a1, a2, a3, tail)
```
alari: - It should be documented
- Let's leave a TODO to move all the arguments to config

node/src/main/scala/fluence/node/MasterNode.scala
```diff
@@ -42,16 +43,54 @@ case class MasterNode(
   implicit ce: ConcurrentEffect[IO]
 ) extends slogging.LazyLogging {
 
+  /**
+    * Downloads file from the Swarm and store it on a disk.
+    * @param swarmPath a code address and a Swarm URL address
+    * @param filePath a path to code to store
+    */
+  private def downloadFromSwarmToFile(swarmPath: SwarmPath, filePath: Path): IO[Unit] = {
+    SwarmClient(swarmPath.url.getHost, swarmPath.url.getPort).download(swarmPath.hexAddress).value.flatMap {
```
alari: Why to get Swarm host/port here? Why to create SwarmClient for each request? Can't we have a prepared local Swarm light node with configured port, a prepared SwarmClient with a known host and port, and avoid using swarmPath.url completely?

node/src/main/scala/fluence/node/tendermint/CodePath.scala
```diff
@@ -0,0 +1,32 @@
...
+
+package fluence.node.tendermint
+import java.net.URL
+
+sealed trait CodePath
+
+/**
+  * Path to a Swarm file.
+  * @param hexAddress an identifier of a file in the Swarm network
+  * @param url an address of a Swarm node
```
alari: Shouldn't it always be a local swarm daemon?

alari: Why to trust a gateway?
Why not to use the same Swarm address that node trusts to dispatch all the Swarm-related activities with?
Why to give the client an ability to provide any Swarm node, isn't it a security hole?
I believe that only Swarm hash and receipt should be given by the client, and the node should manage very carefully what Swarm processes it connects to.

node/src/main/resources/reference.conf
```diff
@@ -0,0 +1,10 @@
+tendermint-path = "target/.tendermint"
+endpoints {
+  ip = "127.0.0.1"
+  min-port = 20000
+  max-port = 20100
+}
+use-swarm: true,
```
alari: Could you add some comments to the reference configuration please

node/src/main/scala/fluence/node/MasterNode.scala
```diff
@@ -37,21 +38,25 @@ case class MasterNode(
...
-          // TODO fetch (from swarm) & cache
-          Paths.get("./statemachine/docker/examples/vmcode-" + clusterData.code).toAbsolutePath.toString
-        )
+        for {
+          path <- codeManager.prepareCode(clusterData.code, solverTendermintPath)
+        } yield {
+          SolverParams(
+            clusterData,
+            solverTendermintPath.toString,
+            path
```
alari: This path becomes a vmcode volume in docker container later. I think it should be completely independent from solverTendermintPath, which is also passed as a volume.

node/src/main/scala/fluence/node/MasterNode.scala
```diff
@@ -37,21 +38,25 @@ case class MasterNode(
...
+    _.evalMap {
       case (clusterData, solverTendermintPath) ⇒
-        SolverParams(
-          clusterData,
-          solverTendermintPath.toString,
-          // TODO fetch (from swarm) & cache
-          Paths.get("./statemachine/docker/examples/vmcode-" + clusterData.code).toAbsolutePath.toString
-        )
+        for {
+          path <- codeManager.prepareCode(clusterData.code, solverTendermintPath)
```
alari: We should keep tendermint path dedicated for tendermint data only, and never put VM code inside.

node/src/main/scala/fluence/node/NodeConfig.scala
```diff
@@ -22,18 +22,27 @@ import fluence.node.tendermint.{KeysPath, ValidatorKey}
...
+ * Information about a node possible endpoints.
  *
  * @param ip p2p host IP
- * @param startPort starting port for p2p port range
- * @param endPort ending port for p2p port range
+ * @param minPort starting port for p2p port range
+ * @param maxPort ending port for p2p port range
+**/
+case class EndpointsConfig(
+  ip: String,
```
alari: Could it be typed?

node/src/main/scala/fluence/node/NodeConfig.scala
```diff
@@ -22,18 +22,27 @@ import fluence.node.tendermint.{KeysPath, ValidatorKey}
 import scala.util.Try
 
 /**
- * Information about a node willing to run solvers to join Fluence clusters.
+ * Information about a node possible endpoints.
```
alari: The comment doesn't make anything clear :) what endpoints, why, how they are used?

node/src/main/scala/fluence/node/helpers.scala
```diff
@@ -0,0 +1,37 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package fluence.node
+import cats.effect.IO
+import pureconfig.error.ConfigReaderFailures
+
+object helpers {
+  implicit class ConfigOps[T](loadedConfig: Either[ConfigReaderFailures, T]) {
```
alari: It's better to make an object with a meaningful name to show what exactly you're going to import. For example, name this object `ConfigOps`, and make an implicit class inside it.

node/src/main/scala/fluence/node/solvers/CodeManager.scala
```diff
@@ -0,0 +1,113 @@
...
+   * @return
+   */
+  def prepareCode(path: CodePath, solverPath: Path): F[String]
+}
+
+/**
+ * Manager only for test purposes, uses precompiled code from Fluence repository.
+ *
+ */
+class TestCodeManager[F[_]](implicit F: Sync[F]) extends CodeManager[F] {
```
alari: - Is `Sync` enough? Maybe `Effect` describes what happens the better way? Or even `LiftIO`?
- Why not to move F to the method, why to keep it in signature?

DieMyst: because I use SwarmClient[F] into SwarmCodeManager

node/src/main/scala/fluence/node/MasterNodeApp.scala
```diff
@@ -16,47 +16,61 @@
...
+  contractConfig: DeployerContractConfig,
+  codeManager: CodeManager[IO]
+)
 
 object MasterNodeApp extends IOApp with LazyLogging {
 
   private val sttpResource: Resource[IO, SttpBackend[IO, Nothing]] =
     Resource.make(IO(AsyncHttpClientCatsBackend[IO]()))(sttpBackend ⇒ IO(sttpBackend.close()))
 
+  private def configure() =
```
folex: Can you please refactor this method to several smaller methods? It's very hard to comprehend

node/src/main/scala/fluence/node/solvers/CodeManager.scala
folex: You can use `SwarmClient[IO]` in `CodeManager`s

externalstorage/src/main/scala/fluence/swarm/SwarmClient.scala
```diff
@@ -309,15 +309,16 @@ class SwarmClient[F[_]: Monad](swarmUri: Uri)(
...
+  )(implicit sttpBackend: SttpBackend[F, Nothing], F: cats.MonadError[F, Throwable]): F[SwarmClient[F]] = {
 
     LoggerConfig.factory = PrintLoggerFactory()
     LoggerConfig.level = LogLevel.INFO
 
     implicit val hasher: Hasher[ByteVector, ByteVector] = Keccak256Hasher.hasher
-    implicit val sttpBackend: SttpBackend[IO, Nothing] = AsyncHttpClientCatsBackend[IO]()
 
-    IO(uri"$address").map(uri => new SwarmClient[IO](uri))
+    F.fromTry(Try(uri"$address")).map(addr => new SwarmClient[F](addr))
```
folex: Why not `F.catchNonFatal`?

node/src/main/scala/fluence/node/MasterNodeApp.scala
```diff
@@ -95,7 +88,16 @@ object MasterNodeApp extends IOApp with LazyLogging {
 
                 pool ← SolversPool[IO]()
 
-                node = MasterNode(masterKeys, nodeConfig, contract, pool, swarmClient, rootPath)
+                codeManager <- if (!swarmEnabled) IO(new TestCodeManager[IO]())
```
folex: I'd move that to a method, something like `getCodeManager(swarmEnabled: Boolean)`

PR #296, opened 2018-11-30
Wrap master node in docker container

folex:
### Changes
- refactor work with config files
- simplify docker entrypoints
- use alpine instead of ubuntu in docker containers, heavily reduce image size
- wrap master node in docker container
- start solver containers from within master container by sharing host `docker.sock`

### Usage
- You can run master node as `docker run -e "TENDERMINT_IP=host.docker.internal" -e "PORTS=20000:20002" -v /var/run/docker.sock:/var/run/docker.sock -it fluencelabs/master`
    - NOTE: `host.docker.internal` should be used for `macOS`, on linux you can just pass your IP


node/src/main/scala/fluence/node/solvers/CodeManager.scala
```diff
@@ -58,16 +58,17 @@ class TestCodeManager[F[_]](implicit F: Sync[F]) extends CodeManager[F] {
...
-      // TODO: REMOVE. It's an ad-hoc fix to avoid copying resources from jar in case of `sbt runMain`
-      Paths.get("./statemachine/docker/examples/vmcode-" + path.asString).toAbsolutePath.toString
-    } else {
-      Paths
-        .get(getClass.getClassLoader.getResource("examples").getPath + "/vmcode-" + path.asString)
-        .toAbsolutePath
-        .toString
-    }
+    "/master/vmcode/vmcode-llamadb"
+//    val examplesPath = getClass.getClassLoader.getResource("examples").getPath
```
DieMyst: what to do with this?

node/src/test/scala/fluence/node/MasterNodeIntegrationSpec.scala
```diff
@@ -43,6 +43,7 @@ import scala.language.higherKinds
 import scala.sys.process.{Process, ProcessLogger}
 import scala.util.Try
 
+/*
```
DieMyst: you can use `@Ignore` annotation

node/src/main/scala/fluence/node/tendermint/ClusterData.scala
```diff
@@ -34,16 +34,16 @@ case class ClusterData(
   persistentPeers: PersistentPeers,
   code: CodePath
 ) {
-  val hostP2PPort: Short = persistentPeers.peers(nodeInfo.node_index.toInt).port
-  val hostRpcPort: Short = (hostP2PPort + 100).toShort
-  val tmPrometheusPort: Short = (hostP2PPort + 200).toShort
-  val smPrometheusPort: Short = (hostP2PPort + 300).toShort
+  private val nodePeer = persistentPeers.peers(nodeInfo.node_index.toInt) //TODO: catch OutOfBounds exception
+  val P2PPort: Short = nodePeer.port
+  val RpcPort: Short = (P2PPort + 100).toShort //TODO: reserve service ports sequentially, right after p2p port
```
DieMyst: add the difference between maximum and minimum port to prevent cross-ports

node/src/main/scala/fluence/node/tendermint/KeysPath.scala
```diff
@@ -26,7 +27,7 @@ import io.circe.parser.parse
  *
  * @param masterTendermintPath Tendermint's home directory
  */
-case class KeysPath(masterTendermintPath: String) extends slogging.LazyLogging {
+case class KeysPath(masterTendermintPath: String)(implicit ec: ContextShift[IO]) extends slogging.LazyLogging {
```
DieMyst: why the implicit is needed here?

node/src/main/scala/fluence/node/docker/DockerIO.scala
```diff
@@ -55,6 +55,7 @@ object DockerIO extends LazyLogging {
           case 0 ⇒ logger.info(s"Container $dockerId successfully removed")
           case x ⇒ logger.warn(s"Stopping docker container $dockerId failed, exit code = $x")
         }
+        shiftDelay()
```
alari: This line drops the previous one, so that `rm -f` is not executed. But the dead code remains here. 

The shiny idea was to clean the controlled container when corresponding Scala stream handling code produces an interruption, to avoid the case when master node is dead, but solver container still consumes the resources.

This idea doesn't shine anymore if you use `run` to run short-living tasks (e.g. with `.compile.lastOrError`) or use docker's internal volumes.

How should we deal with it? Maybe add a couple of boolean flags, like `finallyStopContainer` and/or `finallyRemoveVolumes`? Or just keep `finallyStopContainers: Boolean = (if DockerParams contains -d flag)`?

Anyway this problem should be targeted, not just ignored :)

node/src/main/scala/fluence/node/docker/DockerParams.scala
```diff
@@ -86,6 +86,6 @@ object DockerParams {
     DockerParams(Queue("docker", "run", "-d"))
 
   def run(executable: String, params: String*): DockerParams =
-    DockerParams(Queue("docker", "run", "--rm", "-i", "--entrypoint", executable), params.toList)
+    DockerParams(Queue("docker", "run", "--user", "", "--rm", "-i", "--entrypoint", executable), params.toList)
```
alari: Do you really want `--user=""` option to be used by default?

node/src/main/scala/fluence/node/tendermint/ClusterData.scala
folex: I'll fix that later, ports should be reserved sequentially

node/src/main/scala/fluence/node/eth/EthereumRPCConfig.scala
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package fluence.node.eth
+
+case class EthereumRPCConfig(protocol: String, host: String, port: Int) {
```
alari: As I remember, we decided to use CamelCase convention for abbreviations.

node/src/main/scala/fluence/node/solvers/CodeManager.scala
```diff
@@ -57,18 +57,7 @@ class TestCodeManager[F[_]](implicit F: Sync[F]) extends CodeManager[F] {
...
-      // TODO: REMOVE. It's an ad-hoc fix to avoid copying resources from jar in case of `sbt runMain`
-      Paths.get("./statemachine/docker/examples/vmcode-" + path.asString).toAbsolutePath.toString
-    } else {
-      Paths
-        .get(getClass.getClassLoader.getResource("examples").getPath + "/vmcode-" + path.asString)
-        .toAbsolutePath
-        .toString
-    }
-  }
+  ): F[String] = F.pure("/master/vmcode/vmcode-" + path.asString) // preloaded code in master's docker container
```
alari: How will it work when running master with `sbt run`?

node/src/main/scala/fluence/node/solvers/SolverParams.scala
```diff
@@ -23,26 +24,28 @@ import fluence.node.tendermint.ClusterData
...
-  vmCodeDir: String,
+  solverPath: String,
+  vmCodePath: String,
+  masterNodeContainerId: String
 ) {
 
   override def toString = s"(solver ${clusterData.nodeInfo.node_index} for ${clusterData.nodeInfo.clusterName})"
 
-  def rpcPort: Short = clusterData.hostRpcPort
+  def rpcPort: Short = clusterData.RpcPort
```
alari: Wrong casing, `RpcPort` should begin with a lowercased letter.

node/src/main/scala/fluence/node/solvers/SolverParams.scala
```diff
@@ -23,26 +24,28 @@ import fluence.node.tendermint.ClusterData
...
   /**
    * [[fluence.node.docker.DockerIO.run]]'s command for launching a configured solver
    */
   val dockerCommand: DockerParams.Sealed =
     DockerParams
       .daemonRun()
-      .port(clusterData.hostP2PPort, 26656)
+      .option("-e", s"""CODE_DIR=$vmCodePath""")
+      .option("-e", s"""SOLVER_DIR=$solverPath""")
+      .port(clusterData.P2PPort, 26656)
```
alari: Wrong casing.

node/src/main/scala/fluence/node/tendermint/ClusterData.scala
```diff
@@ -34,16 +34,16 @@ case class ClusterData(
...
+  val RpcHost: String = nodePeer.host
```
alari: Please fix the cases, value names should start with lowercased letters.

node/src/main/scala/fluence/node/tendermint/KeysPath.scala
alari: +1

node/src/main/scala/fluence/node/tendermint/NodeInfo.scala
```diff
@@ -17,7 +17,7 @@
 package fluence.node.tendermint
 import java.nio.file.{Files, Path}
 
-import cats.effect.IO
+import cats.effect.Sync
```
alari: Is it used?

.travis.yml
```diff
@@ -7,7 +7,8 @@ matrix:
       services:
         - docker
       install:
-        - docker pull fluencelabs/solver:latest
+        - docker pull fluencelabs/solver:2018-dec-demo
```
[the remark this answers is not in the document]
folex: There problem is that I can't really release new docker images before I merge this PR to master, and tests in the PR won't pass before I release new docker images. 

So this is a workaround, I will create the next pull request after I release new docker images as 'latest'

PR #297, opened 2018-12-03
Validator for WASM code (draft)

RicoGit:
Subcommand 'check' is added to CLI. It's allowed to find function from defined banned modules.
```shell
$ cargo run -- check target/llama_db.wasm

All right! There is no function from banned modules ["std::sys::wasm::thread::", "std::sys::wasm::fs::", "std::sys::wasm::time::"]

```
```shell
$ cargo run -- check target/app.wasm
There was found next banned functions ->
  idx: 75 name: "std::sys::wasm::thread::Thread::set_name::h9b3f21eca537b2c9"
  idx: 74 name: "std::sys::wasm::thread::Thread::new::hdbe3a5d1a9c1be9b"
  idx: 76 name: "std::sys::wasm::thread::guard::current::h8003a47551f57e88"
```


cli/README.md
```diff
@@ -40,6 +40,7 @@ FLAGS:
 
 SUBCOMMANDS:
     add-to-whitelist    Adds an address to the whitelist of Fluence smart contract
+    check               Verifies wasm file, issue warning for 'dangerous code'
```
xdralex: I think it would be nice to provide some context on what properties the code should have to be considered dangerous. 

In the light of our previous discussion – will syscalls (through the imported host functions) be considered "dangerous"? Maybe there is a whitelist of supported imported functions, and the rest are not supported?

Or, is it just certain Rust modules that make the code "dangerous"? Is there a whitelist or a blacklist of such modules?

cli/clippy.toml
```diff
@@ -0,0 +1 @@
+blacklisted-names=["foo", "baz", "quux"] # 'bar' was excluded
```
xdralex: Not sure what this is... could you explain please?

cli/src/check.rs
```diff
@@ -0,0 +1,124 @@
...
+        .and_then(|module| module.parse_names().map_err(|err| err.0[0].1.clone()))
+        .map_err(Into::into)
+}
+
+/// Finds in the name section all functions from specified banned modules and
+/// returns their names and indexes.
+fn find_banned_fns_idxs<'a>(
+    module: &'a Module,
+    banned_modules: &[&str],
+) -> Option<HashMap<u32, &'a str>> {
```
xdralex: Is there really a difference between None and an empty HashMap?
Seems like returning just a HashMap (which would be none if there were no banned functions founds) would make things simpler a bit...

cli/src/check.rs
```diff
@@ -0,0 +1,124 @@
...
+
+    module.names_section().and_then(|name_sec| {
+        if let NameSection::Function(fn_name_sec) = name_sec {
+            let banned_fns: HashMap<u32, &'a str> = fn_name_sec
+                .names()
+                .iter()
+                .filter_map(|(idx, name)| {
+                    banned_modules
+                        .iter()
+                        .find(|m| name.starts_with(*m))
```
xdralex: Is there a reason to use `starts_with` check instead of exact equality?
Seems like this might mark as banned functions that don't really deserve it...

cli/src/check.rs
```diff
@@ -0,0 +1,124 @@
...
+            }
+        } else {
+            None
+        }
+    })
+}
+
+mod errors {
+    error_chain! {
+
```
xdralex: Not sure about the blank lines here – the way how they are dropped here doesn't look consistent with the rest of this file.

cli/src/check.rs
```diff
@@ -0,0 +1,124 @@
...
+/// Finds in the name section all functions from specified banned modules and
+/// returns their names and indexes.
+fn find_banned_fns_idxs<'a>(
+    module: &'a Module,
+    banned_modules: &[&str],
+) -> Option<HashMap<u32, &'a str>> {
+    use parity_wasm::elements::NameSection;
+
+    module.names_section().and_then(|name_sec| {
```
xdralex: Do you check only imported functions when you use `module.names_section()`? 
Or is it all (including user-defined) functions in the module?

cli/src/check.rs
xdralex: I'm also a bit not sure about the algorithmic complexity here: for **each (N total)** imported(?) function in the module this code checks that **each (K total)** function in banned modules they don't share a **prefix (L average length)**. 

This gives us `O(N*K*L)` complexity which might make things slow if the modules are big and complex enough. Maybe it's possible to achieve quasi-linear complexity – something like `O(max(N, K) * L)`?

cli/src/check.rs
RicoGit: All functions in the module except 'exported' functions.

cli/src/check.rs
```diff
@@ -0,0 +1,124 @@
...
+
+use self::errors::*;
+use clap::{App, Arg, ArgMatches, SubCommand};
+use console::style;
+use parity_wasm::elements::Module;
+use std::collections::HashMap;
+
+const INPUT_ARG: &str = "input";
+
+/// Performs validations according to the specified arguments. Writes results
```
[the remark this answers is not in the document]
RicoGit: "Would probably be nice to explain that it is looking for imported functions..." No so true, this method is the entry point for 'check' subcommand and will provide many different functions, your proposition is explained only one of this function.

"Or is it some Rust-specific stuff to make error handling in main.rs to work?" - yes, and it also idiomatic code in Scala too, when too return IO[Unit] or Try[Unit], it means that function does some effect or fails.
 "Seems like it never fails " - no it is. For example in the line `let module = read_wasm_file(file)?;`
see the article for more information: https://doc.rust-lang.org/book/2018-edition/ch09-02-recoverable-errors-with-result.html

cli/src/check.rs
folex: Maybe it is possible to make a map of banned modules and then create a list of modules by splitting function names on '::'. And then match list of used modules against banned ones. I think that would be a little bit faster.

Another option is to use [BTreeMap::range](https://doc.rust-lang.org/std/collections/btree_map/struct.BTreeMap.html#method.range) like [this](https://stackoverflow.com/questions/27344452/how-can-i-have-a-sorted-key-value-map-with-prefix-key-search). I'm not sure about complexity in that case, though, but I anticipate it will make code cleaner.

cli/src/check.rs
folex: Maybe add a comment on that?

cli/README.md
xdralex: Agreed on additional research, but I think the documentation for the CLI should at least explain what does "dangerous" mean in the current case.

cli/src/check.rs
xdralex: Yeah, I guess I agree with @folex that with some preprocessing the complexity would drop from quadratic to linear (ignoring methods name lengths `L`). 

Regarding whether is it needed or not –  I think: 
- it probably should not be a big deal to add some preprocessing 
- given that we deal with the untrusted code here it would be nice to handle it with linear complexity rather than quadratic

cli/src/check.rs
xdralex: +1 for the comment

cli/src/check.rs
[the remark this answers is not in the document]
xdralex: >will provide many different functions
Hm, what other functions will be there? – I guess based on the code I saw only one... 

Do you think it will be possible to describe those functions as well? This would help the reader to understand what the code does without looking inside.

>yes, and it also idiomatic code in Scala too, when too return IO[Unit] or Try[Unit], it means that function does some effect or fails.
Got it!

cli/src/check.rs
```diff
@@ -0,0 +1,124 @@
...
+}
+
+//
+// Private methods
+//
+
+/// Reads specified file, parses and returns module representation.
+fn read_wasm_file(file: &str) -> Result<Module> {
+    parity_wasm::deserialize_file(file)
+        .and_then(|module| module.parse_names().map_err(|err| err.0[0].1.clone()))
```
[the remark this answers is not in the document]
xdralex: +1

cli/src/check.rs
RicoGit: "Yeah, I guess I agree with @folex that with some preprocessing the complexity would drop from quadratic to linear" - 
Сomplexity will be `(N * L * log K)  + K log K` in this case where `K log K` for building search tree for K elements and `log K` for search element in the tree. Yes, it might be a little faster for big K in practice, but what is the K? 
K - it's banned modules, number of which will be at most 10 or 20, the same for L - it's module name that will contain at most 20 or 30 letters. But N might be actually big - hundreds and thousands of functions. 
The significant multiplier here is N, K and L might be considered as constant. In this case, the complexity can be considered as linear.

"given that we deal with the untrusted code here it would be nice to handle it with linear complexity rather than quadratic" - This code will be launched by the developer for its code, the code was written by itself can hardly be considered untrusted.

RicoGit: "Maybe it is possible..."
"Another option is to use BTreeMap::range like this..."
I think both of this variants have some pitfalls. Let's discuss this by the phone if it's really important to reach less asymptotic complexity for this case.

xdralex: >Сomplexity will be (N * L * log K) + K log K in this case where K log K for building search tree for K elements and log K for search element in the tree

If you replace the substring search with an exact match comparison you will be able to use a hash table with a constant search complexity. In this case we'll face `O(max(N, K) * L)` complexity. For example, you could parse `module1::module2::function_name123456678789` into the module name and the function name, and search for the module name in the hash table.

However, I have missed that you are checking modules, not functions (maybe add a comment on what actually happens in the code?). I agree that the number of banned modules shouldn't be too high, and for a developer checking its own code things should be pretty benign. 

But, I remember we were discussing that we might need to check imported functions, not used modules. This might become a potential security issue if we reuse this code so the node will be performing this check. Let's discuss offline.

RicoGit: We decide to do nothing with the algorithm. It is good enough for its purpose.

cli/README.md
xdralex: Why this was resolved? The CLI still doesn't explain the meaning of "dangerous code".

cli/src/check.rs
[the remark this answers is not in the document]
xdralex: Not sure why this was resolved – I still think it would be nice to explain what kinds of validations are being performed.

cli/src/check.rs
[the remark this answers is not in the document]
xdralex: Ditto unexpected resolve

cli/README.md
RicoGit: It's because I pushed changes only now.

cli/src/main.rs
```diff
@@ -92,6 +96,23 @@ fn main() {
...
+            handle_error(check::process(args));
+        }
+
         c => panic!("Unexpected command: {}", c.0),
     }
 }
+
+fn handle_error<T, E>(result: Result<T, E>)
+where
+    E: std::error::Error + error_chain::ChainedError,
```
folex: Why do you need `std::error::Error` here? `ChainedError` seems to already include it.

PR #301, opened 2018-12-05
Vm bencher

michaelvoronov:


bench/vm/bencher/BenchTestGenerator.py
```diff
@@ -0,0 +1,85 @@
...
+        test_descriptors : [TestDescriptor]
+            Descriptors of tests that specifies how exactly test should be compiled.
+
+        Returns
+        -------
+        [TestDescriptor]
+            Test descriptors with test_full_path filled.
+
+        """
+        call("mkdir {}".format(join(out_dir,self.generated_tests_dir)), shell=True)
```
RicoGit: if `out_dir` consists more than new 2 folders `mkdir` will fail. Adding flag `-p` will prevent this situation.

bench/vm/bencher/BenchTestGenerator.py
```diff
@@ -0,0 +1,85 @@
...
+        """Removes rust cargo target directories.
+
+        Attributes
+        ----------
+        out_dir : str
+            A directory where build results are placed.
+
+        """
+        call(("rm -rf " + join("{}", "wasm32-unknown-unknown")).format(out_dir), shell=True)
+        call(("rm -rf " + join("{}", "release")).format(out_dir), shell=True)
```
RicoGit: 'release' folder is inside 'wasm32-unknown-unknown' folder. Why it is needed to remove it separately?
`target/wasm32-unknown-unknown/release`. Removing `wasm32-unknown-unknown` should be enough.

bench/vm/bencher/settings.py
```diff
@@ -0,0 +1,127 @@
...
+
+    "wasmi"  : VMDescriptor(join("target", "release", "examples", "invoke"),
+                            "{wasm_file_path} {function_name}", False),
+
+    "wasmer" : VMDescriptor(join("target", "release", "wasmer"), "run {wasm_file_path}", True),
+
+    "wagon"  : VMDescriptor(join("cmd", "wasm-run"), "wasm_run {wasm_file_path}", False),
+
+    "asmble" : VMDescriptor(join("asmble", "bin", "asmble"),
+                            "invoke -in {wasm_file_path} {function_name} -defmaxmempages 20000", True)
```
RicoGit: 20_000 pages? it is 20000 * 64Kb = 1 280 Mb. Is it really required so many memory? 
What about other VMs how many memory their consume?

PR #303, opened 2018-12-12
Collect network state from smart contract via CLI

DieMyst:


bootstrap/contracts/Deployer.sol
```diff
@@ -163,51 +154,167 @@ contract Deployer is Whitelist {
...
         return (cluster.code.storageHash, cluster.code.storageReceipt, cluster.genesisTime,
-            solverIDs, solverAddrs, solverPorts);
+            cluster.nodeIDs, cluster.nodeAddresses, cluster.ports);
     }
 
     /** @dev Allows to track currently running clusters for specified node's solvers
-     * @param nodeID ID of node (Tendermint consensus key)
+     *  @param nodeID ID of node (Tendermint consensus key)
      */
     function getNodeClusters(bytes32 nodeID)
```
alari: We should mention that this method has O(N) complexity, in order to optimise it later with appropriate data structure.

bootstrap/contracts/Deployer.sol
```diff
@@ -163,51 +154,167 @@ contract Deployer is Whitelist {
...
+                clusterIDs[count++] = cluster.clusterID;
+              }
+            }
         }
-        return clusters;
+        return clusterIDs;
+    }
+
+    /** @dev Gets info about registered clusters
+     * return (cluster IDs, genesis times, storage hashes (Swarm address), receipts, cluster sized for this codes)
```
alari: Is there any way to return struct instead of tuple?

bootstrap/contracts/Deployer.sol
DieMyst: what's the problem if this method does not waste gas

cli/src/contract_status/status.rs
```diff
@@ -0,0 +1,166 @@
...
+
+    let (cluster_ids, genesis_times, storage_hashes, storage_receipts, cluster_sizes): (
+        Vec<H256>,
+        Vec<U256>,
+        Vec<H256>,
+        Vec<H256>,
+        Vec<u64>,
+    ) = utils::query_contract(contract_address, eth_url, "getClustersInfo", (), options)?;
+
+    let options2 = utils::options();
```
folex: I think `utils::options()` should be moved into `query_contract` and argument be replaced by just gas

PR #307, opened 2018-12-17
Network contract

folex:
- Add `Network` contract and move all status and network information getter there
- Web3j 3.6.0 -> 4.0.1
- rxJava 1 -> rxJava 2
- Rename `Deployer` mentions to Fluence or Fluence Contract where appropriate


bootstrap/contracts/Network.sol
```diff
@@ -0,0 +1,181 @@
+pragma solidity ^0.4.0;
+
+import "./Deployer.sol";
+
+contract Network is Deployer {
```
DieMyst: I think we need to doc what is this

PR #313, opened 2018-12-18
Update Webassembly benchmark readme

michaelvoronov:


bench/vm/README.md
```diff
@@ -1,12 +1,12 @@
...
 
-Each test in the benchmark was crafted in a way that would prevent the virtual machine from optimizing it using the dead code elimination. Most tests also prevent intermediate results memoization and allow to specify an RNG seed to obtain repeatable benchmark results. As a semi-mandatory disclaimer, however, we should note that this benchmark is by definition is not comprehensive and might not reflect performance on real life workloads. 
+Each test in the benchmark was crafted in a way that would prevent the virtual machine from optimizing it using the dead code elimination. Most tests also prevent intermediate results memoization and allow the specification of an RNG seed to obtain repeatable benchmark results. As a semi-mandatory disclaimer, however, we should note that this benchmark is not, by definition, comprehensive and might not reflect performance on real life workloads.
```
folex: > allow the specification of an RNG

Maybe something like the following would be more clear?

> Allow to configure user-defined RNG seed

folex: > As a semi-mandatory disclaimer, however, we should note

`semi-mandatory` looks very complex to me, maybe the following would do?

> We should note, however, that this benchmark ...

folex: > might not reflect performance on real life workloads.

Shouldn't it be "***of*** real life workloads"?

bench/vm/README.md
```diff
@@ -33,34 +33,36 @@ We have used the following virtual machines versions (tied to GitHub commits):
 
 All tests were performed on the `m4.large` AWS instance, and the script running benchmarks can be found [here](https://github.com/fluencelabs/fluence/tree/master/bench/vm/bencher).
 
-It should be noted that a total execution time was measured, which skews results a little bit. For example, the compilation time by WAVM is counted, but ideally shouldn't because the compilation should happen only once. The time for JVM to load the class file produced by Asmble is counted as well. We attempted to counteract this by having tests to run long enough, but of course that's far from perfect.
+It should be noted that a total execution time was measured which skews results a little bit. For example, the compilation time by WAVM is counted, but ideally shouldn’t be because the compilation should happen only once. The time for JVM to load the class file produced by Asmble is counted as well. We attempted to counteract this by having tests run long enough, but of course that’s far from perfect.
```
folex: > For example, the compilation time by WAVM is counted,

Mb this would be better:

> For example, in the WAVM case, compilation time is measured, but ideally shouldn't be ...

or

> For example, in the WAVM case, compilation time is included in the (final evaluation / measures / total execution time), but ideally shouldn't be ...

I'm mainly confused by the word `counted`.

folex: > The time for JVM to load the class file produced by Asmble is counted as well.

mb

> Also, JVM class loading is included in the final evaluation as well.

bench/vm/README.md
```diff
@@ -33,34 +33,36 @@ We have used the following virtual machines versions (tied to GitHub commits):
...
 
-It should also be noted that interpreter VMs in general took much longer time than compiler VMs, so we had to repeat the same tests only 3 times for interpreters and 11 times for compilers.
+It should also be noted that in general, interpreter VMs took much longer time than compiler VMs, so we had to repeat the same tests only 3 times for interpreters and 11 times for compilers.
```
folex: > It should also be noted that in general, interpreter VMs took much longer

mb the shorter alternative will do

> As interpreter VMs took much longer time than compiler VMs, we had to repeat the same tests only 3 times for interpreters and 11 times for compilers.

bench/vm/README.md
```diff
@@ -33,34 +33,36 @@ We have used the following virtual machines versions (tied to GitHub commits):
...
 
 ### Results
 
 <p align="center">
   <img src="results/images/compressions.png" alt="Compression tests" width="1015px"/>
 </p>
 
-It should be somewhat expected that "compilers" (WAVM, Asmble, wasmer) will beat "interpreters" (life, wasmi, wagon) without JIT by large margin, and WAVM is a clear winner here. You might also note that `deflate` is one of two tests where Asmble loses to wasmer.
+It should be somewhat expected that “compilers” (WAVM, Asmble, wasmer) will beat “interpreters” (life, wasmi, wagon) without JIT by a large margin, and WAVM is a clear winner here. You might also note that `deflate` is one of two tests where Asmble loses to wasmer.
```
folex: > It should be somewhat expected 

mb change that to shorter and simpler `as`?

> As expected, compilers beat interpreters ...

PR #315, opened 2018-12-19
Fix last port usage and refactor matchWork

folex:
- fix bug when last port was never used
- refactor matchWork for better readability

P.S. If you have an idea how use just a single counter variable instead of having both `i` and `c`, I would be grateful for an advice.


bootstrap/contracts/Deployer.sol
```diff
@@ -165,51 +170,75 @@ contract Deployer is Whitelist {
         returns (bool)
     {
         uint idx = 0;
```
DieMyst: why it is out of `for`?

bootstrap/contracts/Deployer.sol
```diff
@@ -165,51 +170,75 @@ contract Deployer is Whitelist {
...
-            bytes32 nodeID = readyNodes[nodeIndex];
-            Node memory node = nodes[nodeID];
+        // i holds a position in readyNodes array
+        uint i = 0;
 
-            solverIDs[j] = nodeID;
-            solverAddrs[j] = node.nodeAddress;
-            solverPorts[j] = node.currentPort;
-            solverOwners[j] = node.owner;
+        // c holds the number of currently collected nodes and a position in event data arrays
```
DieMyst: maybe it is better to use `i,j,k,n,m` names or names with some meaning instead of random letters?

folex: It was meant to be `clusterSize` or `count`, which I reduced to `c`.

PR #317, opened 2018-12-20
Implement code pinning & node "privatization" mechanics in contract

folex:


bootstrap/contracts/Deployer.sol
```diff
@@ -263,6 +285,136 @@ contract Deployer is Whitelist {
...
+            // ignoring pinned codes
+            if (!pinned && readyNodesCount >= code.clusterSize) {
+                // suitable code found, stop on current idx
+                break;
+            }
+        }
+
+        // check if we hit the condition `readyNodes.length >= code.clusterSize` above
+        // idx >= enqueuedCodes.length means that we skipped through enqueuedCodes array without hitting condition
+        if (idx >= enqueuedCodes.length) {
```
DieMyst: move this check out of this method and from `findPinnedCode`

bootstrap/contracts/Deployer.sol
```diff
@@ -263,6 +285,136 @@ contract Deployer is Whitelist {
...
+            return -1;
+        }
+
+        return int(idx);
```
DieMyst: return uint and compare `idx >= enqueuedCodes.length`

bootstrap/contracts/Deployer.sol
folex: it's pretty common to return index from function that finds element in an array and return -1 in case no element was found.

bootstrap/contracts/Deployer.sol
folex: I think that would break logic incapsulation. How exactly `findPinnedCode` manages it's internal variables should be kept inside `findPinnedCode`.

bootstrap/contracts/Deployer.sol
```diff
@@ -263,6 +285,136 @@ contract Deployer is Whitelist {
...
+
+        bool found = true;
+
+        for (uint i = 0; i < code.pinnedNodes.length; i++) {
+            bytes32 nodeID = code.pinnedNodes[i];
+            Node memory node = nodes[nodeID];
+
+            // node exists and has open ports
+            bool free = node.nodeAddress > 0 && node.currentPort <= node.endPort;
+            found = found && free;
```
alari: Once an exhausted node is found, we can return false -- no need to keep `found` variable.

node/src/main/scala/fluence/node/eth/FluenceContract.scala
```diff
@@ -258,5 +269,10 @@ object FluenceContract {
      * Returns ending port as uint16.
      */
     def endPortUint16: Uint16 = new Uint16(endpoints.maxPort)
+
+    /**
+     * Returns pinned as Bool
+     */
+    def pinnedBool: Bool = new Bool(pinned)
```
alari: Should it be `pinnedBool` or `isPrivateBool`?

bootstrap/contracts/Deployer.sol
```diff
@@ -150,48 +180,43 @@ contract Deployer is Whitelist {
...
 
-        if (!matchWork()) {
+        if (!matchWork(pinnedNodes.length > 0)) {
             emit CodeEnqueued(storageHash);
         }
     }
 
-    /** @dev Checks if there is enough free Solvers for undeployed Code
+    /** @dev Checks if there is enough free Nodes for not yet deployed Codes
+     * @param pinned true if match private code to private nodes where node.owner == code.developer
```
alari: This text is really unclear.

bootstrap/contracts/Deployer.sol
```diff
@@ -58,38 +58,74 @@ contract Deployer {
...
+
+        App app;
+
         uint genesisTime;
+
         bytes32[] nodeIDs;
         bytes24[] nodeAddresses;
         uint16[] ports;
+
+        // TODO explain
```
DieMyst: do todo?

bootstrap/contracts/Deployer.sol
```diff
@@ -129,125 +166,171 @@ contract Deployer {
...
-        enqueuedCodes.push(Code(storageHash, storageReceipt, clusterSize, msg.sender));
+        require(clusterSize >= pinToNodes.length,
+            "number of pinned nodes should be less or equal to the desired clusterSize");
 
-        if (!matchWork()) {
-            emit CodeEnqueued(storageHash);
-        }
+        App memory app = App(storageHash, storageReceipt, clusterSize, msg.sender, pinToNodes);
+
+        tryToDeployApp(app);
```
DieMyst: you could start matching only if the number of ready nodes equal or more than the cluster size

bootstrap/contracts/Deployer.sol
```diff
@@ -129,125 +166,171 @@ contract Deployer {
...
+        uint j = 0;
+        Node memory node;
 
-        // looking for a code that can be deployed given current number of readyNodes
-        for (; idx < enqueuedCodes.length; idx++) {
-            code = enqueuedCodes[idx];
-            if (readyNodes.length >= code.clusterSize) {
-                // suitable code found, stop on current idx
-                break;
+        for(; i < app.pinToNodes.length; i++) {
```
DieMyst: maybe we split this method into several smaller one?

bootstrap/contracts/Deployer.sol
```diff
@@ -129,125 +166,171 @@ contract Deployer {
...
         internal
-        returns (bool)
+    returns(bool)
     {
-        uint idx = 0;
-        Code memory code;
+        Node[] memory solvers = new Node[](app.clusterSize);
+        uint8 solversLength = 0;
+
+        bool skip = false;
```
DieMyst: need comments, what is `skip`, what are parts of this method doing

bootstrap/contracts/Deployer.sol
```diff
@@ -129,125 +166,171 @@ contract Deployer {
...
-        for (; idx < enqueuedCodes.length; idx++) {
-            code = enqueuedCodes[idx];
-            if (readyNodes.length >= code.clusterSize) {
-                // suitable code found, stop on current idx
-                break;
+        for(; i < app.pinToNodes.length; i++) {
+            skip = false;
+            // It should be better then making a custom data structure due to small expected size of pinToNodes
+            for(; j < i && !skip; j++) {
+                skip = app.pinToNodes[i] == app.pinToNodes[j];
```
DieMyst: i == j?

bootstrap/contracts/Deployer.sol
```diff
@@ -129,125 +166,171 @@ contract Deployer {
...
+            }
+            if(skip) {
+                continue;
+            }
+            node = nodes[app.pinToNodes[i]];
+            if(node.owner == msg.sender) {
```
DieMyst: can we throw an error, if not equals?

bootstrap/contracts/Deployer.sol
```diff
@@ -129,125 +166,171 @@ contract Deployer {
...
-    function matchWork()
+    function tryToDeployApp(App memory app)
         internal
-        returns (bool)
+    returns(bool)
     {
-        uint idx = 0;
-        Code memory code;
+        Node[] memory solvers = new Node[](app.clusterSize);
+        uint8 solversLength = 0;
```
DieMyst: not length, but counter

bootstrap/contracts/Deployer.sol
```diff
@@ -129,125 +166,171 @@ contract Deployer {
...
         }
 
-        // check if we hit the condition `readyNodes.length >= code.clusterSize` above
-        // idx >= enqueuedCodes.length means that we skipped through enqueuedCodes array without hitting condition
-        if (idx >= enqueuedCodes.length) {
+        for(j = 0; j < readyNodes.length && solvers.length < app.clusterSize; j++) {
+            node = nodes[readyNodes[j]];
+
+            // Skip all the private nodes; they should be used only with explicit pinning
+            skip = node.isPrivate;
```
DieMyst: why do we need to do this for-loop for a pinned app?

bootstrap/contracts/Deployer.sol
alari: It's a lookup for duplicates, so no, we need to check node ids, not just indices.
And j could never be equal to i due to for loop condition.

bootstrap/contracts/Deployer.sol
```diff
@@ -129,125 +166,196 @@ contract Deployer {
...
+      * @param pinToNodes list of msg.sender's nodes where the App must reside
+      * emits ClusterFormed event when there is enough nodes for the App and
+      * emits AppEnqueued otherwise, subject to change
       */
-    function addCode(bytes32 storageHash, bytes32 storageReceipt, uint8 clusterSize)
+    function addApp(bytes32 storageHash, bytes32 storageReceipt, uint8 clusterSize, bytes32[] pinToNodes)
         external
     {
-        enqueuedCodes.push(Code(storageHash, storageReceipt, clusterSize, msg.sender));
+        require(clusterSize > 0, "Cluster size must be a positive number");
```
DieMyst: clusterSize cannot be negative

bootstrap/contracts/Network.sol
```diff
@@ -38,145 +38,151 @@ contract Network is Deployer {
...
      */
     function getClustersNodes()
     external
     view
-    returns (bytes32[], bytes24[], uint16[], address[])
+    returns (bytes32[], uint16[])
     {
-        BusyCluster[] memory clusters = new BusyCluster[](clusterCount - 1);
-        uint solversCount = 0;
+        Cluster[] memory _clusters = new Cluster[](clusterCount - 1);
```
DieMyst: I think it is better to rename this variable, assuming a lot of mistakes from developers when working with this method.

bootstrap/contracts/Network.sol
```diff
@@ -186,25 +192,28 @@ contract Network is Deployer {
...
+
+        for (uint i = 0; i < nodesIds.length; ++i) {
+            Node memory node = nodes[nodesIds[i]];
             ids[i] = node.id;
             nodeAddresses[i] = node.nodeAddress;
-            startPorts[i] = node.startPort;
-            endPorts[i] = node.endPort;
-            currentPorts[i] = node.currentPort;
+            // TODO remove start ports?
+            //startPorts[i] = node.startPort;
```
DieMyst: I think you need to remove this :)

bootstrap/contracts/Network.sol
```diff
@@ -38,145 +38,151 @@ contract Network is Deployer {
...
-                solversCount++;
+            for (uint n = 0; n < cluster.nodeIDs.length; n++) {
+                ids[workersCount] = cluster.nodeIDs[n];
+                ports[workersCount] = cluster.ports[n];
+                workersCount++;
             }
         }
 
-        return (ids, addresses, ports, owners);
+        return (ids, ports);
```
DieMyst: why this is without addresses and owners?

bootstrap/contracts/Deployer.sol
alari: And it can't be zero. So it must be a positive number.

DieMyst: it is better to throw "Cluster size must be a non-zero number" because it cannot be negative due to type

bootstrap/contracts/Network.sol
alari: Why to collect them? I can't understand it yet, and even can't get the whole meaning of the method. I'll try to get to the point where it's used and come back then.

PR #318, opened 2018-12-24
Sign transactions in CLI

DieMyst:


cli/src/contract_func.rs
```diff
@@ -0,0 +1,164 @@
...
+    where
+        P: Tokenize,
+    {
+        if let Some(p) = password {
+            web3.personal().unlock_account(account, p, None).wait()?;
+        }
+
+        let contract = utils::init_contract(&web3, self.contract_address)?;
+
+        let result_code_publish = contract.call(func, params, account, options);
```
alari: Why `code_publish` is there?

cli/src/contract_func.rs
```diff
@@ -0,0 +1,164 @@
...
+        P: Tokenize,
+        R: Detokenize,
+    {
+        let (_eloop, transport) = web3::transports::Http::new(&self.eth_url)?;
+        let web3 = web3::Web3::new(transport);
+
+        let contract = utils::init_contract(&web3, self.contract_address)?;
+
+        let result_code_publish = contract.query(func, params, None, utils::options(), None);
+        let res = result_code_publish.wait()?;
```
alari: It looks like `result` is enough, isn't it?

cli/src/contract_func.rs
DieMyst: copy-paste, will change it

cli/src/contract_func.rs
DieMyst: yes

cli/src/register.rs
```diff
@@ -102,35 +113,58 @@ impl Register {
...
+                &self.credentials,
                 "addNode",
                 (
                     self.tendermint_key,
                     hash_addr,
                     u64::from(self.min_port),
                     u64::from(self.max_port),
                 ),
-                options,
+                1_000_000,
```
alari: Why this magic number, why not some another one?

tools/run/cleanup.sh
```diff
@@ -2,11 +2,11 @@
 
 case "$(uname -s)" in
    Darwin)
-     docker-compose -f macos/docker-compose.yml kill
+     docker-compose -f docker-compose.yml kill
      ;;
 
    Linux)
-     docker-compose -f linux/docker-compose.yml kill
+     docker-compose -f docker-compose.yml kill
```
alari: Is there any reason to keep the switch cases?

PR #320, opened 2018-12-25
VM examples decoupling

michaelvoronov:


build.sbt
```diff
@@ -146,6 +95,9 @@ lazy val statemachine = (project in file("statemachine"))
 
       val vmDataRoot = "/vmcode"
 
+      val projectRoot = file("").getAbsolutePath
```
alari: Llamadb should not be bundled inside Solver's Docker image. Llamadb is just one example of Fluence applications.

build.sbt
```diff
@@ -250,6 +207,9 @@ lazy val node = project
       val artifact: File = assembly.value
       val artifactTargetPath = s"/${artifact.name}"
 
+      val projectRoot = file("").getAbsolutePath
```
alari: Node image should not contain any wasm code.

build.sbt
```diff
@@ -274,5 +234,12 @@ lazy val node = project
...
-  .enablePlugins(AutomateHeaderPlugin, DockerPlugin)
-  .dependsOn(ethclient, externalstorage)
+  .enablePlugins(
+    AutomateHeaderPlugin,
+    DockerPlugin
+  )
+  .dependsOn(
+    ethclient,
+    externalstorage,
+    `vm-llamadb`
```
alari: Node could not depend on `vm-llamadb`. If you need to have it in tests, you should use `% Test` scope, and then attach it with a Docker volume.

vm/src/test/scala/fluence/vm/AsmleWasmVmSpec.scala
```diff
@@ -57,7 +57,7 @@ class AsmleWasmVmSpec extends WordSpec with Matchers {
 
         val res = for {
           vm <- WasmVm[IO](Seq(sumTestFile))
-          result ← vm.invoke[IO](None, "wrongFnName").toVmError
+          result <- vm.invoke[IO](None, "wrongFnName").toVmError
```
alari: I'm not sure if such changes worth committing.

statemachine/src/main/scala/fluence/statemachine/ServerRunner.scala
```diff
@@ -165,6 +165,26 @@ object ServerRunner extends IOApp with LazyLogging {
...
 
+  /**
+   * Collects and returns all files in given folder.
+   *
+   * @param pathName the name of a folder where files should be listed.
+   * @return a list of files in given directory or provided file if the path to file has has been given
+   */
+  def listFiles(pathName: File): List[File] = pathName match {
+    case file if pathName.isFile => file :: Nil
+    case dir if pathName.isDirectory => {
```
alari: Unnecessary parentheses

statemachine/src/main/scala/fluence/statemachine/ServerRunner.scala
```diff
@@ -165,6 +165,26 @@ object ServerRunner extends IOApp with LazyLogging {
   private def buildVm[F[_]: Monad](moduleFiles: Seq[String]): EitherT[F, StateMachineError, WasmVm] =
     WasmVm[F](moduleFiles).leftMap(VmOperationInvoker.convertToStateMachineError)
 
+  /**
+   * Collects and returns all files in given folder.
+   *
+   * @param pathName the name of a folder where files should be listed.
+   * @return a list of files in given directory or provided file if the path to file has has been given
+   */
+  def listFiles(pathName: File): List[File] = pathName match {
```
alari: As we have some work with filesystem here, this function is not pure, so the result should be wrapped with `IO`.

statemachine/src/main/scala/fluence/statemachine/ServerRunner.scala
```diff
@@ -168,22 +168,18 @@ object ServerRunner extends IOApp with LazyLogging {
...
    */
-  def listFiles(pathName: File): List[File] = pathName match {
-    case file if pathName.isFile => file :: Nil
-    case dir if pathName.isDirectory => {
-      val rawPathNames = dir.list()
-      val pathNames = if (rawPathNames == null) {
-        Nil
-      } else {
-        rawPathNames.toList
+  def listFiles(path: String): IO[List[File]] = IO(
```
alari: Unnecessary braces

statemachine/src/main/scala/fluence/statemachine/ServerRunner.scala
```diff
@@ -200,8 +196,9 @@ object ServerRunner extends IOApp with LazyLogging {
         Try(
           config.moduleFiles
             .map(
-              pathName => listFiles(new File(pathName))
+              path => listFiles(path)
             )
+            .map(_.unsafeRunSync)
```
alari: You should never do this, it just makes IO useless -- you trigger unsafe operation right ahead.

PR #322, opened 2018-12-26
Deploy script

DieMyst:
- Updates deploy script
- CLI can decode base64 tendermint key
- A small script to decode ethereum keystore to a private key

### DIFF
--- a/bootstrap/unlock-keys.js
+++ b/bootstrap/unlock-keys.js
@@ -1,0 +1,26 @@
+/*
+ * Copyright 2018 Fluence Labs Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+// the script gets private key from ethereum keystore
+var keythereum = require("keythereum");
+const password = "PASSWORD";
+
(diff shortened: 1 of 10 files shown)


bootstrap/unlock-keys.js
```diff
@@ -0,0 +1,26 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+// the script gets private key from ethereum keystore
+var keythereum = require("keythereum");
```
alari: What is the purpose for this script? Why do we need it at all? In what scenario should it be used?

cli/src/register.rs
```diff
@@ -39,6 +40,7 @@ const ETH_URL: &str = "eth_url";
 const PASSWORD: &str = "password";
 const SECRET_KEY: &str = "secret_key";
 const WAIT_SYNCING: &str = "wait_syncing";
+const BASE64_TENDERMINT_KEY: &str = "base64";
```
alari: Is it a good value for the Tendermint key argument?

tools/run/compose.sh
```diff
@@ -1,15 +1,67 @@
...
 
    Linux)
      export DOCKER_IP=$(ifconfig docker0 | grep 'inet ' | awk '{print $2}' | grep -Po "[0-9\.]+")
      export HOST_IP=$(ip route get 8.8.8.8 | grep -Po "(?<=src )[0-9\.]+")
-     docker-compose up -dV
      ;;
 esac
+
+# running parity and swarm containers
+docker-compose -f parity.yml up -d
```
alari: Could we have a single docker-compose file with several services inside?

bootstrap/unlock-keys.js
DieMyst: it is needed to decode keystore to the private key until it will be supported in CLI

alari: You should describe it properly here it comments and/or in readme

PR #324, opened 2018-12-29
Change some comments in contracts for better readability

folex:


bootstrap/contracts/Deployer.sol
```diff
@@ -219,17 +220,19 @@ contract Deployer {
 
         // Check that pinToNodes are distinct nodes owned by msg.sender
         for(uint8 i = 0; i < pinToNodes.length; i++) {
+            require(nodes[pinToNodes[i]].owner == msg.sender, "Can pin only to nodes you own");
```
alari: Let's introduce `Node memory nodei` here, to avoid reading from persistent storage when iterating on `j`.

bootstrap/contracts/Deployer.sol
```diff
@@ -243,27 +246,27 @@ contract Deployer {
...
         uint8 workersCount = 0;
 
-        // Prepare an array of workers to collect
+        // Array of workers that will be used to form a cluster
         Node[] memory workers = new Node[](app.clusterSize);
 
         // There must be enough readyNodes to try to deploy the app
         if(readyNodes.length >= app.clusterSize - app.pinToNodes.length) {
-            // Index up to pinToNodes length
+            // Index in pinToNodes
```
alari: This comment is wrong, as this index is used not only for iterating in `pinToNodes`. The only specific thing about it is its type (uint8).

bootstrap/contracts/Deployer.sol
```diff
@@ -276,15 +279,18 @@ contract Deployer {
             for(uint j = 0; j < readyNodes.length && workersCount < app.clusterSize; j++) {
                 node = nodes[readyNodes[j]];
 
-                // Deduplicate nodes in case a pinTo node is not private
+                // True if node is already in workers array. That could happen if
```
alari: Please add the comment about the purpose of what's going on (deduplication).

bootstrap/contracts/Network.sol
```diff
@@ -57,7 +61,7 @@ contract Network is Deployer {
         return nodes[nodeID].clusters;
     }
 
-    /** @dev Allows anyone with clusterID to retrieve assigned App
+    /** @dev Allows anyone with clusterID to retrieve assigned App and other cluster info
```
alari: We have no allowance logic here. The method just retrieves an assigned app.

bootstrap/contracts/Network.sol
```diff
@@ -82,6 +86,9 @@ contract Network is Deployer {
         );
     }
 
+    /** @dev Returns addresses and ports of cluster's real-time workers
```
alari: I think we shouldn't use `real-time` term here.

bootstrap/contracts/Deployer.sol
```diff
@@ -220,11 +220,13 @@ contract Deployer {
 
         // Check that pinToNodes are distinct nodes owned by msg.sender
         for(uint8 i = 0; i < pinToNodes.length; i++) {
-            require(nodes[pinToNodes[i]].owner == msg.sender, "Can pin only to nodes you own");
+            bytes32 nodeID_i = pinToNodes[i];
+            require(nodes[nodeID_i].owner != 0, "Can pin only to registered nodes");
```
alari: Here you read `nodes[nodeID_i]` from persistent storage twice.

PR #332, opened 2019-01-09
Able to deploy locally with parity node in dev mode

DieMyst:


tools/deploy/fabfile.py
```diff
@@ -9,7 +9,7 @@
 
 # change this if you want to deploy on another hosts
 # TO USE: replace values inside <> with your actual values
```
alari: This comment is outdated, could you update it please to make it clear how to use this magic `info.keys()` thing?

tools/deploy/scripts/compose.sh
```diff
@@ -2,6 +2,18 @@
 
 set -e
 
+if [ -z "$PROD" ]
+then
+    export NAME='node1'
+    export PORTS='25000:25003'
+    export OWNER_ADDRESS=0x00a329c0648769a73afac7f9381e08fb43dbea72
```
alari: Why is it hardcoded? You're setting it in `fabfile.py`, and hardcode here, is it ok?

tools/deploy/scripts/deploy-contract.js
```diff
@@ -0,0 +1,36 @@
...
+// Get encoded request for deploying contract
+let encodedAbi = NetworkContract.deploy({
+    data: bytecode
+}).encodeABI();
+
+// Send transaction and print the result
+web3.eth.sendTransaction({ from: acc, data: encodedAbi, gas: '4500000' }, "")
+.on('error', console.error)
+.then(function(newContractInstance){
+    console.log(newContractInstance.contractAddress) // instance with the new contract address
```
alari: Let's print something useful here.

DieMyst: it printed as result for the bash script

tools/deploy/scripts/compose.sh
DieMyst: it is for dev mode. Run `compose.sh` directly will run containers locally in dev mode.

alari: Could you add a comment about it here? About why it's a constant

PR #334, opened 2019-01-10
Hardcode safe nightly build

folex:
I'll revert `nightly-2019-01-08` back to just `nightly` after https://github.com/rust-lang/rust/issues/57488 is resolved

### DIFF
--- a/project/SbtCommons.scala
+++ b/project/SbtCommons.scala
@@ -35,19 +35,20 @@ object SbtCommons {
 
   def rustVmExample(exampleName: String) = Seq(
     publishArtifact := false,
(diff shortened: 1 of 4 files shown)


project/SbtCommons.scala
```diff
@@ -35,19 +35,20 @@ object SbtCommons {
 
   def rustVmExample(exampleName: String) = Seq(
     publishArtifact := false,
-    test := (test in Test).dependsOn(compile).value,
-    compile := (compile in Compile).dependsOn(Def.task {
+    test            := (test in Test).dependsOn(compile).value,
```
michaelvoronov: Looks like wrong indentation

PR #335, opened 2019-01-10
Update contract status for a browser for new contract API

DieMyst:


tools/deploy/scripts/compose.sh
```diff
@@ -73,4 +81,4 @@ echo "TENDERMINT_KEY="$TENDERMINT_KEY
 
 # check if node is already registered
 # todo build fluence CLI in fly, use cargo from cli directory, or run from target cli directory?
-./fluence register $HOST_IP $TENDERMINT_KEY $OWNER_ADDRESS $CONTRACT_ADDRESS -s $PRIVATE_KEY --wait_syncing --base64_tendermint_key
+./fluence register $HOST_IP $TENDERMINT_KEY $OWNER_ADDRESS $CONTRACT_ADDRESS -s $PRIVATE_KEY --wait_syncing --min_port $MIN_PORT --max_port $MAX_PORT --base64_tendermint_key
```
folex: You're passing `$HOST_IP` here as an external IP address, but it's not external. 

For linux it's the IP of the default tx iface (i.e., iface that serves outbound traffic).

For mac it's string `host.docker.internal` which doesn't parse to `IpAddr` in rust cli, so cli panics. Also, it will resolve to different IPs on different computers, so it should be used when all nodes are registered from the same computer.

tools/deploy/scripts/compose.sh
```diff
@@ -4,12 +4,12 @@ set -e
 
 # `PROD` variable is assigned in `fabfile.py`, so if run `compose.sh` directly,
```
folex: PROD -> REMOTE_DEPLOY

tools/deploy/scripts/deploy-contract.js
```diff
@@ -17,20 +19,34 @@ let bytecode = '0x' + sourceBin;
...
+
+// request to compose partial transaction to full transaction
+let composeRequest = {method:"parity_composeTransaction",params:[partialTransaction],"id":1,"jsonrpc":"2.0"};
+
+request.post(
+    'http://localhost:8545/',
+    { json: composeRequest },
+    function (error, response, body) {
+        // sign transaction with private key
+        web3.eth.accounts.signTransaction(body.result, '0x4d5db4107d237df6a3d58ee5f70ae63d73d7658d4026f2eefd2f204c81682cb7')
```
folex: Where is that key from? Let's add some docs!

PR #341, opened 2019-01-15
Pinning, app deletion, clusters list in status

folex:
In CLI:
- implemented pinning
- implemented app deletion
- added list of clusters to status

In smart-contract:
- implemented app deletion

Tests for CLI will be in the next PR


bootstrap/contracts/Deployer.sol
```diff
@@ -232,15 +242,55 @@ contract Deployer {
...
+    /** @dev Deletes app with appID either from enqueued apps or from clusters mapping
+      * You must be app's owner to delete it. Currently, nodes' ports aren't freed.
+      * @param appID app to be deleted
+      * @param clusterID (optional) if specified, cluster with app will be deleted. Otherwise, delete from enqueued apps.
+      * emits AppDeleted event on successful deletion
+      * reverts if you're not app owner
+      * reverts if app or cluster aren't found
+      * TODO: free nodes' ports after app deletion
+      */
+    function deleteApp(bytes32 appID, bytes32 clusterID)
```
DieMyst: I think it is better to use two methods `deleteApp` and `deleteCluster` and two events. 
Effects, after deleting an app and deleting a cluster, is different and should be processed in a different manner.

bootstrap/contracts/Deployer.sol
```diff
@@ -410,4 +460,37 @@ contract Deployer {
...
+        uint8 index = 0;
+        uint len = clustersIds.length;
+        for (; index < len; index++) {
+            if (clustersIds[index] == clusterID) {
+                break;
+            }
+        }
+
+        // flag we didn't find such clusterID
+        if (index >= len) return false;
```
DieMyst: is it possible to return `false` here? Do we need to go through the whole array for this?

bootstrap/contracts/Network.sol
```diff
@@ -149,7 +150,7 @@ contract Network is Deployer {
             }
         }
 
-        return (storageHashes, storageReceipts, clusterSizes, owners, numberOfPinnedNodes, allPinToNodes);
+        return (storageHashes, appIDs, clusterSizes, owners, numberOfPinnedNodes, allPinToNodes);
```
DieMyst: I think you should add TODO here, because we will use receipts in future

bootstrap/contracts/Deployer.sol
folex: Yes, if such `clusterID` wasn't found in `clustersIds`. Currently it's impossible, because `removeCluster` is called only once and after checking that such cluster exists in `clusters` mapping, but `removeCluster`'s interface doesn't state about that, so I implemented a more general approach.

cli/src/publisher.rs
```diff
@@ -126,37 +130,61 @@ impl Publisher {
...
+    let secret_key = utils::parse_secret_key(matches, SECRET_KEY)?;
     let password = matches.value_of(PASSWORD).map(|s| s.to_string());
 
     let credentials = Credentials::get(secret_key, password);
 
-    let cluster_size: u8 = matches.value_of(CLUSTER_SIZE).unwrap().parse()?;
+    let cluster_size = value_t!(matches, CLUSTER_SIZE, u8)?;
 
-    let mut file = File::open(path)?;
+    let gas: u32 = matches.value_of(GAS).unwrap().parse()?;
```
DieMyst: why not value_t?

bootstrap/contracts/Deployer.sol
folex: Done

bootstrap/contracts/Deployer.sol
```diff
@@ -140,6 +144,12 @@ contract Deployer {
     // Emitted on every new Node
     event NewNode(bytes32 id);
 
+    // Emitted when app is removed from enqueuedApps by owner
+    event AppDequeued(bytes32 appID);
```
alari: Usually `dequeue` means "take the next object in a queue", not "remove an enqueued object by id". Could we find a better name?

bootstrap/contracts/Deployer.sol
```diff
@@ -140,6 +144,12 @@ contract Deployer {
...
+
+    // Emitted when app & cluster were removed by app owner
+    event AppDeleted(bytes32 appID, bytes32 clusterID);
```
alari: I suppose `appID` is enough, and `clusterID` could be removed from the contract.

bootstrap/contracts/Deployer.sol
```diff
@@ -232,15 +245,66 @@ contract Deployer {
...
+      * reverts if app or cluster aren't not found
+      * TODO: free nodes' ports after app deletion
+      */
+    function deleteApp(bytes32 appID, bytes32 clusterID)
+        external
+    {
+        Cluster memory cluster = clusters[clusterID];
+        require(cluster.clusterID != 0, "error deleting app: cluster not found");
+        require(cluster.app.appID == appID, "error deleting app: cluster hosts another app");
+        require(cluster.app.owner == msg.sender, "error deleting app: you must own app to delete it");
```
alari: You must own `the` app?

bootstrap/contracts/Network.sol
alari: Or revert the storageReceipts?

bootstrap/contracts/Deployer.sol
folex: Not necessary, but why not

bootstrap/contracts/Deployer.sol
```diff
@@ -232,15 +238,65 @@ contract Deployer {
...
         }
     }
 
+    /** @dev Deletes app with appID from enqueued apps
+      * You must be app's owner to delete it. Currently, nodes' ports aren't freed.
+      * @param appID app to be deleted
+      * emits AppDequeued event on successful deletion
+      * reverts if you're not app owner
+      * reverts if app not found
+      * TODO: free nodes' ports after app deletion
```
DieMyst: delete todo here

bootstrap/contracts/Network.sol
folex: Added TODO. storageReceipts deleted intentionally to mitigate `stack is too deep` Solidity compilation error. When we find a use for storageReceipts in enqueuedApps, I'll squeeze them back.

bootstrap/contracts/Deployer.sol
folex: removed

bootstrap/contracts/Deployer.sol
folex: why? It's a good TODO

bootstrap/contracts/Deployer.sol
folex: Discussed in chat that dequeue is good for this purpose

bootstrap/contracts/Deployer.sol
```diff
@@ -398,18 +386,18 @@ contract Deployer {
...
-            nodes[node.id].clusters.push(clusterID);
+            nodes[node.id].apps.push(app.appID);
         }
 
         uint genesisTime = now;
 
         // saving selected nodes as a cluster with assigned app
-        clusters[clusterID] = Cluster(clusterID, app, genesisTime, nodeIDs, workerPorts);
-        clustersIds.push(clusterID);
+        clusters[app.appID] = Cluster(app, genesisTime, nodeIDs, workerPorts);
```
alari: - App is a part of Cluster, which is referenced by app ID -- seems inconsistent
- Here you remove app from persistent storage, then put it back on another address -- it consumes too much gas

bootstrap/contracts/Deployer.sol
```diff
@@ -148,18 +155,17 @@ contract Deployer {
...
-    // Number of existing clusters, used for clusterID generation
-    // TODO find a better way to generate cluster id
-    uint256 internal clusterCount = 1;
+    // mapping of appID to App
+    mapping(bytes32 => App) internal apps;
+    // Store app ids to traverse clusters mapping
+    bytes32[] public appIDs;
 
     // Apps waiting for nodes
     // TODO: should they have IDs? so that app owner could cancel deployment of enqueued app, before cluster gets formed
```
alari: As now they have IDs, let's remove this comment

bootstrap/contracts/Deployer.sol
```diff
@@ -75,14 +75,16 @@ contract Deployer {
         // True if this node can be used only by `owner`
         bool isPrivate;
 
-        // Clusters where this node participate
-        bytes32[] clusters;
+        // Apps hosted by this node
+        bytes32[] apps;
```
alari: Shouldn't it be something like `appIDs` for consistency?

bootstrap/contracts/Deployer.sol
```diff
@@ -148,18 +155,17 @@ contract Deployer {
...
-    uint256 internal clusterCount = 1;
+    // mapping of appID to App
+    mapping(bytes32 => App) internal apps;
+    // Store app ids to traverse clusters mapping
+    bytes32[] public appIDs;
 
     // Apps waiting for nodes
     // TODO: should they have IDs? so that app owner could cancel deployment of enqueued app, before cluster gets formed
-    App[] internal enqueuedApps;
+    bytes32[] internal enqueuedApps;
```
alari: Do we really need it to be `internal`? As it's just IDs now, it could be useful to traverse it via web3 or with some other contract.

PR #343, opened 2019-01-15
Ethclient refactoring

alari:


ethclient/src/main/scala/fluence/ethclient/EthClientApp.scala
```diff
@@ -31,19 +31,25 @@ object EthClientApp extends IOApp {
...
           _ ← IO(println("Launching w3j"))
 
-          unsubscribe ← Deferred[IO, Either[Throwable, Unit]]
+          isSyncing ← ethClient.isSyncing[IO].map(_.isSyncing)
+          _ ← IO(println(s"isSyncing: $isSyncing"))
 
           version ← ethClient.clientVersion[IO]()
           _ = println(s"Client version: $version")
+
+          _ ← ethClient.blockStream[IO]().map(println).compile.drain
```
folex: add some marker to println, so it's easy to detect where logs are coming from

alari: I'm not sure it worth doing in this PR, as EthClientApp exists only for development/testing purposes (there's no way to mix its output with something else) and it is going to be rewritten in the following commits, while I'm working on passing the logs to Workers

PR #348, opened 2019-01-19
Remove workers upon receiving AppDeleted in Fluence Node

folex:
- Refactor `WorkerParams`, mainly remove `ClusterData` and related stuff
- Separate data model for Tendermint config from application data models
- Fix workers healthcheck: parsing, workers removal from healthcheck on stop and so on
- Make `workers` map indexed by `appId` instead of `WorkerParams` case class
- Actually stop worker containers on `stop`
- Separate `DockerParams` into `ExecParams` and `DaemonParams`, so it's possible to manage their resources differently
    - also note `unmanagedDaemonRun`, it's a hackity-hack, and I don't like it, but it allows to run daemonized Docker containers and stop them manually at the right point
- Handle `AppDeleted` event and stop mentioned workers


node/src/main/scala/fluence/node/MasterNode.scala
```diff
@@ -45,104 +44,25 @@ case class MasterNode(
...
-    val validator = "priv_validator.json"
-
-    IO {
-      logger.info(s"Copying keys to worker: ${from.resolve(nodeKey)} -> ${to.resolve(nodeKey)}")
-      Files.copy(from.resolve(nodeKey), to.resolve(nodeKey), REPLACE_EXISTING)
-
-      logger.info(s"Copying priv_validator to worker: ${from.resolve(validator)} -> ${to.resolve(validator)}")
-      Files.copy(from.resolve(validator), to.resolve(validator), REPLACE_EXISTING)
-    }
+  private val prepareWorkerParams: Pipe[IO, App, WorkerParams] = {
```
DieMyst: docs?

node/src/main/scala/fluence/node/MasterNode.scala
```diff
@@ -45,104 +44,25 @@ case class MasterNode(
...
+      codeManager
+    )
   }
 
   /**
    * Runs MasterNode. Returns when contract.getAllNodeClusters is exhausted
    * TODO: add a way to cleanup, e.g. unsubscribe and stop
    */
-  val run: IO[ExitCode] =
+  private val masterNode: IO[ExitCode] =
```
DieMyst: in docs this method "runs MasterNode", but the name isn't show this

node/src/main/scala/fluence/node/MasterNodeApp.scala
```diff
@@ -118,14 +119,16 @@ object MasterNodeApp extends IOApp with LazyLogging {
       .attempt
       .flatMap {
         case Left(err) =>
-          logger.error("Error: {}", err)
+          logger.error(s"Error in MasterNodeApp: $err", err)
+          err.printStackTrace(System.err)
```
DieMyst: is there a way to do it in logger?

node/src/main/scala/fluence/node/eth/App.scala
```diff
@@ -0,0 +1,105 @@
...
+      .zip(addresses)
+      .zip(ps)
+      .zipWithIndex
+      .map {
+        case (((validator, address), port), idx) =>
+          WorkerNode(validator, address, port, idx)
+      }
+      .toList
+
+    val currentWorker = workers.find(_.validatorKey == currentWorkerId)
```
DieMyst: why cluster depends on the current worker? it doesn't clear. Maybe the naming is wrong.

node/src/main/scala/fluence/node/eth/FluenceContract.scala
```diff
@@ -59,39 +58,39 @@ class FluenceContract(private val ethClient: EthClient, private val contract: Ne
...
    *
-   * @param nodeConfig Node to pick validatorKey from to lookup the clusters for
+   * @param workerId Tendermint validator key identifying this node
    * @tparam F Effect
    */
-  def getNodeAppIds[F[_]](nodeConfig: NodeConfig)(implicit F: Async[F]): F[List[Bytes32]] =
+  def getNodeAppIds[F[_]](workerId: Bytes32)(implicit F: Async[F]): F[List[Bytes32]] =
     contract
-      .getNodeApps(nodeConfig.validatorKeyBytes32)
+      .getNodeApps(workerId)
```
DieMyst: do `attempt` here and manage an error, if there is no connection. It is useful for understanding errors on the start of the application.

node/src/main/scala/fluence/node/eth/App.scala
```diff
@@ -0,0 +1,105 @@
...
+case class Cluster(genesisTime: FiniteDuration, workers: List[WorkerNode], currentWorker: WorkerNode)
+
+object Cluster {
+
+  /* Builds a Cluster structure, filters for clusters that include current node.
+   * i.e., return None if validatorKeys doesn't contain currentWorkerId
+   * @param time Cluster genesis time
+   * @param validatorKeys array of 32-byte Tendermint validator keys, used as workers' ids
+   * @param ports array of ports for Tendermint p2p connection. For rpc and prometheus ports see [[WorkerNode]]
+   * @param currentWorkerId 32-byte Tendermint validator key corresponding to current node
```
DieMyst: maybe `ownedWorker` or something like that?

node/src/main/scala/fluence/node/eth/FluenceContract.scala
DieMyst: or something like `catchError` instead of `attempt`

node/src/main/scala/fluence/node/tendermint/config/GenesisConfig.scala
```diff
@@ -0,0 +1,60 @@
...
+  app_hash: String,
+  validators: List[ValidatorConfig]
+)
+private[config] object GenesisConfig {
+  import io.circe.Encoder
+  import io.circe.generic.semiauto.deriveEncoder
+  import io.circe.syntax._
+
+  def buildFrom(app: App): String = {
+    val df = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'")
```
DieMyst: what is `df`?

node/src/main/scala/fluence/node/tendermint/config/TendermintConfig.scala
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+  def writeGenesis(app: App, dest: Path) = IO {
+    val df = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'")
+    df.setTimeZone(TimeZone.getTimeZone("UTC"))
+
+    val genesis = GenesisConfig.buildFrom(app)
+
+    logger.info("Writing {}/genesis.json", dest)
+    Files.write(dest.resolve("genesis.json"), genesis.getBytes)
```
DieMyst: write logs after finishing the operation too, if you logging the starting the operation.

node/src/main/scala/fluence/node/tendermint/config/TendermintConfig.scala
```diff
@@ -0,0 +1,133 @@
...
+    val persistentPeers = app.cluster.workers.map(_.peerAddress).mkString(",")
+
+    val lines = Source.fromFile(configSrc.toUri).getLines().map {
+      case s if s.contains("external_address") => s"""external_address = "${currentWorker.address}""""
+      case s if s.contains("persistent_peers") => s"""persistent_peers = "$persistentPeers""""
+      case s if s.contains("moniker") => s"""moniker = "${app.appId}_${currentWorker.index}""""
+      case s => s
+    }
+
+    Files.write(configDest, lines.toIterable.asJava)
```
DieMyst: write logs

node/src/main/scala/fluence/node/tendermint/config/TendermintConfig.scala
```diff
@@ -0,0 +1,133 @@
...
+    val df = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'")
+    df.setTimeZone(TimeZone.getTimeZone("UTC"))
+
+    val genesis = GenesisConfig.buildFrom(app)
+
+    logger.info("Writing {}/genesis.json", dest)
+    Files.write(dest.resolve("genesis.json"), genesis.getBytes)
+  }
+
+  def updateConfigTOML(app: App, workerId: Bytes32, configSrc: Path, configDest: Path) = IO {
```
DieMyst: add type

node/src/main/scala/fluence/node/workers/Worker.scala
```diff
@@ -93,7 +93,7 @@ object Worker extends LazyLogging {
           WorkerRunning(uptime, info)
         case Left(err) ⇒
           logger.error("Worker HTTP check failed: " + err.getLocalizedMessage, err)
-          WorkerHttpCheckFailed(StoppedWorkerInfo(params), err)
+          WorkerHttpCheckFailed(StoppedWorkerInfo.fromWorker(params.currentWorker), err)
```
DieMyst: `currentWorker` is a bad name. Need to think about a more clear name.

node/src/main/scala/fluence/node/workers/WorkerParams.scala
```diff
@@ -15,43 +15,45 @@
...
-      )
-      .option("--name", clusterData.nodeInfo.nodeName)
-      .image(image.imageName)
+  val dockerCommand: DockerParams.DaemonParams = {
+    val params = DockerParams
+      .build()
+      .option("-e", s"""CODE_DIR=$vmCodePath""")
+      .option("-e", s"""WORKER_DIR=$workerPath""")
+      .option("--name", s"${appIdHex}_worker_${currentWorker.index}")
+      .port(currentWorker.p2pPort, 26656)
```
DieMyst: why these ports are hardcoded?

node/src/main/scala/fluence/node/docker/DockerIO.scala
```diff
@@ -39,25 +39,38 @@ object DockerIO extends LazyLogging {
     implicitly[ContextShift[F]].shift *> Sync[F].delay(fn)
 
   /**
-   * Runs a docker container, providing a single String with the container ID.
-   * Calls `rm -f` on that ID when stream is over.
+   * Runs a temporary docker container with custom executable. Returns stdout of execution as a string.
```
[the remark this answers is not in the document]
folex: Because the idea is that `Exec` containers are created only for a single command execution, and then they're auto-killed via `--rm` flag in `docker run`. `unmanagedDaemonRun` is an exception from this rule, but I hope we will get rid of it.

node/src/main/scala/fluence/node/eth/App.scala
folex: Because it's a common task to query this DTO for a worker which belongs to current Fluence node. I saved it as a field here to avoid handling Option's everywhere. 

It's kinda leaky, but it makes sense to store info about which cluster member belongs to current node.

node/src/main/scala/fluence/node/eth/App.scala
folex: Nope, I'd better stick with `current`. `owned` is a really overloaded word. I think `current` does good job in conveying the meaning of being run on the same node where that code is running.

node/src/main/scala/fluence/node/tendermint/config/TendermintConfig.scala
folex: Why? It's way too much logs. An error will be printed out anyway.

node/src/main/scala/fluence/node/tendermint/config/GenesisConfig.scala
```diff
@@ -0,0 +1,60 @@
...
+  chain_id: String,
+  app_hash: String,
+  validators: List[ValidatorConfig]
+)
+private[config] object GenesisConfig {
+  import io.circe.Encoder
+  import io.circe.generic.semiauto.deriveEncoder
+  import io.circe.syntax._
+
+  def buildFrom(app: App): String = {
```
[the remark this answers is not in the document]
alari: It could be something like `toGenesisConfigString`.

node/src/main/scala/fluence/node/tendermint/config/TendermintConfig.scala
folex: Same. I think success logs should be used for debugging only, and since that part of code is already debugged, these logs are enough.

node/src/main/scala/fluence/node/tendermint/config/GenesisConfig.scala
```diff
@@ -0,0 +1,60 @@
...
+
+package fluence.node.tendermint.config
+import java.text.SimpleDateFormat
+import java.util.TimeZone
+
+import fluence.ethclient.helpers.Web3jConverters
+import fluence.node.eth.App
+import fluence.node.tendermint.json.ValidatorKey
+
+private[config] case class ValidatorConfig(pub_key: ValidatorKey, power: String, name: String)
```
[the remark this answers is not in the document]
alari: Another option is to move these intermediary classes inside the companion object where they are used, marking them private.

node/src/main/scala/fluence/node/workers/Worker.scala
folex: any ideas? I like `currentWorker`

node/src/main/scala/fluence/node/tendermint/config/TendermintConfig.scala
```diff
@@ -0,0 +1,133 @@
...
+   *
+   * At the end of execution `workerPath` will contain:
+   *    - vm code at `codePath`
+   *    - tendermint configuration in `workerConfigDir`:
+   *        - node_key.json, containing private P2P key
+   *        - priv_validator.json, containing validator's private & public keys and it's address
+   *        - genesis.json, generated from [[App.cluster]] and [[App.appId]]
+   *        - config.toml, copied from `templateConfigDir/default_config.toml` and updated
+   */
+  def prepareWorkerParams(
```
alari: This seems as a mixed responsibility. Maybe have everything related to Tendermint itself there, and its composition somewhere else, like in `WorkerConfig` class?

Especially I don't like to have `codePath` handling there.

build.sbt
```diff
@@ -194,6 +194,8 @@ lazy val node = project
     test in Test := (test in Test)
       .dependsOn(docker)
       .dependsOn(docker in statemachine)
+      .dependsOn(compile in `vm-llamadb`)
+      .dependsOn(compile in Test)
```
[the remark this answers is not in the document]
folex: Yes, but without that line, compilation happens only after Docker container build, and it's painful to wait for containers just to see a small compilation error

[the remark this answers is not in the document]
folex: I'll add a comment

node/src/main/scala/fluence/node/workers/Worker.scala
alari: I'd agree with some need for renaming.

node/src/main/scala/fluence/node/workers/WorkerParams.scala
```diff
@@ -15,43 +15,45 @@
...
+  currentWorker: WorkerNode,
   workerPath: String,
   vmCodePath: String,
   masterNodeContainerId: Option[String],
   image: WorkerImage
 ) {
 
-  override def toString =
-    s"(worker ${clusterData.nodeInfo.node_index} with port $rpcPort for ${clusterData.nodeInfo.clusterName})"
+  private val appIdHex: String = bytes32ToHexStringTrimZeros(appId)
```
alari: Add some comments please about why do we need it and why this formatting is used

node/src/main/scala/fluence/node/MasterNode.scala
folex: Ok, I renamed to `runMasterNode`

node/src/main/scala/fluence/node/eth/FluenceContract.scala
folex: I think it's a good idea! But it should be done for all operations with `web3` and it's a good task for a separate PR.

node/src/main/scala/fluence/node/workers/WorkersPool.scala
```diff
@@ -25,18 +25,19 @@ import cats.syntax.applicativeError._
...
+import org.web3j.abi.datatypes.generated.Bytes32
 import slogging.LazyLogging
 
 import scala.language.higherKinds
 
 /**
  * Wraps several [[Worker]]s in a pool, providing running and monitoring functionality.
  *
- * @param workers a storage for running [[Worker]]s
+ * @param workers a storage for running [[Worker]]s, indexed by appIds
```
alari: Is `Bytes32` a case class? Is it immutable? Is it definitely safe to use as a key?

node/src/main/scala/fluence/node/tendermint/config/GenesisConfig.scala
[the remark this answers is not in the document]
folex: Renamed to `generateJson`

node/src/main/scala/fluence/node/workers/WorkerParams.scala
folex: These are internal worker's ports, they're the same inside every container

node/src/main/scala/fluence/node/workers/WorkersPool.scala
```diff
@@ -102,11 +103,25 @@ class WorkersPool[F[_]: ContextShift: Timer](
...
    * @param P [[Parallel]] instance is required as all workers are stopped concurrently
    */
-  def healths[G[_]](implicit P: Parallel[F, G]): F[Map[WorkerParams, WorkerInfo]] =
+  def healths[G[_]](implicit P: Parallel[F, G]): F[Map[WorkerParams, WorkerHealth]] =
     for {
       workersMap ← workers.get
       workersHealth ← Parallel.parTraverse(workersMap.values.toList)(s ⇒ s.healthReport.map(s.params → _))
     } yield workersHealth.toMap
+
+  def stopWorkerForApp(appId: Bytes32): F[Unit] =
```
alari: Docs needed.

node/src/main/scala/fluence/node/workers/WorkersPool.scala
```diff
@@ -102,11 +103,25 @@ class WorkersPool[F[_]: ContextShift: Timer](
...
+      worker = map.get(appId)
+      _ <- worker
+        .map(
+          w =>
+            stop(w).attempt.map { stopped =>
+              logger.info(s"Stopped: ${w.params} => $stopped")
+          }
+        )
+        .getOrElse(F.unit)
+    } yield {}
```
alari: Please don't do it ;)

node/src/test/scala/fluence/util/util.scala
```diff
@@ -0,0 +1,160 @@
...
+import org.scalatest.time.Span
+import org.scalatest.{Timer => _}
+
+import scala.concurrent.duration._
+import scala.io.Source
+import scala.language.higherKinds
+import scala.sys.process.{Process, ProcessLogger}
+import scala.util.Try
+
+package object util {
```
alari: Let's move it to a trait or a set of traits dedicated to particular responsibilities instead. Package objects with a huge load of everything useful, named a meaningless word like `util`, looks as a good idea at first, but it proved to be unmaintainable in the future.

ethclient/src/main/scala/fluence/ethclient/helpers/Web3jConverters.scala
```diff
@@ -80,6 +101,13 @@ object Web3jConverters {
   def hexToBinary(hex: String): Array[Byte] =
     ByteVector.fromHex(hex, Bases.Alphabets.HexUppercase).map(_.toArray).getOrElse(new Array[Byte](hex.length / 2))
 
+  /**
+   * Converts hex string to Bytes32.
+   *
+   * @param hex hex string
+   */
+  def hexToBytes32(hex: String): Bytes32 = new Bytes32(hexToBinary(hex))
```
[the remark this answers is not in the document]
folex: Nice catch! I'll wrap it in either

node/src/main/scala/fluence/node/workers/WorkersPool.scala
folex: Apparently it's mutable, missed it. Good catch!

node/src/main/scala/fluence/node/MasterNode.scala
```diff
@@ -152,4 +72,25 @@ case class MasterNode(
...
+      .drain
+
+  /**
+   * Runs master node and starts listening for AppDeleted event in different threads,
+   * then joins the threads and returns back exit code from master node
+   */
+  val run: IO[ExitCode] = for {
+    node <- Concurrent[IO].start(masterNode)
+    appDelete <- Concurrent[IO].start(listenForDeletion)
+    exitCode <- node.join
```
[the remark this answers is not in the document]
folex: I'll cancel `appDelete` after `node.join` finishes. Otherwise `runMasterNode` could exit with an error, and still application will run forever

node/src/main/scala/fluence/node/docker/DockerIO.scala
```diff
@@ -39,25 +39,38 @@ object DockerIO extends LazyLogging {
...
+   * Caller is responsible for container removal.
+   * @param params parameters for Docker container
+   * @return a stream with execution stdout
+   */
+  def run[F[_]: Sync: ContextShift](params: DockerParams.ExecParams): fs2.Stream[F, String] =
```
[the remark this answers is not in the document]
folex: Renamed to `exec`

node/src/main/scala/fluence/node/eth/FluenceContract.scala
```diff
@@ -219,7 +219,9 @@ object FluenceContract {
   ): Option[App] = {
     val cluster =
       Cluster.build(event.genesisTime, event.nodeIDs, event.nodeAddresses, event.ports, currentWorkerId = workerId)
-    cluster.map(App(event.appID, event.storageHash, _))
+    val app = cluster.map(App(event.appID, event.storageHash, _))
+    println(s"app is $app")
```
alari: Use logging?

node/src/test/scala/fluence/node/TendermintSetup.scala
```diff
@@ -0,0 +1,46 @@
...
+import cats.effect._
+import cats.syntax.functor._
+import fluence.node.eth.WorkerNode
+import org.scalatest.{Timer => _}
+
+import scala.io.Source
+import scala.language.higherKinds
+import scala.util.Try
+
+trait TendermintSetup {
```
alari: - It doesn't set up Tendermint by any mean
- Documentation is missing

node/src/test/scala/fluence/node/OsSetup.scala
```diff
@@ -0,0 +1,68 @@
...
+import java.io.File
+import java.net.InetAddress
+
+import cats.syntax.functor._
+import org.scalatest.{Timer => _}
+
+import scala.language.higherKinds
+import scala.sys.process.{Process, ProcessLogger}
+
+trait OsSetup {
```
alari: - It doesn't set up anything
- Documentation is missed

node/src/test/scala/fluence/node/GanacheSetup.scala
```diff
@@ -0,0 +1,40 @@
...
+ * limitations under the License.
+ */
+
+package fluence.node
+import cats.syntax.functor._
+import org.scalatest.{Timer => _}
+
+import scala.language.higherKinds
+
+trait GanacheSetup extends OsSetup with slogging.LazyLogging {
```
alari: - This doesn't set up anything actually. To set smth up, it should override `beforeAll`/`afterAll`, for example. In this case, it just provides some lifecycle functions
- Documentation is missed

node/src/test/scala/fluence/node/DockerSetup.scala
```diff
@@ -0,0 +1,56 @@
...
+package fluence.node
+
+import cats.effect._
+import cats.syntax.functor._
+import fluence.node.docker.{DockerIO, DockerParams}
+import org.scalatest.{Timer => _}
+
+import scala.language.higherKinds
+
+trait DockerSetup extends OsSetup {
```
alari: Setup means something prepared to set up and work within the setup. Here we have only an utility function.

PR #349, opened 2019-01-20
Refactoring of the VM part

michaelvoronov:
- a little changing in function invoking - now a client should call Wasm module function smth like this `[optional module name](<argument as hex string>)`;
- fix issue with possible absent allocation/deallocation functions;
- decouple WasmFunction to separate class;
- introduce WasmModule class that incapsulate all Wasm memory relative operations;
- fix bug with possible `Sum(null)` as a result of 'getMemory' invoking;
- move payload parsing to extractors;
- add some new tests to Statemachine and VM;


vm/src/main/scala/fluence/vm/wasm/WasmModuleMemory.scala
```diff
@@ -0,0 +1,99 @@
...
+
+final case class WasmModuleMemory(memory: ByteBuffer) {
+
+  /**
+   * Reads [offset, offset+size) region from the memory.
+   *
+   * @param offset offset from which read should be started
+   * @param size bytes count to read
+   */
+  def readBytes[F[_]: Applicative](
```
alari: Strictly speaking, `Applicative` is not enough, as invocation of the method actually performs IO on the memory buffer, doesn't it? So the function is impure.

statemachine/src/main/scala/fluence/statemachine/tx/PayloadParser.scala
```diff
@@ -0,0 +1,73 @@
...
+ */
+case class VmFunctionCallDescription(module: Option[String], arg: Array[Byte]) extends FunctionCallDescription
+
+/**
+ * Extractor for command that explicitly closes session.
+ */
+object SmCloseSession {
+  val CloseSession = "@closeSession"
+
+  def unapply(payload: String): Option[SmCloseSessionDescription] = payload match {
```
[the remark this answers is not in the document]
michaelvoronov: done

vm/src/main/scala/fluence/vm/wasm/WasmModule.scala
```diff
@@ -0,0 +1,235 @@
...
+ * Wrapper of Wasm Module instance compiled by Asmble to Java class. Provides all functionality of Wasm modules
+ * according to the Fluence protocol (invoke, parameter passing, hash computing). TODO: after removing alloc/
+ * dealloc should be refactored to two modules types (extends the same trait): "master" (that has invoke method
+ * and can routes call from user to "slaves") and "slave" (without invoke method that does only computation).
+ *
+ * @param name an optional module name (according to Wasm specification module name can be empty string (that is also
+ *             "valid UTF-8") or even absent)
+ * @param wasmMemory the memory of this module (please see comment in apply method to understand why it optional now)
+ * @param moduleInstance a instance of Wasm Module compiled by Asmble
+ * @param allocateFunction a function used for allocation memory region for parameter passing
```
folex: used for allocation memory region -> used for allocation of the memory region

used for deallocation memory region -> used for deallocation of the memory region

vm/src/main/scala/fluence/vm/wasm/WasmModule.scala
```diff
@@ -0,0 +1,235 @@
...
+
+  def getName: Option[String] = name
+
+  /**
+   * Allocates a memory region in Wasm module of supplied size by allocateFunction.
+   *
+   * @param size a size of memory that need to be allocated
+   */
+  def allocate[F[_]: LiftIO: Monad](size: Int): EitherT[F, InvokeError, Int] =
+    invokeWasmFunction(allocateFunction, size.asInstanceOf[AnyRef] :: Nil)
```
folex: Int.box

vm/src/main/scala/fluence/vm/wasm/WasmModule.scala
```diff
@@ -0,0 +1,235 @@
...
+
+  /**
+   * Deallocates a previously allocated memory region in Wasm module by deallocateFunction.
+   * This function is a temporary solution and should be removed in the nearest future.
+   *
+   * @param offset an address of the memory region to deallocate
+   * @param size a size of memory region to deallocate
+   */
+  def deallocate[F[_]: LiftIO: Monad](offset: Int, size: Int): EitherT[F, InvokeError, Unit] =
+    invokeWasmFunction(deallocateFunction, offset.asInstanceOf[AnyRef] :: size.asInstanceOf[AnyRef] :: Nil)
```
folex: Int.box

vm/src/main/scala/fluence/vm/wasm/WasmModule.scala
```diff
@@ -0,0 +1,235 @@
...
+   * Allocates a memory region in Wasm module of supplied size by allocateFunction.
+   *
+   * @param size a size of memory that need to be allocated
+   */
+  def allocate[F[_]: LiftIO: Monad](size: Int): EitherT[F, InvokeError, Int] =
+    invokeWasmFunction(allocateFunction, size.asInstanceOf[AnyRef] :: Nil)
+
+  /**
+   * Deallocates a previously allocated memory region in Wasm module by deallocateFunction.
+   * This function is a temporary solution and should be removed in the nearest future.
```
folex: Maybe a few words on why it's temporary and should be removed? If it can be explained in a few words.

vm/src/main/scala/fluence/vm/wasm/WasmModule.scala
```diff
@@ -0,0 +1,235 @@
...
+        )
+      }
+
+      // TODO: there are two ways of getting memory from a Wasm module: by exported getMemory function or
+      // through moduleInstance.getMem interface. Based on Asmble source code It seems that the behavior
+      // of getMemory is returning smth like 'export memory' in terms of Wasm specification. getMemory
+      // is more suitable for the scenario when at first Wasm module translates to Java class and then
+      // compiled with the rest of Java code. In our approach, it needs to use reflection for invocation.
+      // Another issue lies in the fact that Asmble may not generate this function. But ByteBuffer presence
+      // in every translated module (it is created in ctors of generated Java class). The second approach
```
folex: > But ByteBuffer presence in every translated module

`But ByteBuffer is present in every ...`?

vm/src/main/scala/fluence/vm/wasm/WasmModule.scala
```diff
@@ -0,0 +1,235 @@
...
+        val getMemoryMethod: Option[Method] = Try(
+          moduleInstance.getClass.getMethod("getMemory")
+        ).toOption
+        getMemoryMethod
+          .flatMap(Option(_))
+          .map(_.invoke(moduleInstance).asInstanceOf[ByteBuffer])
+          .flatMap(Option(_))
+      }.toEither.left.map { e ⇒
+        InitializationError(
+          s"Unable to getting memory from module=${moduleDescription.getName}",
```
folex: Unable to get

vm/src/main/scala/fluence/vm/wasm/WasmModule.scala
```diff
@@ -0,0 +1,235 @@
...
+          case (acc @ (_, _, None), m @ WasmFunction(`invokeFunctionName`, _)) ⇒
+            acc.copy(_3 = Some(m))
+
+          case (acc @ (_, _, _), _) ⇒ acc
+        }
+
+    } yield
+      new WasmModule(
+        Option(moduleDescription.getName),
+        memory.map(memory ⇒ WasmModuleMemory(memory)),
```
folex: You can do `memory.map(WasmModuleMemory(_))`

vm/src/test/scala/fluence/vm/WasmVmSpec.scala
```diff
@@ -52,40 +52,38 @@ class WasmVmSpec extends WordSpec with Matchers {
...
 
-    "there are two files" in {
+    "with two files with different module name" in {
       val sumFile = getClass.getResource("/wast/sum.wast").getPath
       val mulFile = getClass.getResource("/wast/mul.wast").getPath
 
       WasmVm[IO](NonEmptyList.of(mulFile, sumFile)).success()
     }
 
+    "2 module has function with equal name" in {
```
folex: `two modules has functions with same names` / `two modules with clashing function names`

vm/src/test/scala/fluence/vm/WasmVmSpec.scala
```diff
@@ -52,40 +52,38 @@ class WasmVmSpec extends WordSpec with Matchers {
...
+      // module without name and with some functions with the same name ("allocate", "deallocate", "invoke", ...)
+      val sum1File = getClass.getResource("/wast/no-getMemory.wast").getPath
+      // module without name and with some functions with the same name ("allocate", "deallocate", "invoke", ...)
+      val sum2File = getClass.getResource("/wast/bad-allocation-function-i64.wast").getPath
+
+      val res = for {
+        vm <- WasmVm[IO](NonEmptyList.of(sum1File, sum2File))
+      } yield vm
+
+      res.success()
```
folex: What does this test check? Will names be overwritten in a determined fashion? If so, maybe test against it? E.g., check that the second module's functions are available to call while the first one's aren't

vm/src/main/scala/fluence/vm/wasm/WasmModuleMemory.scala
michaelvoronov: Yes, seemds it does. Fixed.

vm/src/main/scala/fluence/vm/package.scala
```diff
@@ -44,4 +32,17 @@ package object vm {
     }
   }
 
+  /**
+   *  Runs action inside Try block, convert to EitherT with specified effect F.
+   */
+  def runThrowable[F[_]: Applicative, T, E <: VmError](
```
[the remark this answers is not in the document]
michaelvoronov: Hmm, seems reasonable, will place it to separate util file and rename to smth like `safeRunThrowable`.

vm/src/main/scala/fluence/vm/wasm/WasmModule.scala
michaelvoronov: Yes, but seems `a` memory region, since it isn't a concrete region.

vm/src/main/scala/fluence/vm/wasm/WasmModule.scala
michaelvoronov: It seems that it is better to remove this comment since allocation function also will be deleted. All of this is needed for verification game.

vm/src/test/scala/fluence/vm/WasmVmSpec.scala
michaelvoronov: Let it be `two modules have function with the same names`.

vm/src/test/scala/fluence/vm/WasmVmSpec.scala
michaelvoronov: Yes, you are right, but It is better to resolve it in the following PR.

PR #357, opened 2019-01-22
Fix deploy scripts

DieMyst:
* add cors origin policy to tendermint config
* update contract
* update parity container version
* a deploy fixes


tools/deploy/fabfile.py
```diff
@@ -30,21 +30,22 @@
 
 # Set the username
 env.user = "root"
- 
+
 def copy_resources():
 
+    run('rm -r scripts')
```
folex: Let's add comment on why we're doing this

tools/deploy/scripts/deploy-contract.js
```diff
@@ -47,6 +47,6 @@ request.post(
                     })
 
                 });
-            })
+            }).catch(console.log)
```
folex: Let's add some prefix to log, so it's easy to find it in source code. E.g. `e => console.log("error in deploy-contract.js: " + e)`

PR #358, opened 2019-01-22
Migrate from Box<Error> to failure

folex:


cli/src/contract_func.rs
```diff
@@ -55,8 +56,8 @@ impl ContractCaller {
         credentials: &Credentials,
         call_data: ethabi::Bytes,
         gas: u32,
-    ) -> Result<H256, Box<Error>> {
-        let (_eloop, transport) = Http::new(&self.eth_url)?;
+    ) -> Result<H256, Error> {
+        let (_eloop, transport) = Http::new(&self.eth_url.as_str()).map_err(SyncFailure::new)?;
```
DieMyst: there is `str` already, no need to use `as_str()`

DieMyst: what is `SyncFailure`?

cli/src/contract_func.rs
```diff
@@ -88,14 +89,15 @@ impl ContractCaller {
         secret: &Secret,
         call_data: ethabi::Bytes,
         gas: u32,
-    ) -> Result<H256, Box<Error>> {
-        let gas_price = web3.eth().gas_price().wait()?;
+    ) -> Result<H256, Error> {
+        let gas_price = web3.eth().gas_price().wait().map_err(SyncFailure::new)?;
```
DieMyst: is it possible to transform it implicitly?

cli/src/contract_status/mod.rs
```diff
@@ -79,17 +60,22 @@ mod tests {
         let contract_address: Address = CONTRACT_ADDR.parse().unwrap();
 
         let creds: Credentials = Credentials::No;
+        let account: Address = OWNER.parse().unwrap();
+
+        let eth = EthereumArgs {
+            credentials: creds,
+            gas: 1000000,
```
DieMyst: maybe use gas from arguments? As I remember, there is such argument.

cli/src/contract_func.rs
folex: as_str here to convert `String` to `&str`

folex: `SyncFailure` is a way to add `Sync` to an `std::Error + Send + 'static`. It's needed to create a `failure::Error` which requires `std::Error + Sync + Send + 'static`

cli/src/contract_func.rs
folex: It's possible to later use our own error type based on `Failure`, I hope it will handle conversion implicitly.

cli/src/contract_status/mod.rs
folex: There's no gas argument in tests :)

cli/src/command.rs
```diff
@@ -0,0 +1,148 @@
...
+        .help("public key of tendermint node")
+}
+
+pub fn base64_tendermint_key<'a, 'b>() -> Arg<'a, 'b> {
+    Arg::with_name(BASE64_TENDERMINT_KEY)
+        .long(BASE64_TENDERMINT_KEY)
+        .help("allows to use base64 tendermint key")
+}
+
+pub fn with_ethereum_args<'a, 'b>(args: &[Arg<'a, 'b>]) -> Vec<Arg<'a, 'b>> {
```
DieMyst: docs for this method

cli/src/delete_app.rs
```diff
@@ -14,154 +14,84 @@
...
-    gas: u32,
-    account: Address,
-    contract_address: Address,
-    eth_url: String,
     deployed: bool,
+    eth: command::EthereumArgs,
 }
 
 pub fn subcommand<'a, 'b>() -> App<'a, 'b> {
+    let my_args = &[
```
DieMyst: what is `my_args`?

folex: That's an array of arguments used by this command. It's used in `with_ethereum_args` below.

PR #359, opened 2019-01-22
Implement deleteNode in smart contract

folex:


bootstrap/contracts/Deployer.sol
```diff
@@ -301,13 +317,129 @@ contract Deployer {
...
+                }
+            }
+
+            // This should never happen if there's no bugs. But it's better to revert if there is some inconsistency.
+            require(j < nodeIDs.length, "error deleting node: nodeID wasn't found in nodeIDs");
+
+            // Check if that node is the last one hosting that app
+            if (nodeIDs.length == 1) {
+                // Remove app from clustersIds array and clusters mapping, and emit AppDeleted event of successful removal
+                removeApp(appID);
```
DieMyst: if the node is removed, we should find another one to work with the app. Am I right?
It is incorrect to delete an app when deleting a node.

bootstrap/contracts/Deployer.sol
```diff
@@ -301,13 +317,129 @@ contract Deployer {
...
+                // Remove nodeID from nodeIDs array
+                if (j != nodeIDs.length - 1) {
+                    // Remove j-th node from nodeIDs replacing it by the last node in the array
+                    nodeIDs[j] = nodeIDs[nodeIDs.length - 1];
+                }
+                // Release the storage
+                delete nodeIDs[nodeIDs.length - 1];
+
+                // Decrease array length manually
+                assembly { mstore(nodeIDs, sub(mload(nodeIDs), 1)) }
```
DieMyst: I see in code something like this: `readyNodes.length--;` Is it the same, is it?

bootstrap/contracts/Deployer.sol
```diff
@@ -301,13 +317,129 @@ contract Deployer {
...
+    function removeNodeFromAppsStorage(bytes32 nodeID, bytes32[] storage appIDsArray)
+        internal
+    {
+        for (uint i = 0; i < appIDsArray.length; i++) {
+            bytes32 appID = appIDsArray[i];
+            bytes32[] storage nodeIDs = apps[appID].cluster.nodeIDs;
+            uint nodeIDsLength = nodeIDs.length;
+
+            uint8 j = 0;
+            for (; j < nodeIDsLength; j++) {
```
DieMyst: this part is repeated, maybe move it to a method?

DieMyst: function `foundNodeIndex`

bootstrap/test/NodeDeleteTest.js
```diff
@@ -0,0 +1,102 @@
...
+            let receipt = await global.contract.deleteNode(nodeID, { from: anyone });
+            truffleAssert.eventEmitted(receipt, utils.nodeDeletedEvent, ev => {
+                return ev.id == nodeID;
+            });
+            return receipt;
+        }));
+
+        truffleAssert.eventEmitted(receipts.pop(), utils.appDeletedEvent);
+    });
+});
```
DieMyst: new line

bootstrap/contracts/Deployer.sol
folex: Great idea! Done that

PR #362, opened 2019-01-23
VM/optimization of statemachine payloads

michaelvoronov:


vm/src/main/scala/fluence/vm/AsmbleWasmVm.scala
```diff
@@ -85,7 +85,7 @@ class AsmbleWasmVm(
 
             concatHashes = Array.concat(moduleStateHash, prevModulesHash)
 
-            // TODO : There is known the 2nd preimage attack to such scheme with the same hash function
+            // TODO : It is known the 2nd preimage attack to such scheme with the same hash function
```
alari: Both variants are incorrect :)

PR #375, opened 2019-01-26
Wait for events after submitting a transaction

folex:
- Wait for tx inclusion and events in publish, register, delete_app and delete_node
- Report results of app publish & node registration back to user 
    - e.g., was app deployed or enqueued, etc
- Implement waiting for block inclusion using `futures-retry` librart
    - it's kinda questionable, since it's a bit too complex for such a simple task. But it showcases how to use tokio & futures-retry, so maybe it worth it

There's **SO. MUCH. BOILERPLATE.** I'm not sure of the right way to refactor it out. Let's discuss!


cli/src/delete_node.rs
```diff
@@ -61,26 +64,69 @@ impl DeleteNode {
...
+            )))?;
+
+            Ok(())
         };
 
         if show_progress {
-            utils::with_progress(
+            let sync_inc = self.eth.wait_eth_sync as u32;
+            let steps = 1 + (self.eth.wait_tx_include as u32) + sync_inc;
+            let step = |s| format!("{}/{}", s + sync_inc, steps);
```
DieMyst: logic is obscure here.

DieMyst: this logic can be generalized into class with `register step` method and `get next step number` method.

cli/tests/utils/mod.rs
```diff
@@ -82,6 +78,8 @@ impl TestOpts {
             credentials,
             eth_url,
             gas,
+            wait_tx_include: false,
```
DieMyst: what with tests with waiting for sync and transaction?

DieMyst:
- check that new logic is working correctly
- tests on the new logic

cli/src/delete_node.rs
folex: That would require to store closures in a struct, and that's tricky in Rust, AFAIK.

I would leave it like it is now, because it's really a local knowledge, until the whole thing with command steps is generalized into a separate struct with `register`/`next_step`, but I can move it to utils func
```rust
pub fn variable_step_prefix(prestep_active: bool, total_steps: u32, current_step: u32) -> String {
    let prestep_inc = prestep_active as u32;
    format!("{}/{}", current_step + prestep_inc, total_steps)
}
```
but it's HORRIBLE :(

folex: I really want leave it for now, and refactor later.

cli/src/contract_func.rs
```diff
@@ -14,165 +14,248 @@
...
 
-impl ContractCaller {
-    pub fn new(contract_address: Address, eth_url: &str) -> Result<ContractCaller, Error> {
-        let eth_url = eth_url.to_owned();
-        Ok(ContractCaller {
-            eth_url,
-            contract_address,
-        })
-    }
+pub fn get_transaction_logs_raw<T, F>(
```
DieMyst: docs

cli/src/contract_func.rs
```diff
@@ -14,165 +14,248 @@
...
+        .filter(|s| !s.is_zero())
+        .ok_or(err_msg(format!(
+            "Transaction {:#x} has status of 0, meaning it was reverted",
+            tx
+        )))?;
+
+    Ok(receipt.logs.into_iter().filter_map(parse_log).collect())
+}
+
+pub fn get_transaction_logs<T, F>(eth_url: &str, tx: &H256, parse_log: F) -> Result<Vec<T>, Error>
```
DieMyst: docs

cli/src/contract_func.rs
```diff
@@ -14,165 +14,248 @@
...
+where
+    F: Fn(RawLog) -> ethabi::Result<T>,
+{
+    get_transaction_logs_raw(eth_url, tx, |l| {
+        let raw = RawLog::from((l.topics, l.data.0));
+        parse_log(raw).ok()
+    })
+}
+
+pub fn wait_tx_included(eth_url: String, tx: &H256) -> Result<Web3Transaction, Error> {
```
DieMyst: docs

PR #382, opened 2019-01-30
Mining guide: pinning & node removal 

folex:


cli/README.md
```diff
@@ -157,3 +157,25 @@ The results will be in JSON and should resemble the following
 
 Here you can see two apps, the first one is enqueued (cluster is `null`) and waiting for enough nodes to host it, and the second one is already hosted on top of 5 nodes. Second app also specifies all 5 nodes in `pin_to_nodes`, and you can see the same nodes in `cluster.node_ids`.
 
+## Tips and tricks
+### Waiting for an Ethereum node to sync
+There is a flag `--wait_syncing` that, when supplied, will make CLI to wait until your Ethereum node is fully synced. It works by querying `eth_syncing` until it returns `false`. 
+
+This is handy when you don't want to manually check if Ethereum node is synced.
+
+### Waiting for a transaction to be included in a block
```
DieMyst: An output examples or pictures will be useful

folex: For every command? It's kinda different everywhere... Why? There's just a progress bar, nothing unexpected, if you used CLI at least once.

DieMyst: I am about results with app_id. It is for some expectations from user

PR #389, opened 2019-02-01
Pre- Remove peer: WorkersPool refactoring, Worker Control API

alari:


statemachine/src/main/scala/fluence/statemachine/control/ControlServer.scala
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+package fluence.statemachine.control
+import cats.effect.{Concurrent, Resource}
+
+import scala.language.higherKinds
+
+case class ControlServer[F[_]](signals: ControlSignals[F])
+
+object ControlServer {
```
folex: A few words on a future purpose of this class would be nice. Is this an rx end of `controlRPC` on the node?

node/src/main/scala/fluence/node/workers/DockerWorkersPool.scala
```diff
@@ -59,12 +57,27 @@ class DockerWorkersPool[F[_]: ContextShift: Timer](
...
+   * Runs a worker concurrently, registers it in `workers` map
+   *
+   * @param params Worker's description
+   * @return Unit; no failures are expected
+   */
   private def runWorker(params: WorkerParams): F[Unit] =
     for {
       deferred ← Deferred[F, Unit]
-      _ <- Concurrent[F].start(
+      fiberDef ← Deferred[F, Fiber[F, Unit]]
```
folex: It's an interesting pattern you have implemented here. But I think this pattern would look better if it's refactored to a more general function, with descriptive names.

folex: If not, then there should be a lot of comments on which deferred means what, and when each `complete` fires and for what purpose

node/src/main/scala/fluence/node/workers/Worker.scala
```diff
@@ -56,64 +59,99 @@ case class Worker[F[_]] private (
...
-                health
-              }
-
-            case (_, false) ⇒
-              logger.error(s"Healthcheck is failing for worker: $params")
-              Applicative[F].pure(WorkerContainerNotRunning(StoppedWorkerInfo(params.currentWorker)))
+  ): fs2.Stream[F, WorkerHealth] =
+    container
+      .checkPeriodically[F](healthCheckConfig.period)
+      .evalMap(st ⇒ Timer[F].clock.realTime(MILLISECONDS).map(now ⇒ (now - st.startedAt) → st.isRunning))
```
folex: I'd move that logic with Timer to a small inner function

node/src/main/scala/fluence/node/workers/Worker.scala
```diff
@@ -56,64 +59,99 @@ case class Worker[F[_]] private (
...
+   * @param healthCheck Stream of [[WorkerHealth]]'s, updates the given ref
+   * @return Resource that will stop healthchecks stram once released
+   */
+  private def runHealthCheck[F[_]: Concurrent: ContextShift: Timer](
+    healthReportRef: Ref[F, WorkerHealth],
+    healthCheck: fs2.Stream[F, WorkerHealth],
+  ): Resource[F, Unit] =
+    Resource
+      .make(
+        // We use Deferred to stop the stream when resource is released
```
folex: > We use Deferred to stop the stream when resource is released

->

> We create new Deferred instance and use it to signal to stop the stream when resource is released

node/src/main/scala/fluence/node/workers/Worker.scala
```diff
@@ -56,64 +59,99 @@ case class Worker[F[_]] private (
...
+        Deferred[F, Either[Throwable, Unit]].flatMap(
+          stop ⇒
+            // Run possibly infinite stream concurrently, get the fiber to join later
+            Concurrent[F]
+              .start(
+                healthCheck
+                  .evalTap(healthReportRef.set)
+                  .interruptWhen(stop)
+                  // TODO .sliding(healthcheck.slide) -- currently we have no behavior definition for failed healthcheck
```
folex: That's a crypty todo. I think it could be either deleted or be on top of the method, stating what exactly is missing, without references to a dead code.

node/src/main/scala/fluence/node/workers/Worker.scala
```diff
@@ -56,64 +59,99 @@ case class Worker[F[_]] private (
...
+   * @return Resource that will stop healthchecks stram once released
+   */
+  private def runHealthCheck[F[_]: Concurrent: ContextShift: Timer](
+    healthReportRef: Ref[F, WorkerHealth],
+    healthCheck: fs2.Stream[F, WorkerHealth],
+  ): Resource[F, Unit] =
+    Resource
+      .make(
+        // We use Deferred to stop the stream when resource is released
+        Deferred[F, Either[Throwable, Unit]].flatMap(
```
folex: Maybe it's possible to simplify here by using `stop` from Worker? Maybe not.

statemachine/src/main/scala/fluence/statemachine/config/StateMachineConfig.scala
```diff
@@ -25,4 +36,74 @@ package fluence.statemachine.config
...
+object StateMachineConfig {
+
+  /**
+   * Loads State machine config using `pureconfig` Scala config loading mechanism.
+   */
+  def load[F[_]: Sync](): F[StateMachineConfig] = {
+    import pureconfig.generic.auto._
+
+    Sync[F]
+      .defer(
```
DieMyst: why not use `delay`?
as from docs, it will hide an effect in `F`
```
/**
   * Lifts any by-name parameter into the `F` context.
   *
   * Equivalent to `Applicative.pure` for pure expressions,
   * the purpose of this function is to suspend side effects
   * in `F`.
   */
```

node/src/main/scala/fluence/node/workers/Worker.scala
alari: I'm not sure it could make things simpler. Here we encapsulate concurrent run of the stream, it's stopping when resource is released, _and_ waiting for the fiber to join, so it's kind of a concurrent-stream-specific logic that I'd better keep in one place.

folex: Ok

node/src/main/scala/fluence/node/workers/Worker.scala
```diff
@@ -58,100 +59,118 @@ case class Worker[F[_]] private (
...
       rpc ← TendermintRpc[F](params)
 
-    } yield new Worker[F](params, rpc, healthReportRef, stop.complete(Right(())) *> rpc.stop *> fiber.join *> onStop)
+      container ← DockerIO.run[F](params.dockerCommand)
+
+      healthChecks = healthCheckStream(container, params, healthCheckConfig, rpc)
+
+      _ ← runHealthCheck(healthReportRef, healthChecks)
+
+      control = ControlRpc[F]()
```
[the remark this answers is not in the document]
alari: It will be fixed by @folex in his PR.

node/src/main/scala/fluence/node/workers/tendermint/rpc/TendermintRpc.scala
```diff
@@ -83,6 +88,28 @@ object TendermintRpc {
   private def toSomePipe[F[_], T]: fs2.Pipe[F, T, Option[T]] =
     _.map(Some(_))
 
+  private def status[F[_]: Sync](
+    params: WorkerParams
+  )(implicit sttpBackend: SttpBackend[F, Nothing]): EitherT[F, Throwable, StatusResponse.WorkerTendermintInfo] =
+    EitherT {
+      val url = uri"http://${params.currentWorker.ip.getHostAddress}:${params.currentWorker.rpcPort}/status"
```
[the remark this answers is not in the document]
alari: It's not so trivial, as these URIs are different, they are not plain strings, and actually they are tendermint-related. I'd better keep them here.

node/src/main/scala/fluence/node/workers/DockerWorkersPool.scala
```diff
@@ -65,30 +65,45 @@ class DockerWorkersPool[F[_]: ContextShift: Timer](
...
               _ ← fiber.join
             } yield logger.info(s"Worker's Fiber joined: $params")
           )
           .use(
             worker ⇒
               for {
+                // Register worker in the pool
                 _ ← workers.update(_.updated(params.appId, worker))
-                _ ← deferred.get
+                // Wait for the deferred to be completed -- that's asynchronous blocking
```
folex: Comment should explain 'why' better than 'how'

PR #391, opened 2019-02-01
Filter status by owner, app_id, etc

folex:


cli/src/command.rs
```diff
@@ -98,18 +103,29 @@ pub fn tendermint_node_id<'a, 'b>() -> Arg<'a, 'b> {
         .help("Tendermint node ID (20-byte from SHA of p2p public key)")
 }
 
+pub fn node_ip<'a, 'b>() -> Arg<'a, 'b> {
+    Arg::with_name(NODE_IP)
+        .long(NODE_IP)
+        .value_name("ip address")
+        .short("i")
+        .required(true)
```
DieMyst: why it is required?

cli/src/contract_status/mod.rs
```diff
@@ -19,28 +19,204 @@ pub mod status;
...
+                .help("Filter nodes and apps by Tendermint validator key (node id)")
+                .display_order(3),
+            Arg::with_name(FILTER_MODE)
+                .long(FILTER_MODE)
+                .short(FILTER_MODE)
+                .required(false)
+                .takes_value(true)
+                .value_name("and|or")
+                .default_value("and")
+                .help("Logical mode of the filter")
```
DieMyst: More useful information will be good. It is not clear how to use these filters.

cli/src/publisher.rs
```diff
@@ -233,36 +234,39 @@ pub fn subcommand<'a, 'b>() -> App<'a, 'b> {
...
             .takes_value(true)
-            .help("Path to compiled `wasm` code"),
-        Arg::with_name(SWARM_URL)
-            .long(SWARM_URL)
-            .short("w")
-            .required(false)
-            .takes_value(true)
-            .help("Http address to swarm node")
-            .default_value("http://localhost:8500/"),
+            .help("Path to compiled `wasm` code")
```
DieMyst: how do you think, maybe it will be good to rewrite help information to make it more detailed.

cli/src/register.rs
```diff
@@ -235,32 +234,30 @@ pub fn parse(args: &ArgMatches) -> Result<Register, Error> {
...
+            .display_order(5),
         Arg::with_name(LAST_PORT)
             .alias(LAST_PORT)
             .default_value("20196")
             .long(LAST_PORT)
             .takes_value(true)
-            .help("Maximum port in the port range"),
-        base64_tendermint_key(),
+            .help("Maximum port in the port range")
+            .display_order(5),
```
DieMyst: start port and last port has the same display order.

folex: It's ok

cli/src/publisher.rs
folex: I guess we should review usefulness of all help messages in CLI.

cli/src/contract_status/mod.rs
folex: Will do

PR #396, opened 2019-02-06
Gas price as argument in CLI

DieMyst:


cli/src/command.rs
```diff
@@ -145,6 +149,12 @@ pub fn with_ethereum_args<'a, 'b>(args: &[Arg<'a, 'b>]) -> Vec<Arg<'a, 'b>> {
             .takes_value(true)
             .default_value("1000000")
             .help("Maximum gas to spend"),
+        Arg::with_name(GAS_PRICE)
+            .long(GAS_PRICE)
+            .short("G")
+            .required(false)
+            .takes_value(true)
+            .help("Gas price in Gwei"),
```
folex: add default value

PR #398, opened 2019-02-06
Сlean up worker containers on master stop + Tendermint in separate container

alari:
- Tendermint and Worker(statemachine) are in separate containers now
- Docker containers and networks are cleaned up properly
- Tendermint config generation refactored


node/src/main/scala/fluence/node/MasterNode.scala
```diff
@@ -40,60 +38,50 @@ import scala.language.higherKinds
...
       _ <- IO(logger.info(s"Worker started (newly=$newly) {}", params)).to[F]
     } yield ()
 
+  /**
+   * All app worker's data is stored here. Currently the folder is never purged
+   */
+  private def resolveAppPath(app: eth.state.App): F[Path] =
+    IO(rootPath.resolve("app-" + app.id + "-" + app.cluster.currentWorker.index)).to[F]
+
+  private def makeDataPath(app: eth.state.App): F[Path] =
```
DieMyst: `makeDataPath` and `makeVmCodePath` look like the same methods.

node/src/main/scala/fluence/node/eth/NodeEth.scala
```diff
@@ -70,7 +70,8 @@ object NodeEth extends LazyLogging {
     for {
       stateRef <- MakeResource.refOf[F, NodeEthState](initialState)
       blockQueue ← Resource.liftF(fs2.concurrent.Queue.circularBuffer[F, (Option[String], F[Block])](8))
-      _ ← MakeResource.concurrentStream(contract.ethClient.blockStream[F] to blockQueue.enqueue)
+      _ ← MakeResource
+        .concurrentStream(contract.ethClient.blockStream[F] to blockQueue.enqueue, name = "ethClient.blockStream")
```
DieMyst: it is only one place with postfix, why?

alari: What do you mean?

PR #399, opened 2019-02-06
Check if a node is already registered on registration in CLI

DieMyst:


cli/src/register.rs
```diff
@@ -159,51 +172,83 @@ impl Register {
...
-            let tx = utils::with_progress(
-                "Registering the node in the smart contract...",
-                step(1).as_str(),
-                "Transaction with node registration was sent.",
-                publish_to_contract_fn,
-            )?;
-
-            if self.eth.wait_tx_include {
-                utils::print_tx_hash(tx);
+            let is_registered = if !self.no_status_check {
```
michaelvoronov: It seems that `is_registered` is redundant here. You can write smth like this:
```
            if !self.no_status_check &&
                utils::with_progress(
                    "Check if node in smart contract is already registered...",
                    step_counter.format_next_step().as_str(),
                    "Smart contract checked.",
                    || check_node_registered_fn(),
                )? {
                return Ok(Registered::AlreadyRegistered)
            }

            let tx = utils::with_progress(
                "Registering the node in the smart contract...",
                step_counter.format_next_step().as_str(),
                "Transaction with node registration was sent.",
                publish_to_contract_fn,
            )?;
...
```

cli/src/register.rs
```diff
@@ -159,51 +172,83 @@ impl Register {
...
             } else {
-                Ok(Registered::TransactionSent(tx))
+                let tx = utils::with_progress(
+                    "Registering the node in the smart contract...",
+                    step_counter.format_next_step().as_str(),
+                    "Transaction with node registration was sent.",
+                    publish_to_contract_fn,
+                )?;
+
+                if self.eth.wait_tx_include {
```
michaelvoronov: It is usually a better style to inverting condition in such cases:
```
            if !self.eth.wait_tx_include {
                return Ok(Registered::TransactionSent(tx))
            }
            utils::print_tx_hash(tx);
            utils::with_progress(
                "Waiting for the transaction to be included in a block...",
                step_counter.format_next_step().as_str(),
                "Transaction was included.",
                || {
                    wait_tx_included(&tx, web3)?;
                    wait_event_fn(&tx)
                },
            )
```

It leads to decreasing bracket nesting level and improves code readability.

cli/src/register.rs
```diff
@@ -159,51 +172,83 @@ impl Register {
...
+                    Ok(Registered::TransactionSent(tx))
+                }
             }
         } else {
             if self.eth.wait_eth_sync {
                 wait_sync(web3)?;
             }
 
-            let tx = publish_to_contract_fn()?;
+            let is_registered = if !self.no_status_check {
```
michaelvoronov: the same with `is_registered`.

cli/src/register.rs
```diff
@@ -159,51 +172,83 @@ impl Register {
 
         // sending transaction with the hash of file with code to ethereum
         if show_progress {
-            let sync_inc = self.eth.wait_eth_sync as u32;
-            let steps = 1 + (self.eth.wait_tx_include as u32) + sync_inc;
-            let step = |s| format!("{}/{}", s + sync_inc, steps);
+            let mut step_counter = StepCounter::new(1);
```
michaelvoronov: What about incapsulating logic of StepCounter `register` inside `new` method? There could be either three methods:
`new(register: &Register)`
`new(publisher: &Publisher)`
`new(deleteApp: &DeleteApp)`
which would initialize StepCounter in a proper way (pattern `Builder`) or a new trait `StepCountable` (but it seems lire overkill for this purpose).

DieMyst: For now, the registration of steps and generation is in one place. So it is simple to manage both parts. I think, that if `StepCounter` creation will be written in separate methods it will be not in `StepCounter` but in `Register`, `Publisher`, etc. I will think about it.

PR #400, opened 2019-02-06
ControlRPC: dropPeer, status, stop.

folex:
There is a quirk in how we currently handle NodeDeleted event, so integration test for this will be implemented later.


statemachine/control/src/main/scala/fluence/statemachine/control/ControlSignals.scala
```diff
@@ -16,41 +16,49 @@
...
-   * @param change ChangePeer event
+   * Add a new DropPeer event
    */
-  def changePeer(change: ChangePeer): F[Unit] =
+  private[control] def dropPeer(drop: DropPeer): F[Unit] =
     for {
-      changes <- changePeersRef.take
-      _ <- changePeersRef.put(changes :+ change)
+      changes <- dropPeersRef.take
+      _ <- dropPeersRef.put(changes :+ drop)
```
alari: You're adding to the end of a linked list here. It's better either to add to the beginning, or to change the used data structure (I suppose `Queue` is a good fit).

statemachine/control/src/main/scala/fluence/statemachine/control/ControlSignals.scala
```diff
@@ -16,41 +16,49 @@
...
     } yield ()
 
-  // Retrieve list of current ChangePeer events
-  val changePeers: Resource[F, List[ChangePeer]] =
-    Resource.make(changePeersRef.tryTake.map(_.toList.flatten))(_ => changePeersRef.tryPut(Nil).void)
+  /**
+   * Retrieve list of current DropPeer events
```
alari: Add the docs why `Resource` is used here.

PR #401, opened 2019-02-06
Sdk/version 0.0.8

michaelvoronov:


sdk/rust/crates/macro/src/parser.rs
```diff
@@ -0,0 +1,166 @@
...
+                "Incorrect argument type - only Vec<u8> and String are supported",
+            )),
+        }?;
+
+        let type_segment = path
+            .segments
+            // argument can be given in full path form: ::std::string::String
+            // that why the last one used
+            .last()
+            .ok_or_else(|| Error::new(path.span(), "It has to be a valid input value"))?;
```
folex: `It` is confusing. I think error message should be `Type of an input argument cannot be empty`. Not sure if that error is ever possible, but error message will still server as a description for a code reader.

folex: Or `It has to have a non-empty input argument type`

sdk/rust/crates/macro/src/parser.rs
```diff
@@ -0,0 +1,166 @@
...
+            let arg_path = match arg_type {
+                syn::Type::Path(path) => Ok(&path.path),
+                _ => Err(Error::new(
+                    arg_type.span(),
+                    "Unsuitable type in Vec brackets - only Vec<u8> is supported>",
+                )),
+            }?;
+
+            // converts T to String
+            let arg_segment = arg_path.segments.first().ok_or_else(|| {
```
folex: Why `first()`, not `last()`? If we some day add our own type for `u8` or maybe allow `Vec<String>` or anything else, it can shoot us in the foot :)

sdk/rust/crates/macro/src/parser.rs
```diff
@@ -0,0 +1,166 @@
...
+}
+
+impl ReturnTypeGenerator for ParsedType {
+    fn generate_fn_epilog(&self) -> proc_macro2::TokenStream {
+        match self {
+            ParsedType::Utf8String => quote! {
+                memory::write_result_to_mem(
+                    result.as_bytes()
+                )
+                .expect("Putting result string to memory was failed.")
```
folex: has failed

sdk/rust/crates/macro/src/parser.rs
```diff
@@ -0,0 +1,166 @@
...
+            },
+            ParsedType::ByteVector => quote! {
+                memory::write_result_to_mem(&result[..])
+                    .expect("Putting result vector to memory was failed.")
```
folex: has failed

sdk/rust/src/lib.rs
```diff
@@ -15,19 +15,30 @@
...
-#![doc(html_root_url = "https://docs.rs/fluence/0.0.7")]
+#![doc(html_root_url = "https://docs.rs/fluence/0.0.8")]
 #![feature(allocator_api)]
 
-extern crate core;
+extern crate fluence_sdk_macro;
+extern crate fluence_sdk_main;
 
-pub mod memory;
+/// A module which should be typically imported:
```
folex: mb `A module which should be typically imported as a whole:`?

vm/examples/counter/Cargo.toml
```diff
@@ -8,19 +7,19 @@ description = "Simple counter for Fluence"
...
 panic = "abort"
 
 [dependencies]
-
-fluence = { version = "0.0.7", features = ["export_allocator"]}
+fluence = { version = "0.0.8", features = ["export_allocator"]}
 byteorder = { version = "1" }
+
+[replace]
+"fluence:0.0.8" = {path = "../../../sdk/rust"}
```
folex: I believe it can be removed by now

sdk/rust/src/lib.rs
michaelvoronov: I think that `A module which should be typically globally imported` is more suitable here

sdk/rust/crates/macro/src/parser.rs
michaelvoronov: Yes, i did it on purpose - there could be situation like this:
`Vec<some_crate::some_module::u8>` where `u8` can be any type.

But i think that additional check to segments count == 1 would be good here.

PR #404, opened 2019-02-07
Wrap block number in Option

folex:


ethclient/src/main/scala/fluence/ethclient/data/Block.scala
```diff
@@ -51,7 +53,7 @@ object Block {
     import scala.collection.convert.ImplicitConversionsToScala._
 
     new Block(
-      getNumber,
+      Option(getNumber).map(BigInt(_)).getOrElse(BigInt(0)), // null on Kovan with lightclient
```
michaelvoronov: mb `fold(BigInt(0))(BigInt(_))`?

folex: Yep, that would work too.

michaelvoronov: But `fold` is more safe (https://stackoverflow.com/a/50893488).

PR #417, opened 2019-02-12
add all arguments in block in option, handle errors in status aggregator

DieMyst:


ethclient/src/main/scala/fluence/ethclient/data/Block.scala
```diff
@@ -47,33 +47,35 @@ case class Block(
...
+      convertBigInteger(getDifficulty),
+      convertBigInteger(getTotalDifficulty),
+      Option(getExtraData).getOrElse(""),
+      convertBigInteger(getSize),
+      convertBigInteger(getGasLimit),
+      convertBigInteger(getGasUsed),
+      convertBigInteger(getTimestamp),
       getTransactions.toSeq.map(Transaction.apply),
-      getUncles,
+      getUncles.map(Option(_).getOrElse("")),
```
folex: `Option(getUncles).getOrElse(List.empty)`

node/src/main/scala/fluence/node/status/StatusAggregator.scala
```diff
@@ -87,16 +88,17 @@ object StatusAggregator extends LazyLogging {
...
-                  IO {
-                    logger.warn(s"Cannot produce MasterStatus response $err")
-                    err.printStackTrace(System.err)
-                  }
-                case _ ⇒ IO(logger.trace("MasterStatus responded successfully"))
-              }
+            val response = for {
+              status <- sm.getStatus
+              json <- IO(status.asJson.spaces2)
+                .onError({ case e => IO(logger.error(s"Status cannot be generated to JSON. Status: $status", e)) })
```
folex: This will not print error since slogging is broken. 
```
IO{ 
logger.error(s"Status cannot be generated to JSON. Status: $status, err: $e") 
e.printStackTrace(System.err)
}
```

ethclient/src/main/scala/fluence/ethclient/data/Block.scala
```diff
@@ -17,6 +17,7 @@
 package fluence.ethclient.data
 
 import java.math.BigInteger
+import scala.collection.JavaConverters._
```
alari: Why do you need this import? We have a converter already at line 56

PR #443, opened 2019-02-12
Sdk/version 0.0.10

### LINKED ISSUE #444: Add empty type to sdk ParsedType
It is needed to add Empty type for input arguments to sdk. It can be useful for examples such as counter.

michaelvoronov:
Closes #444


sdk/rust/crates/macro/src/lib.rs
```diff
@@ -80,57 +81,77 @@ fn invoke_handler_impl(fn_item: &syn::ItemFn) -> syn::Result<proc_macro2::TokenS
...
-            .unwrap()
-            .into_value(),
-    )?;
+    let input_type =
+        match decl.inputs.len() {
+            0 => ParsedType::Empty,
+            1 => ParsedType::from_fn_arg(decl.inputs.first().unwrap().into_value())?,
+            _ => return Err(Error::new(
+                decl.paren_token.span,
+                "The main module invocation handler has to don't have more then one input param",
```
folex: "The main module invocation handler has to don't have more than one input param",

sdk/rust/crates/macro/src/parser.rs
```diff
@@ -175,6 +174,7 @@ impl ReturnTypeGenerator for ParsedType {
                 memory::write_result_to_mem(&result[..])
                     .expect("Putting result vector to memory has failed")
             },
+            ParsedType::Empty => quote! {},
```
folex: Shouldn't you panic here since "The main module invocation handler has to have return type"?

michaelvoronov: It is undesirable. ParsedType shouldn't know much about context where it can be used and doesn't do job related only to outer context (Single Responsibility Principle).

PR #455, opened 2019-02-14
Generate Tendermint config.toml from default config

### LINKED ISSUE #427: Tendermint consumes 80-120% CPU while idle
This happens on Devnet nodes. Tendermint should expose a `pprof` endpoint, profile through it.

folex:
fixes #427

I used a library to parse TOML because Tendermint uses golang library viper for parsing TOML, and it doesn't seem to support inline sections, like this:
```toml
consensus.first = 1
consensus.second = 2
```
So I had to generate config with explicit tables (sections):
```
[consensus]
first = 1
second = 2
```
It's complicated to implement, so I used a library.


node/src/main/scala/fluence/node/workers/tendermint/config/ConfigTemplate.scala
```diff
@@ -15,24 +15,48 @@
...
+   * `rootPath` is usually /master inside Master container
+   * `templateConfigDir` contain:
+   *    - configs generated by `tendermint --init` (see [[Configuration.tendermintInit]])
+   *
+   * At the end of execution `dataPath` will contain:
+   *    - tendermint configuration in config:
+   *        - node_key.json, containing private P2P key
+   *        - priv_validator.json, containing validator's private & public keys and it's address
+   *        - genesis.json, generated from [[App.cluster]] and [[App.id]]
+   *        - config.toml, copied from `tpl.srcPath/config.toml` and updated
```
DieMyst: `tpl` is an unreadable argument. I think we should rename it.

node/src/main/scala/fluence/node/workers/tendermint/config/ConfigTemplate.scala
```diff
@@ -15,24 +15,48 @@
...
+   *
+   * `rootPath` is usually /master inside Master container
+   * `templateConfigDir` contain:
+   *    - configs generated by `tendermint --init` (see [[Configuration.tendermintInit]])
+   *
+   * At the end of execution `dataPath` will contain:
+   *    - tendermint configuration in config:
+   *        - node_key.json, containing private P2P key
+   *        - priv_validator.json, containing validator's private & public keys and it's address
+   *        - genesis.json, generated from [[App.cluster]] and [[App.id]]
```
DieMyst: some words about `genesis.json` and `config.toml` will be great, as about `node_key` and `priv_validator`.

node/src/main/scala/fluence/node/workers/tendermint/config/ConfigTemplate.scala
```diff
@@ -94,33 +116,6 @@ object ConfigTemplate {
...
-        configPath ← IO(dataPath.resolve("config"))
-        _ ← IO(Files.createDirectories(configPath))
-        _ ← tpl.copyMasterKeys(configPath)
-        _ ← tpl.updateConfigTOML(configPath, app, abciHost)
-        _ ← GenesisConfig(app).writeTo(configPath)
-      } yield ()
-    ).to[F]
-
+  def apply[F[_]: LiftIO](rootPath: Path, tendermintConfig: TendermintConfig): F[ConfigTemplate] =
+    IO(new ConfigTemplate(rootPath.resolve("tendermint").resolve("config"), tendermintConfig)).to[F]
```
DieMyst: why `LiftIO`? Maybe use `Sync` and `Sync.delay`?

node/src/main/scala/fluence/node/workers/tendermint/config/TendermintConfig.scala
```diff
@@ -0,0 +1,107 @@
...
+    "mempool.size" -> s"$mempoolSize",
+    "mempool.cache_size" -> s"$mempoolCacheSize",
+    "consensus.timeout_commit" -> s"${commitTimeoutMs}ms",
+    "consensus.skip_timeout_commit" -> s"$skipCommitTimeout",
+    "consensus.create_empty_blocks" -> s"$createEmptyBlocks",
+    "instrumentation.prometheus" -> s"$prometheus",
+  )
+
+  /**
+   * Takes Tendermint config in TOML at `src` and writes updated config to `dst`
```
DieMyst: missing new line between description and parameters.

and the description is not consistent with the method name. Maybe something like `updateSourceConfig`.

node/src/main/scala/fluence/node/workers/tendermint/config/ConfigTemplate.scala
folex: oh, it's just in doc. missed to remove it.

node/src/main/scala/fluence/node/workers/tendermint/config/ConfigTemplate.scala
folex: What do you mean? There are words about all of them in doc.

node/src/main/scala/fluence/node/workers/tendermint/config/ConfigTemplate.scala
folex: Because `rootPath.resolve` is IO, so it's wrapped in IO, and then converted to F via `LiftIO.to[F]`

node/src/main/scala/fluence/node/workers/tendermint/config/TendermintConfig.scala
folex: Why not consistent? Read, modify & write is pretty much `generate`. `generate` is more generic, though.

PR #460, opened 2019-02-15
Generate application.conf from env variables

### LINKED ISSUE #433: Remove application.conf generation from Node's entrypoint.sh
It's very obscure approach to build configuration file from docker's entrypoint bash script.

Turn to a simpler approach of making application.conf build itself from environment variables, in typesafe config style.

folex:
fixes #433


node/src/it/scala/fluence.node/MasterNodeIntegrationSpec.scala
```diff
@@ -221,8 +221,8 @@ class MasterNodeIntegrationSpec
       }.unsafeRunSync()
     }
 
-    "stop workers on AppDelete event" in {
-      deleteApp(26000).unsafeRunSync()
-    }
+//    "stop workers on AppDelete event" in {
```
DieMyst: is it needed?

node/src/main/resources/reference.conf
```diff
@@ -1,6 +1,6 @@
 // a path to a working directory, where all keys, configurations and codes files will be stored
 // it's usually set by entrypoint.sh
-root-path = "/master"
+root-path = "target/.tendermint"
```
DieMyst: is it possible to delete this for fail fast on deploy mistakes?
if it needed for tests, add this into test config.

node/src/main/scala/fluence/node/config/MasterConfig.scala
```diff
@@ -47,10 +49,18 @@ case class MasterConfig(
 )
 
 object MasterConfig {
+  import pureconfig.generic.auto._
+
   implicit val encodeMasterConfig: Encoder[MasterConfig] = deriveEncoder
   implicit val decodeMasterConfig: Decoder[MasterConfig] = deriveDecoder
 
-  import pureconfig.generic.auto._
+  implicit def reader[T: ConfigReader]: ConfigReader[Option[T]] = ConfigReader.fromCursor[Option[T]] { cv =>
```
DieMyst: add docs

node/src/main/resources/docker/application.conf
```diff
@@ -0,0 +1,22 @@
+root-path = "/master"
+master-container-id = ${?CONTAINER_ID}
+
+endpoints {
+  ip = ${TENDERMINT_IP} // always defined, guaranteed by entrypoint.sh
+  min-port = ${?MIN_PORT}
```
alari: Why should node know it?

folex: No reason for this. It should be read from smart contract, but currently `min-port` and `max-port` are only used in node's status, and there's no consumer for that information.

So I'll just delete it from endpoints config for now.

PR #481, opened 2019-02-20
Limit Docker resources: cpu, mem

folex:
Also add swarm.enabled flag in reference.conf


node/src/main/scala/fluence/node/docker/DockerImage.scala
```diff
@@ -19,7 +19,13 @@ package fluence.node.docker
...
 import io.circe.{Decoder, Encoder}
 
-case class DockerImage(image: String, tag: String) {
+/**
+ * Representation of a docker image
+ * @param image Fully qualified name of an image, including a repository and a name. E.g., fluencelabs/worker
+ * @param tag Tag of the image, will be appended to [[image]] after a colon
+ * @param limits Container cpu and memory limits
+ */
+case class DockerImage(image: String, tag: String, limits: DockerLimits) {
```
alari: It seems inconsistent that _image_ contains limits for _all its containers_. Perhaps we should decouple config case class and Docker domain model, and move limits e.g. to WorkerParams.

node/src/main/scala/fluence/node/docker/DockerImage.scala
```diff
@@ -19,7 +19,13 @@ package fluence.node.docker
 import io.circe.generic.semiauto.{deriveDecoder, deriveEncoder}
 import io.circe.{Decoder, Encoder}
 
-case class DockerImage(image: String, tag: String) {
+/**
+ * Representation of a docker image
```
alari: Blank line after the first line of docs?

node/src/main/scala/fluence/node/docker/DockerImage.scala
folex: I think it's totally OK. `DockerImage` is a description of how to run an image. Limits are semantically inseparable part of that description. Inseparable because there are no cases where images should be run with different limits. Where previously we said "tendermint/tendermint:latest", now we say "tendermint/tendermint:latest memory=M, cpus=C", because that what it means to be running tendermint, there's no other notion of tendermint in Fluence.

Also, I see absolutely no benefits in having two additional fields in `WorkerParams`, that will always go along `DockerImage`'s:
```scala
case class WorkerParams(
  app: App,
  tendermintPath: Path,
  vmCodePath: Path,
  masterNodeContainerId: Option[String],
  image: DockerImage,
+  imageLimits: DockerLimits,
  tmImage: DockerImage,
+  tmLimits: DockerLimits,
  configTemplate: ConfigTemplate
)
```
And additional boilerplate when running
```scala
def image(dockerImage: DockerImage, limits: DockerLimits): DockerParams.WithImage
... .image(tmImage, tmLimits).daemonRun("node")
```

I think it's much much better how is it now.

folex: Also, take a look at how nicely it composes in `reference.conf`:
```hocon
tendermint {
  image = "tendermint/tendermint"
  tag = "0.30.0"

  limits {
    cpus = 0.5
    memory-mb = 100
    memory-reservation-mb = 50
  }
}
```

For me, it makes total sense.

PR #483, opened 2019-02-21
Basic RocksDB KVStore implementation

alari:


effects/kvstore/src/main/scala/fluence/effects/kvstore/KVStore.scala
```diff
@@ -0,0 +1,157 @@
...
+package fluence.effects.kvstore
+
+import cats.Monad
+import cats.data.EitherT
+import fluence.codec.PureCodec
+
+import scala.language.higherKinds
+
+abstract class KVStore[F[_]: Monad, K, V] {
+  self ⇒
```
folex: self-typing without a type?

effects/kvstore/src/main/scala/fluence/effects/kvstore/RocksDBStore.scala
```diff
@@ -0,0 +1,158 @@
...
+        }
+      )
+
+      // Read options -- could be used for optimizations later, e.g. snapshots
+      readOptions ← Resource.make(IO(new ReadOptions()))(
+        rOpts ⇒
+          IO(rOpts.close()).handleError {
+            case NonFatal(err) ⇒
+              err.printStackTrace()
+              logger.error(s"Cannot close RocksDB object during cleanup: $err", err)
```
folex: stack trace should be after log

effects/kvstore/src/main/scala/fluence/effects/kvstore/RocksDBStore.scala
```diff
@@ -0,0 +1,158 @@
...
+        }
+      )
+
+    } yield
+      new KVStore[F, Array[Byte], Array[Byte]] {
```
folex: Why create anonymous class? Better move it to a named class

effects/kvstore/src/main/scala/fluence/effects/kvstore/KVStore.scala
alari: It's not typing, it's a self-reference, used to build anonymous inner classes with delegating the logic to `self`.

PR #488, opened 2019-02-22
Support capacity and api port in CLI, deployment scripts and docs

folex:
to be merged into #485


DieMyst:
update status schemes

PR #513, opened 2019-02-27
Fix quickstart

folex:


michaelvoronov:
In current version there are two similar sections about backend SDK. One in the `Backend guide` and one in `Tools and SDK`. I suggest to remove the second. In this case `CLI` can be moved to `Backend guide` as it really related to backend and `JS SDK` can be moved to `Frontend part` (which apparently was deleted ...).

What do you think?

folex:
> In current version there are two similar sections about backend SDK. One in the `Backend guide` and one in `Tools and SDK`. I suggest to remove the second. In this case `CLI` can be moved to `Backend guide` as it really related to backend and `JS SDK` can be moved to `Frontend part` (which apparently was deleted or lost ...).
> 
> What do you think?

I think it's totally OK to have two links to the same READMEs in different places. Users who won't read Backend Guide still might need to use CLI. It's useful for Miners and for Frontend developers. 

Same with other tools: I think it's better to have them listed separately, so someone who won't read our quickstart guides can learn about our tools right away.

folex:
I'll merge, let's discuss that in chat, and then change ToC if needed.

PR #516, opened 2019-02-28
Simplify statemachine

alari:
- Unbalanced Merkle Tree removed
- No more hex, base64, json


statemachine/src/main/scala/fluence/statemachine/state/AbciState.scala
```diff
@@ -0,0 +1,128 @@
...
+  def boundResponses[F[_]: Monad](limit: Int): StateT[F, AbciState, List[(Tx.Head, Array[Byte])]] =
+    StateT.get[F, AbciState].flatMap {
+      case st if st.responses.size > limit ⇒
+        // Can dequeue safely
+        val (drop, keep) = st.responses.dequeue
+        for {
+          // Drop one
+          _ ← StateT.set(st.copy(responses = keep))
+          // Drop more. TODO is it effective?
+          others ← boundResponses(limit)
```
michaelvoronov: It seems, that this function is a good candidate to be tailreced, doesn't it?

alari: A kind of monadic recursion is used there, it's not possible to use `@tailrec` directly. The question is, maybe it's possible just to take the head of the queue, with `limit` elements in it? Why to drop one by one?

Anyway, I'd consider this minor for now.

PR #520, opened 2019-03-02
Rust guessing game

folex:


vm/examples/guessing-game/src/lib.rs
```diff
@@ -0,0 +1,144 @@
...
+fn init() {
+    // Set seed to some random value, just to show how to use init()
+    SEED.with(|seed| *seed.borrow_mut() = 123456789);
+}
+
+#[invocation_handler(init_fn = init)]
+fn game(input: String) -> String {
+    count_request(&input);
+    update_seed(&input);
+    init_secret();
```
michaelvoronov: Hmm, seems like a good candidate to move to `init` function. And in this case `update_secret` could be use instead.

vm/examples/guessing-game/src/lib.rs
```diff
@@ -0,0 +1,144 @@
...
+                next_game();
+                format!("Success! You guessed right.")
+            }
+        },
+    }
+}
+
+/// Increment number of tries
+fn count_request(input: &String) {
+    TRIES.with(|t| *t.borrow_mut() += 1);
```
michaelvoronov: It is also possible to use `.add_assign(1)` here (`std::ops::AddAssign` should be included). But it is up to you.

folex: It would still look like
```rust
TRIES.with(|t| t.borrow_mut().add_assign(1))
```

so even more verbose

michaelvoronov: I don't think so. For me using `*` is less clear that `add_assign`. Btw it is very minor and both variant are good (e.g. in `ttt` i used `add_assign`).

PR #528, opened 2019-03-04
Cli/setup

DieMyst:


cli/src/config.rs
```diff
@@ -0,0 +1,115 @@
...
+    pub account: Option<Address>,
+    pub eth_url: String,
+    pub swarm_url: String,
+    pub secret_key: Option<H256>,
+    pub keystore_path: Option<String>,
+    pub password: Option<String>,
+}
+
+// returns default value if value is empty and vice versa
+pub fn empty_or_default(value: String, default: String) -> String {
```
michaelvoronov: Is it used anywhere?

cli/src/config.rs
```diff
@@ -0,0 +1,115 @@
...
+    if value.is_empty() {
```
michaelvoronov: You can use `as_some` from [boolinator](https://github.com/DanielKeep/rust-boolinator/blob/master/src/lib.rs) crate.

cli/src/config.rs
```diff
@@ -0,0 +1,115 @@
...
+        default
+    } else {
+        value
+    }
+}
+
+// returns None if string value is empty and Some if non empty
+pub fn none_if_empty(value: &str) -> Option<&str> {
```
michaelvoronov: Maybe it is better to use `as_some` instead this entire method?

cli/src/config.rs
```diff
@@ -0,0 +1,115 @@
...
+            secret_key,
+            keystore_path,
+            password,
+        };
+    }
+
+    // reads config file or generates default config if file does not exist
+    pub fn read_from_file_or_default(home_dir: &str) -> Result<SetupConfig, Error> {
+        let path = format!("{}{}", home_dir, "cli.json");
+        let path = Path::new(path.as_str());
```
michaelvoronov: Instead of `as_str` & can be used. In my opinion, it is more clear here, but up to you.

cli/src/config.rs
```diff
@@ -0,0 +1,115 @@
...
+
+        if !path.exists() {
```
michaelvoronov: Generally, it is better to trying to reduce number of `{`, `}`. From my point of view, In this case it is better to rewrite this section like this:

```
  if path.exist() {
    let content = read_to_string(path)?;
    let config: SetupConfig = serde_json::from_str(content.as_str())?;
    return Ok(config)
  }
  let contract: Address = DEFAULT_CONTRACT_ADDRESS
                .to_owned()
                .trim_start_matches("0x")
                .parse()?;
  let eth_url = DEFAULT_ETH_URL.to_owned();
  ...
  return OK(...)
```

cli/src/config.rs
```diff
@@ -0,0 +1,115 @@
...
+            let config: SetupConfig = serde_json::from_str(content.as_str())?;
+            Ok(config)
+        }
+    }
+
+    // writes config to file
+    pub fn write_to_file(&self, home_dir: &str) -> Result<(), Error> {
+        create_dir_all(HOME_DIR)?;
+        let config_str = serde_json::to_string(&self)?;
+        let path = format!("{}{}", home_dir, "cli.json");
```
michaelvoronov: Why "cli.json" is a separate parameter in format!?

cli/src/config.rs
DieMyst: I will need to implement a boolinator trait for strings. I don't think, that this is simpler than one method.

cli/src/config.rs
DieMyst: I don't like to return values in the middle of a method :)
So, I will move blocks under the if-else statement to separated methods.

cli/src/config.rs
michaelvoronov: Sorry, i should write this comment to `none_if_empty` function. It can be implemented like this:
`value.is_empty().as_some() without if`.

cli/src/config.rs
michaelvoronov: I mean `value.is_empty().as_some()` instead of this function. Now you write `none_if_empty(ethereum_address.as_str())` that consists of two additional call: `as_str()` and `none_if_empty`. The proposed variant also has two call but shorter :).

cli/src/config.rs
```diff
@@ -0,0 +1,110 @@
...
+            account,
+            eth_url,
+            swarm_url,
+            secret_key,
+            keystore_path,
+            password,
+        };
+    }
+
+    fn default() -> Result<SetupConfig, Error> {
```
michaelvoronov: I think it would be better to implement it like a Default trait implementation for SetupConfig.

cli/src/credentials.rs
```diff
@@ -40,3 +46,34 @@ impl Credentials {
...
+    keystore: Option<String>,
+    password: Option<String>,
+    secret_key: Option<Secret>,
+) -> Result<Credentials, Error> {
+    match keystore {
+        Some(keystore) => match password {
+            Some(password) => load_keystore(keystore, password).map(Credentials::Secret),
+            None => Err(err_msg("password is required for keystore")),
+        },
+        None => Ok(Credentials::get(secret_key, password.clone())),
```
michaelvoronov: Why you need to clone password here?

PR #564, opened 2019-03-15
Circleci

folex:
You can see CircleCI dashboard here: https://circleci.com/gh/fluencelabs/workflows
Here's what it looks like for me:
image

If it doesn't look the same for you, press 'Follow project' on the top right.


.circleci/config.yml
```diff
@@ -0,0 +1,184 @@
...
+
+  sbt integration tests:
+    <<: *jdk_image
+    <<: *jvm_env
+    steps:
+      - setup_remote_docker
+      - checkout
+      - <<: *rc
+      - <<: *rustup
+      - run: docker pull tendermint/tendermint:0.30.1
```
alari: 0.30.2 plz

PR #571, opened 2019-03-21
Add hardcoded public-key signature checking to llamadb

folex:


fluence-js/src/fluence.ts
```diff
@@ -46,7 +47,13 @@ export interface AppSession {
 /*
  * Creates connection with an app (all nodes related to an app in Fluence contract)
  */
-export async function connect(contract: string, appId: string, ethereumUrl?: string): Promise<AppSession> {
+export async function connect(contract: string, appId: string, ethereumUrl?: string, privateKey?: PrivateKey): Promise<AppSession> {
```
DieMyst: i think it is ok to create `connect` method without ethereumUrl. To avoid this
```
session.connect("contract", "appId", undefined, privateKey)
```

vm/src/it/resources/test-cases/llamadb/src/signature.rs
```diff
@@ -0,0 +1,101 @@
...
+
+fn get_pk() -> PublicKey {
+    PublicKey::parse_slice(&PUBLIC_KEY, None).expect("Invalid public key")
+}
+
+fn err_msg(s: &str) -> Box<Error> {
+    MyError(s.to_string()).into()
+}
+
+fn decode_hex(s: &str) -> GenResult<Vec<u8>> {
```
DieMyst: you can validate that string has the right length

vm/src/it/resources/test-cases/llamadb/src/signature.rs
```diff
@@ -0,0 +1,101 @@
...
+        .collect::<Result<Vec<u8>, ParseIntError>>()
+        .map_err(Into::into)
+}
+
+fn hash_message(message: &str) -> [u8; 32] {
+    let mut sha = Sha256::default();
+    sha.input(message.as_bytes());
+    let hash = sha.result();
+    let mut result = [0; 32];
+    result.copy_from_slice(hash.as_slice()); //TODO: is there a better way for GenericArray<u8, 32> -> [u8; 32]
```
DieMyst: why we cannot return `hash.as_slice()`?

vm/src/it/resources/test-cases/llamadb/src/signature.rs
```diff
@@ -0,0 +1,101 @@
...
+    Ok(verify(&message, &signature, &PK))
+}
+
+fn parse_signed(input: &String) -> GenResult<Signed> {
+    let pos: usize = input.find("\n").ok_or(err_msg(
+        &format!("Invalid input: no '\\n' between signature and nonce in `{}`. Should be <signature hex>\\n<nonce>\\n<sql_query>", input),
+    ))?;
+    let signature: &str = &input[..pos];
+    let nonce_payload: &str = &input[pos + 1..];
+    Ok(Signed {
```
DieMyst: why not split nonce and payload here too?

vm/src/it/resources/test-cases/llamadb/src/signature.rs
```diff
@@ -0,0 +1,101 @@
...
+}
+
+fn err_msg(s: &str) -> Box<Error> {
+    MyError(s.to_string()).into()
+}
+
+fn decode_hex(s: &str) -> GenResult<Vec<u8>> {
+    (0..s.len())
+        .step_by(2)
+        .map(|i| u8::from_str_radix(&s[i..i + 2], 16))
```
DieMyst: what if an error will be thrown in `from_str_radix`? It won't be collected in `collect`, am I right?

vm/src/it/scala/fluence/vm/LlamadbIntegrationTest.scala
```diff
@@ -69,11 +70,16 @@ class LlamadbIntegrationTest extends AppIntegrationTest with EitherValues {
       }
 
       "be able to create table and insert to it" in {
+        val vmT = WasmVm[IO](NonEmptyList.one(llamadbFilePath), "fluence.vm.client.4Mb")
+        println(s"vmT: ${vmT.value}")
```
DieMyst: is this `println`s needed?

vm/src/it/scala/fluence/vm/LlamadbIntegrationTest.scala
```diff
@@ -36,7 +36,8 @@ class LlamadbIntegrationTest extends AppIntegrationTest with EitherValues {
 
   private def createTestTable(vm: WasmVm): EitherT[IO, VmError, Array[Byte]] =
     for {
-      _ ← executeSql(vm, "CREATE TABLE Users(id INT, name TEXT, age INT)")
+      res ← executeSql(vm, "c631c67389432a7fc780c8c1a21db5eacaf95458ab1dfa7b05831add724851096b4da38731cc0d318154dea94a3016cac05744705df582d0713127af43897def\nCREATE TABLE Users(id INT, name TEXT, age INT)")
```
DieMyst: maybe use existing methods to sign this?

fluence-js/src/fluence.ts
folex: Seems like that's the way of TypeScript. I'll look into other options, but there's little hope

vm/src/it/resources/test-cases/llamadb/src/signature.rs
folex: Because here I'm returning not just `&[u8]`, but `&[u8; 32]`, so it's more strictly typed, and thus avoiding error handling in `Message::parse_slice` later. It could also be done via "safe" unsafe type casting.

Also, I hope, but not sure, that compiler will optimize this to be a no-op.

vm/src/it/resources/test-cases/llamadb/src/signature.rs
folex: Because I later need to calculate hash from concatenation of `nonce \n payload`, and if I split here, I would need to allocate new string or several bytes array. I tried that at first, but it was a bit clumsy.

vm/src/it/resources/test-cases/llamadb/src/signature.rs
folex: It would be collected. Look at turbo-fish 🐠 operator: `.collect::<Result<Vec<u8>, ParseIntError>>()`. collect does its magick, and transposes `Vec<Result>` to `Result<Vec>`

PR #588, opened 2019-03-28
[WIP] Ipfs short

folex:


folex:
There's no use for that anymore, see branch devnet/ipfs_hack

PR #601, opened 2019-04-03
Cli ipfs directories

DieMyst:


cli/src/storage.rs
```diff
@@ -87,7 +87,8 @@ pub fn upload_to_storage(
     Ok(hash)
 }
 
-/// Uploads bytes of code to the Swarm
+/// Uploads file of code to the Swarm
+/// TODO add directories support
 fn upload_code_to_swarm(url: &str, path: PathBuf) -> Result<String, Error> {
```
folex: Let's check if path is a directory, and throw an error?

folex: `upload_file_to_swarm`

cli/src/storage.rs
```diff
@@ -108,14 +109,25 @@ fn upload_code_to_swarm(url: &str, path: PathBuf) -> Result<String, Error> {
     Ok(res)
 }
 
-/// Uploads bytes of code to IPFS
+/// Uploads files to IPFS
 fn upload_code_to_ipfs(url: &str, path: PathBuf) -> Result<H256, Error> {
```
folex: `upload_file_to_ipfs`?

cli/src/storage.rs
```diff
@@ -125,9 +137,31 @@ fn upload_code_to_ipfs(url: &str, path: PathBuf) -> Result<H256, Error> {
...
+        // directory hash is in response with empty name
+        let dir_response: Option<&str> = responses
+            .iter()
+            .find(|r| r.name.is_empty())
+            .map(|r| r.hash.as_str());
+        dir_response.ok_or_else(|| err_msg("Multiple files uploaded, but no hash of directory."))?
+    };
 
-    let base58_str = ipfs_response.hash.as_str();
+    utils::print_info_msg("IPFS file address", base58_str.to_owned());
```
folex: file address? Could be a directory, better remove "file".

And what does that message say to user? Maybe print "Code was uploaded to IPFS: {}"?

PR #614, opened 2019-04-08
DockerWorkersPool.runWorker refactored

alari:
This PR should help not only with the code maintenance but also with some rare bugs.

### DIFF
--- a/node/src/main/scala/fluence/node/workers/DockerWorkersPool.scala
+++ b/node/src/main/scala/fluence/node/workers/DockerWorkersPool.scala
@@ -66,6 +66,11 @@ class DockerWorkersPool[F[_]: DockerIO: Timer, G[_]](
     } yield (healthy, oldWorker)
   }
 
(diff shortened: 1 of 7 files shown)


node/src/main/scala/fluence/node/workers/DockerWorkersPool.scala
```diff
@@ -65,64 +67,66 @@ class DockerWorkersPool[F[_]: DockerIO: Timer, G[_]](
...
+   * @param params Prepare WorkerParams; could be an expensive operation
+   * @param p2pPort P2p port
+   * @param stopTimeout Docker stop timeout
+   * @return Worker resource to be used
    */
-  private def workerServices(
-    params: WorkerParams,
+  private def workerResource(
+    onStop: F[Unit],
+    params: F[WorkerParams],
```
folex: Better name would be `init`, `create`, something like this

PR #657, opened 2019-04-24
Use upload snippet for Redis

folex:
Instead of llamadb's
image


alari:
Should we prepare some example requests for Redis?

folex:
> Should we prepare some example requests for Redis?

Definitely. It's just a hot-fix, so the snippet is at very least correct. I hope by the end of the day, I'll add a custom Redis snippet.

PR #659, opened 2019-04-24
Docs/cpp backend guide

michaelvoronov:


docs/src/backend/c_cpp/best_practices.md
```diff
@@ -0,0 +1,36 @@
...
+
+Generally, it is good to investigate all tracking compiler and linker warnings and try to get rid of them. But especially it is important during compilation to Webassembly. One of these seemingly innocuous warnings could look like this:
+
+```bash
+wasm-ld: warning: function signature mismatch: qsort
+>>> defined as (i32, i32, i32, i32) -> i32 in geo.o
+>>> defined as (i32, i32, i32, i32) -> void in /opt/wasi-sdk/share/sysroot/lib/wasm32-wasi/libc.a(qsort.o)
+```
+
+It could happen because of lack of `stdlib.h` in includes and manages compiler to generate an unnecessary `drop` instruction since the compiler will use the different signature of this function. And finally, such binary is invalid and couldn't run on Fluence.
```
alari: Could you restate this paragraph, please?

PR #664, opened 2019-04-25
Compile rust in frun

folex:


vm/frun/src/main/scala/Rust.scala
```diff
@@ -0,0 +1,107 @@
...
+import cats.data.NonEmptyList
+import cats.effect.IO
+
+import scala.sys.process.ProcessLogger
+import scala.util.control.NoStackTrace
+
+object NoProjectFound extends NoStackTrace {
+  override def getMessage: String =
+    s"$RustDirectory/Cargo.toml not found\n" +
+      s"Please pass Rust project as a volume: `-v /path/to/your/project:$RustDirectory`"
```
DieMyst: maybe it is better to use one directory and check if there is wasm files or rust project?

folex: It's always /rust for rust. You also specify a command `rust`, so there's no interference with wasm.

PR #673, opened 2019-05-06
Kademlia core

alari:
Bringing core Kademlia logic into the Fluence node.


kademlia/src/main/scala/fluence/kad/routing/LocalRoutingImpl.scala
```diff
@@ -0,0 +1,131 @@
...
+        )
+        .map(
+          idx ⇒
+            // Take contacts from the bucket, and sort them
+            buckets(idx).map(_.stream)
+        )
+    }
+
+    // Fetch the minimal necessary amount of nodes close to target from buckets
+    def fetchEnoughFromBuckets(
```
DieMyst: docs, the meaning of arguments is very unclear

kademlia/src/main/scala/fluence/kad/routing/LocalRoutingImpl.scala
```diff
@@ -0,0 +1,131 @@
...
+          collected.pure[F]
+        case next #:: evenMore ⇒
+          next.map(_.filter(predicate)).flatMap { fetched ⇒
+            fetchEnoughFromBuckets(evenMore, collected enqueue fetched)
+          }
+        case Stream.Empty ⇒
+          collected.pure[F]
+      }
+
+    def combine(left: Stream[Node[C]], right: Stream[Node[C]], seen: Set[Key] = Set.empty): Stream[Node[C]] =
```
DieMyst: docs, what does it mean `combine`?

kademlia/src/main/scala/fluence/kad/routing/LocalRoutingImpl.scala
```diff
@@ -0,0 +1,131 @@
...
+        case (hl #:: tl, _) if seen(hl.key) ⇒ combine(tl, right, seen)
+        case (_, hr #:: tr) if seen(hr.key) ⇒ combine(left, tr, seen)
+        case (hl #:: tl, hr #:: _) if ordering.lt(hl, hr) ⇒ hl #:: combine(tl, right, seen + hl.key)
+        case (hl #:: _, hr #:: tr) if ordering.gt(hl, hr) ⇒ hr #:: combine(left, tr, seen + hr.key)
+        case (hl #:: tl, hr #:: tr) if ordering.equiv(hl, hr) ⇒ hr #:: combine(tl, tr, seen + hr.key)
+        case (Stream.Empty, _) ⇒ right
+        case (_, Stream.Empty) ⇒ left
+      }
+
+    P sequential P.apply.map2(
```
DieMyst: it is better with parentheses I think

kademlia/src/main/scala/fluence/kad/routing/RefreshingIterativeRouting.scala
```diff
@@ -0,0 +1,156 @@
...
+      case Some(fiber) ⇒
+        // Do not do what was scheduled, schedule refresh instead, save schedule fiber
+        (fiber.cancel >> scheduleRefresh(idx)).map(mFiber.put)
+      case None ⇒
+        // It's in progress of re-scheduling, do nothing
+        ().pure[F]
+    }
+  }
+
+  override def lookupIterative(key: Key, neighbors: Int, parallelism: Int): F[Seq[Node[C]]] =
```
DieMyst: I don't know, but maybe it is better to copy docs from parent classes?

kademlia/src/main/scala/fluence/kad/routing/RoutingTable.scala
```diff
@@ -0,0 +1,108 @@
...
+  )
+
+  def bootstrapWithStore[F[_]: Concurrent: Clock, C: ContactAccess](store: KVStore[F, Key, Node[C]]): Extension[F, C] =
+    new Extension[F, C](
+      state ⇒ RoutingState.bootstrapWithStore(state, store),
+      _.pure[F],
+      _.pure[F]
+    )
+
+  def refreshing[F[_]: Concurrent: Timer, C](
```
DieMyst: docs

kademlia/src/main/scala/fluence/kad/routing/RoutingTable.scala
```diff
@@ -0,0 +1,108 @@
...
+
+object RoutingTable {
+
+  sealed class Extension[F[_], C](
+    val mapState: RoutingState[F, C] ⇒ F[RoutingState[F, C]],
+    val mapLocal: LocalRouting[F, C] ⇒ F[LocalRouting[F, C]],
+    val mapIterative: IterativeRouting[F, C] ⇒ F[IterativeRouting[F, C]]
+  )
+
+  def bootstrapWithStore[F[_]: Concurrent: Clock, C: ContactAccess](store: KVStore[F, Key, Node[C]]): Extension[F, C] =
```
DieMyst: docs

kademlia/src/main/scala/fluence/kad/routing/RoutingTable.scala
```diff
@@ -0,0 +1,108 @@
...
+) {
+
+  def nodeId: Key = local.nodeId
+
+}
+
+object RoutingTable {
+
+  sealed class Extension[F[_], C](
+    val mapState: RoutingState[F, C] ⇒ F[RoutingState[F, C]],
```
DieMyst: docs

kademlia/src/main/scala/fluence/kad/routing/RoutingTable.scala
```diff
@@ -0,0 +1,108 @@
...
+  private[kad] val state: RoutingState[F, C]
+) {
+
+  def nodeId: Key = local.nodeId
+
+}
+
+object RoutingTable {
+
+  sealed class Extension[F[_], C](
```
DieMyst: Maybe better name for this class?

kademlia/src/main/scala/fluence/kad/routing/RefreshingIterativeRouting.scala
alari: It's very verbose, and it makes it hard to maintain the docs.

kademlia/src/main/scala/fluence/kad/routing/RoutingTable.scala
alari: Do you have any better idea? Currently, it's `RoutingTable.Extension`, which sounds precise enough for me. So when you build a new `RoutingTable` with the `apply` method, you may choose some extensions for it.

DieMyst: Is it incorrect if these methods will be in separate classes (RoutingState, LocalRouting, IterativeRouting)?
I expect monad-like things from `map***` methods.

alari: `RoutingTable` is a single entry point with `apply` acting as a factory method. Extensions take parts of the `RoutingTable` during the process of wiring them one to another, and apply `map***` function on them. So the structure of `Extension` class contains knowledge about how exactly `RoutingTable` is going to be built.

I agree that `map***` is the wrong name for these methods, will rename.

kademlia/src/main/scala/fluence/kad/routing/LocalRoutingImpl.scala
```diff
@@ -0,0 +1,131 @@
...
+      Stream(idx)
+        .filter(_ < Key.BitLength) // In case current node is given, this will remove IndexOutOfBoundsException
+        .append(
+          Stream.from(1).takeWhile(i ⇒ idx + i < Key.BitLength || idx - i >= 0).flatMap { i ⇒
+            (if (idx - i >= 0) Stream(idx - i) else Stream.empty) append
+              (if (idx + i < Key.BitLength) Stream(idx + i) else Stream.empty)
+          }
+        )
+        .map(
+          idx ⇒
```
folex: maybe `idx` -> `bucketId`?

kademlia/src/main/scala/fluence/kad/routing/LocalRoutingImpl.scala
```diff
@@ -0,0 +1,131 @@
...
+   * @return
+   */
+  override def lookup(key: Key, numOfNodes: Int, predicate: Node[C] ⇒ Boolean = _ ⇒ true): F[Seq[Node[C]]] = {
+
+    implicit val ordering: Ordering[Node[C]] = Node.relativeOrdering(key)
+
+    // Build stream of neighbors, taken from buckets
+    val bucketsStream: Stream[F[Stream[Node[C]]]] = {
+      // Base index: nodes as far from this one as the target key is
+      val idx = (nodeId |+| key).zerosPrefixLen
```
folex: I think `(nodeId |+| key).zerosPrefixLen` could be moved to a separate function `bucketId` or `nodeBucket` since it's widely used.

Also `idx` isn't some new entity, and I think new terminology like 'base index` could be avoided. I see that `idx` as `bucketId`.

kademlia/src/main/scala/fluence/kad/routing/LocalRoutingImpl.scala
folex: Maybe call it `def mergeSort`?

kademlia/src/main/scala/fluence/kad/routing/LocalRoutingImpl.scala
```diff
@@ -0,0 +1,131 @@
...
+        case (hl #:: tl, hr #:: _) if ordering.lt(hl, hr) ⇒ hl #:: combine(tl, right, seen + hl.key)
+        case (hl #:: _, hr #:: tr) if ordering.gt(hl, hr) ⇒ hr #:: combine(left, tr, seen + hr.key)
+        case (hl #:: tl, hr #:: tr) if ordering.equiv(hl, hr) ⇒ hr #:: combine(tl, tr, seen + hr.key)
+        case (Stream.Empty, _) ⇒ right
+        case (_, Stream.Empty) ⇒ left
+      }
+
+    P sequential P.apply.map2(
+      // Stream of neighbors, taken from siblings
+      P parallel siblings.map(_.nodes.filter(predicate).toStream.sorted),
```
folex: Can't see reason for infix notation here and on the next line

kademlia/src/main/scala/fluence/kad/routing/LocalRoutingImpl.scala
```diff
@@ -0,0 +1,131 @@
...
+      // Stream of buckets, sorted by closeness
+      P parallel fetchEnoughFromBuckets(bucketsStream).map(_.toStream.sorted)
+    )(
+      // Combine stream, taking closer nodes first
```
folex: > taking closer nodes first

"Sorting by distance"

kademlia/src/main/scala/fluence/kad/routing/LocalRoutingImpl.scala
```diff
@@ -0,0 +1,131 @@
...
+      combine(_, _).take(numOfNodes)
+    )
+  }
+
+  /**
+   * Perform a lookup in local RoutingTable for a key,
```
folex: >    * Perform a lookup in local RoutingTable for a key,

"... skipping nodes that closer to `moveAwayFrom` than to `key`"

kademlia/src/main/scala/fluence/kad/routing/RefreshingIterativeRouting.scala
```diff
@@ -0,0 +1,156 @@
...
+) extends IterativeRouting[F, C] {
+
+  override def nodeId: Key = routing.nodeId
+
+  /**
+   * Called each time when any *Iterative method is called by user, reschedules the refreshing job for the bucket
+   *
+   * @param key Touched key
+   */
+  private def touchedIterative(key: Key): F[Unit] = {
```
folex: Maybe name it `rescheduleRefresh`? If not, let's at least avoid using past tense verbs in method names.

kademlia/src/main/scala/fluence/kad/routing/RefreshingIterativeRouting.scala
```diff
@@ -0,0 +1,156 @@
...
+        .traverse(Stream.range[Int](0, Key.BitLength - 1))(
+          idx ⇒ MVar.empty[F, Fiber[F, Unit]].map(idx -> _)
+        )
+        .map(_.toMap)
+    ).tupled.flatMap {
+      case (rnd, refreshFibers) ⇒
+        // Generate new timeout, sleep
+        val sleep = Sync[F].delay(refreshTimeout + rnd.nextInt.abs.millis) >>= Timer[F].sleep
+
+        // Lookup iteratively the bucket
```
folex: >         // Lookup iteratively the bucket

Incorrect word ordering. Should be "Lookup the bucket iteratively"

kademlia/src/main/scala/fluence/kad/routing/RefreshingIterativeRouting.scala
```diff
@@ -0,0 +1,156 @@
...
+ *
+ * @param routing Actual routing is delegated there
+ * @param refreshFibers State of refresh fibers // TODO is it optimal? won't it cause deadlocks? maybe map of mvars is better then mvar of map?
+ * @tparam F Effect
+ * @tparam C Contact
+ */
+private class RefreshingIterativeRouting[F[_]: Concurrent, C](
+  routing: IterativeRouting[F, C],
+  scheduleRefresh: Int ⇒ F[Fiber[F, Unit]],
+  refreshFibers: Map[Int, MVar[F, Fiber[F, Unit]]]
```
folex: `refreshFibers` sounds like a functon. Maybe `refreshingFibers`?

kademlia/src/main/scala/fluence/kad/protocol/Key.scala
```diff
@@ -0,0 +1,159 @@
...
+  override def toString: String = b64
+
+  /**
+   * Provides a Key with the same prefix of ''keepNumBits'' size and random suffix.
+   *
+   * @param keepNumBits Size of this key's prefix to keep
+   * @param rnd Random instance
+   * @return Randomized key
+   */
+  def randomize(keepNumBits: Int, rnd: Random = Random): Key =
```
folex: If I understand correctly, what this method does is generates a random key "at" the specified distance from `this`.

Maybe call it `def distantKey(distance: Int, rnd: ...): Key`? I was a bit confused by the naming, as it sound like it should be in the companion object instead.

folex: Or maybe rename `keepNumBits` to a `bucketId`? Not sure about that, though. 

Or maybe it's good as it is...

kademlia/src/main/scala/fluence/kad/routing/LocalRoutingImpl.scala
```diff
@@ -0,0 +1,131 @@
...
+  /**
+   * Tries to route a key to a Contact, if it's known locally
+   *
+   * @param key Key to lookup
+   */
+  override def find(key: Key): F[Option[Node[C]]] =
+    if (key === nodeId) Option.empty[Node[C]].pure[F]
+    else
+      P sequential P.apply.map2( // TODO: it's enough to find one non-empty reply; is there any way to explain it?
+        P parallel siblings.map(_.find(key)),
```
folex: There should be some parallelizm cap `alpha`, right? Is it respected here?

alari: Here we have just 2 jobs running in parallel, and both make no network calls. So there should be no problems.

kademlia/src/main/scala/fluence/kad/routing/IterativeRoutingImpl.scala
```diff
@@ -0,0 +1,406 @@
...
+  ): EitherT[F, JoinError, Unit] =
+    EitherT(
+      Parallel
+        .parTraverse(peers.toList) { peer: C ⇒
+          logger.trace("Going to ping Peer to join: " + peer)
+
+          // For each peer
+          // Try to ping the peer; if no pings are performed, join is failed
+          ContactAccess[C].rpc(peer).ping().attempt.to[F].flatMap[Option[(Node[C], List[Node[C]])]] {
+            case Right(peerNode) if peerNode.key =!= localRouting.nodeId ⇒ // Ping successful, lookup node's neighbors
```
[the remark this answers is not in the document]
alari: We have just contacts of peers, not `Key`s. To filter them in advance, we either need to serialize and store them with keys (as `Node[C]`) or assume that the contact of this node hasn't changed, hence we can filter it away. I suppose, making this check here hurts less than any fix of it.

kademlia/src/main/scala/fluence/kad/state/RoutingStateImpl.scala
```diff
@@ -0,0 +1,178 @@
...
+
+    // Update each portion in parallel, and return all updated nodes
+    def updateParPortions(portions: List[List[Node[C]]]): F[ModResult[C]] =
+      Parallel.parTraverse(portions)(updatePortion(_)).map(_.foldLeft(ModResult.noop[C])(_ <+> _))
+
+    updateParPortions(
+      // Rearrange in portions with distinct bucket ids, so that it's possible to update it in parallel
+      rearrange(
+        // Group by bucketId, so that each group should never be updated in parallel
+        nodes.groupBy(p ⇒ (p.key |+| nodeId).zerosPrefixLen).values
```
folex: `nodes.groupBy(n => nodeId.bucketId(n.key)).values` would be more readable. Note `p` -> `n`.

Also it's kinda awkward to xor **key** with **nodeId**. Maybe call `nodeId` -> `nodeKey`?

kademlia/src/main/scala/fluence/kad/state/RoutingStateImpl.scala
```diff
@@ -0,0 +1,178 @@
...
+   */
+  override def updateList(
+    nodes: List[Node[C]]
+  )(implicit clock: Clock[F], liftIO: LiftIO[F], ca: ContactAccess[C]): F[ModResult[C]] = {
+    // From iterable of groups, make list of list of items from different groups
+    @tailrec
+    def rearrange(groups: Iterable[List[Node[C]]], agg: List[List[Node[C]]] = Nil): List[List[Node[C]]] = {
+      if (groups.isEmpty) agg
+      else {
+        val current = ListBuffer[Node[C]]()
```
folex: Couldn't it be rewritten shorter as
```
val (current, next) = groups.collect{ case head :: tail => head -> tail }.unzip
rearrange(next.toList, current.toList :: agg)
```
?

kademlia/src/main/scala/fluence/kad/KadError.scala
```diff
@@ -0,0 +1,23 @@
...
+ * limitations under the License.
+ */
+
+package fluence.kad
+
+sealed trait KadError
+
+sealed trait JoinError extends KadError
+
+case object CantJoinAnyNode extends JoinError
```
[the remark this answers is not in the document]
alari: There's only one place in the code where this error occurs, and we have no additional information for it except the fact that it was not possible to join any node.

Once we find out any differences in erroneous scenarios or find any other Kademlia-specific error, we'll make errors hierarchy more complex.

kademlia/src/main/scala/fluence/kad/Kademlia.scala
```diff
@@ -0,0 +1,228 @@
...
+   */
+  def handleRPC: KademliaRpc[C]
+
+  /**
+   * Finds a node by its key, either in a local RoutingTable or doing up to ''maxRequests'' lookup calls
+   *
+   * @param key Kademlia key to find node for
+   * @param maxRequests Max number of remote requests
+   */
+  def findNode(key: Key, maxRequests: Int): F[Option[Node[C]]]
```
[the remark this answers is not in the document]
alari: The meaning is different. `maxNumOfCalls` limits the number of `fn` calls during `callIterative`, while there could be made more requests while looking up the nodes to call `fn` on.  `maxRequests` limits the number of requests literally.

However, I agree that this difference is not obvious from the docs and that naming does not sound, so I'll try to fix it.

kademlia/src/main/scala/fluence/kad/routing/IterativeRoutingImpl.scala
```diff
@@ -0,0 +1,406 @@
...
+        lookedUp: Set[Key],
+        fnCalled: Set[Key],
+        requestsRemaining: Int
+      ): F[Vector[(Node[C], A)]] = {
+        val needCollect = numToCollect - replies.size
+        // If we've collected enough, stop
+        if (needCollect <= 0) replies.pure[F]
+        else {
+          // For idempotent requests, we could make more calls then needed to increase chances to success
+          val callsNeeded = if (isIdempotentFn) parallelism else needCollect min parallelism
```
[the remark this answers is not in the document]
alari: They are colored differently

kademlia/src/main/scala/fluence/kad/routing/IterativeRoutingImpl.scala
```diff
@@ -0,0 +1,406 @@
...
+    EitherT(
+      Parallel
+        .parTraverse(peers.toList) { peer: C ⇒
+          logger.trace("Going to ping Peer to join: " + peer)
+
+          // For each peer
+          // Try to ping the peer; if no pings are performed, join is failed
+          ContactAccess[C].rpc(peer).ping().attempt.to[F].flatMap[Option[(Node[C], List[Node[C]])]] {
+            case Right(peerNode) if peerNode.key =!= localRouting.nodeId ⇒ // Ping successful, lookup node's neighbors
+              logger.info("PeerPing successful to " + peerNode.key)
```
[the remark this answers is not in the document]
alari: It shows the bootstrap stages, which nodes were pinged in what order, etc. I remember it was useful. Let's consider changing logging levels after we look on the logs at least on staging.

kademlia/src/main/scala/fluence/kad/routing/IterativeRoutingImpl.scala
```diff
@@ -0,0 +1,406 @@
...
+            )(p ⇒ ContactAccess[C].rpc(p.contact).ping().attempt.to[F])
+            .map(_.collect {
+              case Right(n) ⇒ n
+            })
+            .map(_ ::: ps)
+
+        }
+        .flatMap { ns ⇒
+          // Save discovered nodes to the routing table
+          logger.info("Discovered neighbors: " + ns.map(_.key))
```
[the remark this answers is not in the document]
alari: Expected messages during bootstrapping could be really useful, especially when you notice that they don't appear.

Issue #672, opened 2019-05-06
Separate smart contract to read & write contracts


folex:
Deployer.sol an Network.sol should be deployed as a separate contracts, so we can change Network.sol without affecting the main contract's address.

PR #681, opened 2019-05-07
Contract for dashboard

### LINKED ISSUE #669: Smart contract method to query appIDs with storageHashes
It should return at least `[(appID, storageHash)]`. Maybe also include `storageReceipt`, so it can be used for custom labels on apps.

This is needed in the dashboard to query apps in a single request, and avoid multi-request latency.

folex:
fixes #669 #672 

Unfortunately, it seems impossible to separate Deployer.sol & Network.sol in a reasonable fashion. At least until ABIEncoderV2 is included in the stable solidity branch. 

Right now, it's not possible to pass structs between different contracts, and this is required to read from `nodes` and `apps` mappings from a separate contract. The only solution I see, is to pass struct as a tuple, but that's exactly what half a methods in `Network.sol` do. So these methods are to be kept in the same contract that defines `nodes` and `apps` mapping. 

And about #669: I've implemented this method in a separate contract, so the main one doesn't need to be redeployed.


bootstrap/contracts/Dashboard.sol
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+pragma solidity ^0.5.8;
+
+import "./Network.sol";
+
+contract Dashboard {
```
DieMyst: docs

bootstrap/contracts/Dashboard.sol
```diff
@@ -0,0 +1,43 @@
...
+
+        bytes32[] memory storageHashes = new bytes32[](appIDs.length);
+        for (uint8 i = 0; i < appIDs.length; i++) {
+            (bytes32 storageHash,,,,,,,) = network.getApp(appIDs[i]);
+            storageHashes[i] = storageHash;
+        }
+
+        return (appIDs, storageHashes);
+    }
+}
```
DieMyst: new line

bootstrap/scripts/compile_sol.js
```diff
@@ -57,15 +60,20 @@ if (!fs.existsSync(compiledPath)){
...
+    }
+
+    writeFile(compiledPath + `/${name}.bin`, contract.evm.bytecode.object);
+    writeFile(compiledPath + `/${name}.abi`, JSON.stringify(contract.abi));
 }
 
-writeFile(compiledPath + '/Network.bin', networkContract.evm.bytecode.object);
-writeFile(compiledPath + '/Network.abi', JSON.stringify(networkContract.abi));
+writeContract("Network");
+writeContract("Dashboard");
```
DieMyst: new line

bootstrap/test/DashboardTest.js
```diff
@@ -0,0 +1,46 @@
...
+    it("Get apps (dashboard)", async function() {
+        await addNodes(3, 3);
+        await addApp(3);
+        await addApp(3);
+        await addApp(3);
+        var result = await global.dashboard.getApps();
+        assert.equal(result[0].length, 3);
+        assert.equal(result[1].length, 3);
+    });
+});
```
DieMyst: new line

PR #694, opened 2019-05-20
Kademlia HTTP transport & Fluence Log facade

alari:


kademlia/http/src/main/scala/fluence/kad/http/UriContact.scala
```diff
@@ -0,0 +1,181 @@
...
+    signer.signWithPK.pointAt(msg).rmap(UriContact(host, port, _))
+  }
+
+  /**
+   * Parse contact from string, check its signature
+   *
+   * @param checkerFn Signature checker function
+   */
+  def readAndCheckContact(checkerFn: CheckerFn): Crypto.Func[String, UriContact] =
+    Crypto.fromOtherFunc(readContact).rmap(i ⇒ (i, i)) >>> checkContact(checkerFn)
```
DieMyst: why not only this?
```
Crypto.fromOtherFunc(readContact) >>> checkContact(checkerFn)
```

kademlia/http/src/main/scala/fluence/kad/http/UriContact.scala
```diff
@@ -0,0 +1,181 @@
...
+      .second[UriContact]
+      .rmap(_._1)
+
+  // codec for base58-encoded public key and signature
+  implicit val pkWithSignatureCodec: (String, String) <~> PubKeyAndSignature = {
```
DieMyst: as I see, you don't use it like implicit class anywhere.

kademlia/src/test/scala/fluence/kad/KeyLawsSpec.scala
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+package fluence.kad
+
+import cats.kernel.laws.discipline.{MonoidTests, OrderTests}
+import cats.tests.CatsSuite
+import fluence.kad.protocol.Key
+import org.scalacheck.{Arbitrary, Gen}
+
+class KeyLawsSpec extends CatsSuite {
```
DieMyst: some comments maybe? because there is no some clue of what happens.

log/src/main/scala/fluence/log/ChainLog.scala
```diff
@@ -0,0 +1,59 @@
...
+import cats.data.Chain
+import cats.syntax.functor._
+import cats.syntax.order._
+import cats.effect.{Clock, Sync}
+import cats.effect.concurrent.Ref
+
+import scala.language.higherKinds
+
+/**
+ * Functional logger facade
```
DieMyst: make a description of this type of logging

log/src/main/scala/fluence/log/ChainLog.scala
```diff
@@ -0,0 +1,59 @@
...
+
+  /**
+   * Provide a logger with modified context
+   *
+   * @param modContext Context modification
+   * @param fn         Function to use the new logger
+   * @tparam A Return type
+   * @return What the inner function returns
+   */
+  override def scope[A](modContext: Context ⇒ Context)(fn: Log[F] ⇒ F[A]): F[A] =
```
DieMyst: it will be great if some examples of usage and output will be in docs.

kademlia/http/src/main/scala/fluence/kad/http/UriContact.scala
```diff
@@ -0,0 +1,181 @@
...
+
+  /**
+   * Build a contact with the given params
+   *
+   * @param host Host
+   * @param port Port
+   * @param signer Signer associated with this node's keypair
+   */
+  def buildContact(host: String, port: Int, signer: Signer): Crypto.Point[UriContact] = {
+    val msg = signer.publicKey.value ++ ByteVector(host.getBytes) ++ ByteVector.fromInt(port)
```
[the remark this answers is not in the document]
alari: There are two ways to get a `UriContact`: one way is when you have a `Signer` to sign your own contact, in this unique situation the `msg` to sign is required _before_ you have `UriContact` instance. Another more common way is when `UriContact` is received from an external node in Kademlia-related response. In this case, it's serialized as a string, and it need to be reconstructed to parts, and then `msg` should be composed to check the signature. I suppose that it could be optimized for both cases, e.g. `msg` might be removed from the `UriContact` class after the signature is checked. I'll leave a TODO for that.

kademlia/http/src/main/scala/fluence/kad/http/KademliaHttpClient.scala
```diff
@@ -0,0 +1,75 @@
...
+class KademliaHttpClient[F[_]: Effect, C](hostname: String, port: Short, auth: String)(
+  implicit s: SttpBackend[EitherT[F, Throwable, ?], Nothing],
+  readNode: Crypto.Func[String, Node[C]]
+) extends KademliaRpc[C] {
+
+  // TODO: do not drop cause
+  private implicit val decodeNode: Decoder[Node[C]] =
+    _.as[String].flatMap(readNode.runEither[Id](_).leftMap(ce ⇒ DecodingFailure(ce.message, Nil)))
+
+  private def call[T: Decoder](call: sttp.type ⇒ Uri ⇒ Request[String, Nothing], uri: Uri): IO[T] =
```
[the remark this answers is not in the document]
alari: I've tried to do it, but let's leave it as it is for now: as we have `Either` in the response body, and as a result of `decode`, it's just more handy to work with `EitherT`.

log/src/main/scala/fluence/log/ChainLog.scala
alari: Let's stabilize the API a little and then write comprehensive docs for the module.

kademlia/testkit/src/test/scala/fluence/kad/KademliaSimulationSpec.scala
```diff
@@ -0,0 +1,185 @@
...
+    "launch with 200 nodes" in {
+      implicit val ctx = Context.init("spec", "launch")
+      implicit val log = ChainLog.forCtx[IO]
+
+      val random = new Random(seed)
+
+      lazy val nodes: Map[Long, Kademlia[IO, Long]] =
+        TestKademlia.simulationIO(K, N, keyToLong, key(random.nextLong()), 3)
+
+      //println("\n\n\n======================================\n\n\n")
```
DieMyst: I think you can delete this

kademlia/src/main/scala/fluence/kad/state/RoutingStateImpl.scala
```diff
@@ -114,13 +114,12 @@ private[state] class RoutingStateImpl[F[_]: Monad, P[_], C](
   )(implicit clock: Clock[F], liftIO: LiftIO[F], ca: ContactAccess[C]): F[ModResult[C]] = {
```
DieMyst: ca - contactAccess, don't save characters

alari: I wish I could omit the implicit argument's name completely. Naming an implicit argument is just a mess if it's not to be used directly: type is enough.

PR #695, opened 2019-05-20
Zero downtime dashboard dep

### LINKED ISSUE #685: Zero-downtime dashboard deployment
Currently, dashboard is served directly from `dashboard/build` which is generated by `npm run build`. 

The problem is that `build` directory is removed upon `npm run build`, and dashboard is unavailable during build times. More to that, if build fails, dashboard will be unavailable until the  next successful build. 

To avoid that, add command `deploy` to dashboard's `package.json`, which will trigger `npm run build`, and then copy `build` to `bundle directory. Dashboard should be served from `bundle` directory.

DieMyst:
fixes #685


dashboard/package.json
```diff
@@ -7,6 +7,7 @@
     "lint": "npm run tslint",
     "tslint": "tslint -c tslint.json -p tsconfig.json",
     "build": "BABEL_ENV=client NODE_ENV=production webpack",
+    "deploy": "npm run build && mkdir bundle && rm -rf bundle/* && cp -r build/* bundle/",
```
folex: Let's make it shorter, and avoid using `*` since it's behaviour isn't always obvious. 

I think this would suffice `npm run build && rm -rf bundle && cp -r build bundle`

Issue #623, opened 2019-04-11
Upload all the Tendermint block data to external storage


alari:
In order to enable delayed validation in the Security layer, first, all the transactions and Tendermint signatures should be uploaded to external storage (e.g. IPFS).

The client can be sure that validation of a distinct transaction will eventually happen if the block that includes it is uploaded to the storage. That should be done prior to reaching Tendermint consensus, so that Tendermint block includes corresponding hash.

PR #698, opened 2019-05-27
KV receipt storage

folex:
part of #623 

The trait `ReceiptStorage` is currently implemented on top of RocksDB KVStore, so master node can store receipts, and server them to state machine after restart.

In future, `ReceiptStorage` could (should) be implemented on top of Kademlia, and used as a drop-in replacement for `KVReceiptStorage`


effects/receipt-storage/src/main/scala/fluence/effects/receipt/storage/ReceiptStorage.scala
```diff
@@ -0,0 +1,42 @@
...
+ * Algebra for storage of block receipts
+ *
+ */
+trait ReceiptStorage[F[_]] {
+  val appId: Long
+
+  /**
+   * Stores receipt for the specified app at a given height
+   *
+   * @param height Block height for the receipt
```
DieMyst: the first word should not be capitalized

build.sbt
```diff
@@ -368,7 +380,7 @@ lazy val `kademlia-testkit` = (project in file("kademlia/testkit"))
     libraryDependencies ++= Seq(
       scalaTest
     )
-  ).dependsOn(`kademlia`)
+  ).dependsOn(kademlia)
```
alari: Backticks plz

effects/receipt-storage/src/main/scala/fluence/effects/receipt/storage/KVReceiptStorage.scala
```diff
@@ -0,0 +1,68 @@
...
+}
+
+object KVReceiptStorage {
+  import cats.syntax.compose._
+  import cats.syntax.flatMap._
+
+  private val ReceiptStoragePath = "receipt-storage"
+
+  private implicit val receiptCodec: codec.PureCodec[Array[Byte], Receipt] =
+    codec.PureCodec.liftB(Receipt.fromBytesCompact, _.bytesCompact())
```
alari: Are you sure it could never fail?

effects/receipt-storage/src/main/scala/fluence/effects/receipt/storage/KVReceiptStorage.scala
```diff
@@ -0,0 +1,68 @@
...
+
+  implicit val stringCodec: PureCodec[String, Array[Byte]] =
+    PureCodec.liftB(_.getBytes(), bs ⇒ new String(bs))
+
+  implicit val longCodec: PureCodec[Long, Array[Byte]] =
+    PureCodec.liftB[Long, String](_.toString, _.toLong) andThen
```
alari: Are you sure these functions are pure?

effects/receipt-storage/src/main/scala/fluence/effects/receipt/storage/KVReceiptStorage.scala
```diff
@@ -0,0 +1,68 @@
...
+  import cats.syntax.compose._
+  import cats.syntax.flatMap._
+
+  private val ReceiptStoragePath = "receipt-storage"
+
+  private implicit val receiptCodec: codec.PureCodec[Array[Byte], Receipt] =
+    codec.PureCodec.liftB(Receipt.fromBytesCompact, _.bytesCompact())
+
+  implicit val stringCodec: PureCodec[String, Array[Byte]] =
+    PureCodec.liftB(_.getBytes(), bs ⇒ new String(bs))
```
alari: Idea is that `PureCodec`s are pure so that they always catch all the errors and lift them into the appropriate type.

An example could be the following: 

```scala
    implicit val stringCodec: PureCodec[String, Array[Byte]] = PureCodec.liftEitherB[String, Array[Byte]](
      str ⇒ scala.util.Try(str.getBytes()).toEither.left.map(t ⇒ fluence.codec.CodecError("Cannot serialize string to bytes", Some(t))),
      bs ⇒ scala.util.Try(new String(bs)).toEither.left.map(t ⇒ fluence.codec.CodecError("Cannot parse bytes to string", Some(t)))
    )
```

Looks weird, but that makes it impossible to have an exception thrown -- and debugging unpredictable obscure exceptions in asynchronous Scala code is even weirder.

Then you just take `PureCodec`s and combine them together using `andThen`, or modify with Profunctor's `rmap`/`lmap` (using only functions which can never throw), having all the guarantees kept: both for composability, error handling, and ability to get inverse (so that `codec.direct andThen codec.inverse` should be an identity).

Codecs are devoted to provide a serialization algebra, so, just changing types in some manner, checking the constraints on the way. In case we have some other source for errors or custom logic, then we need to show it in type system: presumably in types.

That's how `Crypto` arises: it shows different ways to convert things to others, like blob to hash, or to signature, or text to ciphertext and vice versa, which looks similar to codecs but differs in nature. This time you connect smth like `Crypto.Hasher` to smth like `PureCodec` serializer, finally having a function that takes a custom object as an input and provides a hash as a result.

But to perform such a concatenation, you need to lift the errors of one type to another, resulting in a single errors type. That's doable with `Crypto.codec[A, B]` summoner that takes an implicit codec and lifts its errors into Crypto.

PR #703, opened 2019-06-03
update asmble version to fix memory hash calculation

DieMyst:


folex:
Plz, add log about which memory hasher will be used to `MemoryHasher` initialization.

PR #707, opened 2019-06-04
Kademlia config for Node

alari:


node/src/main/scala/fluence/node/KademliaNode.scala
```diff
@@ -72,25 +72,48 @@ object KademliaNode {
...
+          })
+          .flatMap(
+            seeds ⇒
+              Log[F].scope("kad" -> "join")(
+                log ⇒
+                  kad.join(seeds, conf.join.numOfNodes)(log).flatMap {
+                    case true ⇒
+                      log.info("Joined")
+                    case false ⇒
+                      log.warn("Unable to join any Kademlia seed")
```
DieMyst: log seeds somewhere

PR #712, opened 2019-06-07
Store Tendermint blocks in IPFS

### LINKED ISSUE #623: Upload all the Tendermint block data to external storage

folex:
fixes #623 

Please, share your ideas on possible tests to cover all this complex logic. 

Performance testing is yet to be done.


node/src/main/scala/fluence/node/workers/tendermint/BlockUploading.scala
```diff
@@ -51,74 +58,144 @@ class BlockUploading[F[_]: ConcurrentEffect: Timer](history: BlockHistory[F]) ex
...
     }
+
+    // Either height of the last stored receipt, or the 1st block. 1st because we will load it manually.
+    val lastKnownHeight = storedReceipts.last.map(_.map(_._1).getOrElse(1L))
+
+    // Skip first block due to race condition (see details above)
+    val subscriptionBlocks = lastKnownHeight >>= (rpc.subscribeNewBlock(_).dropWhile(_.header.height < 2))
+
+    // Group empty blocks with the first non-empty block; upload empty blocks right away
+    val grouped = (lastOrFirstBlock ++ subscriptionBlocks)
```
DieMyst: perhaps there may be duplication

PR #713, opened 2019-06-07
Priv validator key parse

DieMyst:


node/src/main/scala/fluence/node/MasterNodeApp.scala
```diff
@@ -41,6 +45,33 @@ object MasterNodeApp extends IOApp {
   private val sttpResource: Resource[IO, SttpBackend[EitherT[IO, Throwable, ?], fs2.Stream[IO, ByteBuffer]]] =
     Resource.make(IO(EitherTSttpBackend[IO]()))(sttpBackend ⇒ IO(sttpBackend.close()))
 
+  /**
+   * Reads KeyPair from priv_validator_key.json file in tendermint path.
+   *
+   */
+  private def readTendermintKeyPair(rootPath: String): IO[KeyPair] = {
```
folex: This method should be somewhere in `fluence.node.config` package, e.g., in `fluence.node.config.Configuration`

PR #733, opened 2019-07-03
Kademlia js

DieMyst:


log/js/src/main/scala/fluence/log/DateFormat.scala
```diff
@@ -0,0 +1,26 @@
...
+ * limitations under the License.
+ */
+
+package fluence.log
+
+import java.util.Date
+
+import scala.scalajs.js
+
+object DateFormat {
```
alari: Docs please (explain why it exists).

log/jvm/src/main/scala/fluence/log/DateFormat.scala
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package fluence.log
+
+import java.text.SimpleDateFormat
+import java.util.Date
+
+object DateFormat {
```
alari: Docs, please.

And maybe a shared interface?

node/src/main/scala/fluence/node/MasterNodeApp.scala
```diff
@@ -72,7 +72,7 @@ object MasterNodeApp extends IOApp {
               keyPair <- Resource.liftF(Configuration.readTendermintKeyPair(masterConf.rootPath))
               kad ← KademliaNode.make[IO, IO.Par](
                 masterConf.kademlia,
-                Ed25519.tendermintAlgo,
+                Ed25519.signAlgo,
```
alari: Is it the same?

DieMyst: yes

PR #734, opened 2019-07-05
Kademlia network and Tendermint v0.32.0 

folex:


.scalafmt.conf
```diff
@@ -5,7 +5,7 @@ docstrings = JavaDoc
 maxColumn = 120
 
 align = some
-align.tokens = [{code = "=>", owner = "Case"}]
+align.tokens = [{code = "=>", owner = "Case"}, ":=", "%", "%%"]
```
alari: Add `%%%` for scala.js.

kademlia/http/src/main/scala/fluence/kad/http/KademliaHttp.scala
```diff
@@ -81,7 +81,7 @@ class KademliaHttp[F[_]: Sync, C](
 
           }
 
-        case req @ POST -> Root / "ping" ⇒
+        case req @ (POST | GET) -> Root / "ping" ⇒
```
alari: Actually `ping` is a mutating action: it promotes a pinging node to a pingee. So we could leave `GET` for now, but later on, we should add a `status` endpoint just to read current Kad status.

log/src/main/scala/fluence/log/Context.scala
```diff
@@ -37,8 +37,8 @@ case class Context(data: Map[String, String] = Map.empty, loggingLevel: Log.Leve
 
 object Context {
 
-  def init(k: String, v: String = "", level: Log.Level = Log.Info): Context =
-    Context(Map(k -> v), level)
+  def init(kv: (String, String)*)(level: Log.Level = Log.Info): Context =
```
alari: Let's make two different `init` methods instead.

node/src/main/resources/docker/application.conf
```diff
@@ -41,4 +41,16 @@ worker {
...
 
-kademlia.advertize.host = ${TENDERMINT_IP}
+kademlia {
+  advertize {
+    host = ${EXTERNAL_IP}
+    port = ${API_PORT}
+  }
+
+  join {
+    seeds = [
```
alari: Meh, should it be there? Could we move the seeds to deployment configuration?

log/src/main/scala/fluence/log/Context.scala
folex: unfortunately there can't be two overloaded methods with default arguments

PR #735, opened 2019-07-06
Add redux state typings

krivchun:


dashboard/src/front/actions/ethereum-connection.ts
```diff
@@ -1,7 +1,8 @@
 import { getEthConnectionState, EthConnectionState } from '../../fluence/contract';
+import { AnyAction } from 'redux';
 
 export const UPDATE_ETH_CONNECTION_STATE = 'UPDATE_ETH_CONNECTION_STATE';
-export const updateEthereumConnectionState = (ethConnectionState: EthConnectionState) => {
+export const updateEthereumConnectionState = (ethConnectionState: EthConnectionState): AnyAction => {
```
folex: Is it possible to add a specific type here? Something like `EthConnectionStateUpdateAction` instead of `AnyAction`

dashboard/src/front/actions/ethereum-connection.ts
```diff
@@ -11,7 +12,7 @@ export const updateEthereumConnectionState = (ethConnectionState: EthConnectionS
 /*
  * Reducer
  */
-export default (state: EthConnectionState = getEthConnectionState(), action: any) => {
+export default (state: EthConnectionState = getEthConnectionState(), action: any): EthConnectionState => {
```
folex: Is it possible to avoid using `any` here and use a specific type?

PR #742, opened 2019-07-17
[view only] Block uploading: tests

folex:


build.sbt
```diff
@@ -154,9 +154,18 @@ lazy val `statemachine` = (project in file("statemachine"))
     dockerfile in docker              := DockerContainers.worker(assembly.value, baseDirectory.value)
   )
   .enablePlugins(AutomateHeaderPlugin, DockerPlugin)
-  .dependsOn(`vm`, `statemachine-control`, `tendermint-rpc`, `sttpEitherT`, `tendermint-block`)
+  .dependsOn(
+    `vm`,
+    `statemachine-control`,
+    `statemachine-control` % "test",
```
alari: - If you have `statemachine-control` as a compile dependency, it's already available in the test scope.
- Let's use sbt's `Test` const instead of `"test"` string literal

node/src/test/scala/fluence/node/Eventually.scala
```diff
@@ -0,0 +1,64 @@
...
+import cats.syntax.monadError._
+import org.scalactic.source.Position
+import org.scalatest.exceptions.{TestFailedDueToTimeoutException, TestFailedException}
+import org.scalatest.time.Span
+import org.scalatest.{Timer => _}
+
+import scala.concurrent.duration._
+import scala.language.higherKinds
+
+trait Eventually {
```
DieMyst: some docs will be good here

node/src/test/scala/fluence/node/BlockUploadingSpec.scala
```diff
@@ -0,0 +1,343 @@
...
+      uploadNBlocks(33)
+    }
+
+    "upload 337 blocks" in {
+      uploadNBlocks(337)
+    }
+  }
+
+  "blocks + stored receipts" should {
+    "upload 1 + 2" in {
```
DieMyst: improve test names for better understanding what's going on

node/src/test/scala/fluence/node/BlockUploadingSpec.scala
```diff
@@ -0,0 +1,343 @@
...
+
+  private val rootPath = Paths.get("/tmp")
+
+  val appId = 1L
+  val p2pPort = 10001.toShort
+  val description = "worker #1"
+  val workerPeer = WorkerPeer(ByteVector.empty, "", 25000.toShort, InetAddress.getLocalHost, 0)
+  val cluster = Cluster(currentTime.millis, Vector.empty, workerPeer)
+  val app = App(123L, StorageRef(ByteVector.empty, StorageType.Ipfs), cluster)
+  val dockerConfig = DockerConfig(DockerImage("fluencelabs/worker", "v0.2.0"), DockerLimits(None, None, None))
```
alari: Maybe use image tags from node's config file? To test what's actually is going to be run?

folex: Nothing's going to run here. There's no docker containers involved, it's just mocked data, could be anything.

PR #745, opened 2019-07-17
[view only] Block uploading: node

folex:


node/src/main/scala/fluence/node/workers/tendermint/BlockUploading.scala
```diff
@@ -52,70 +65,129 @@ class BlockUploading[F[_]: ConcurrentEffect: Timer: Log](history: BlockHistory[F
...
   }
+
+  private def uploadBlock(
+    block: BlockUpload,
+    appId: Long,
+    lastManifestReceipt: MVar[F, Option[Receipt]],
+    receiptStorage: ReceiptStorage[F]
+  )(implicit backoff: Backoff[EffectError], log: Log[F]): F[List[Receipt]] =
+    log.scope("block" -> block.block.header.height.toString, "upload block" -> "") { log =>
+      def logError[E <: EffectError](e: E) = log.error("", e)
```
DieMyst: empty message

PR #754, opened 2019-07-23
Timeouts for remote storage 

folex:


effects/ipfs/src/main/scala/fluence/effects/ipfs/IpfsStore.scala
```diff
@@ -49,7 +50,8 @@ class IpfsStore[F[_]](client: IpfsClient[F]) extends ContentAddressableStore[F]
 object IpfsStore {
 
   def apply[F[_]](
```
alari: Docs on what happens when `readTimeout` is exhausted?

PR #757, opened 2019-07-25
Access last worker's block manifest via HTTP

alari:


node/src/main/scala/fluence/node/workers/DockerWorkersPool.scala
```diff
@@ -178,7 +183,7 @@ class DockerWorkersPool[F[_]: DockerIO: Timer, G[_]](
             // stop the old worker
             _ ← oldWorker.fold(().pure[F])(stop)
 
-            _ ← runWorker(p2pPort, params)
+            _ ← runWorker(p2pPort, params, 5, rootPath)
```
DieMyst: name for `5` parameter

node/src/main/scala/fluence/node/workers/WorkerBlockManifests.scala
```diff
@@ -0,0 +1,83 @@
...
+
+import scala.language.higherKinds
+
+/**
+ * Worker's block manifests service
+ *
+ * @param receiptStorage Receipt storage
+ * @param lastManifestRef Last stored manifest, could be empty
+ */
+class WorkerBlockManifests[F[_]: Monad](
```
DieMyst: name of class and docs says nothing about what this class is doing. Maybe something like `LastManifestStorage`

alari: It's intended to cover more Block Manifests-related functions later, so I don't want to narrow down its name just to LastManifestStorage or smth similar. Let's let it grow, taking responsibility for more worker's block manifests functionality, and then try to find out a better class name.

effects/receipt-storage/src/main/scala/fluence/effects/receipt/storage/KVReceiptStorage.scala
```diff
@@ -67,4 +65,7 @@ object KVReceiptStorage {
       path <- Resource.liftF(Sync[F].catchNonFatal(storagePath.resolve(ReceiptStoragePath).resolve(appId.toString)))
       store <- RocksDBStore.make[F, Long, Receipt](path.toAbsolutePath.toString)
     } yield new KVReceiptStorage[F](appId, store)
+
+  def makeInMemory[F[_]: Concurrent](appId: Long): Resource[F, ReceiptStorage[F]] =
```
folex: doc

node/src/main/scala/fluence/node/workers/WorkerServices.scala
```diff
@@ -34,4 +34,6 @@ trait WorkerServices[F[_]] {
   // Retrieves worker's health
   def status(timeout: FiniteDuration): F[WorkerStatus]
 
+  // Block manifests services: uploading, retrieving
+  def blockManifests: WorkerBlockManifests[F]
```
folex: By the name, I would assume it's a kind of block manifest stream. So maybe use another name? Like `blockServices`

node/src/main/scala/fluence/node/workers/WorkerBlockManifests.scala
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+  /**
+   * Updates the last known manifest
+   * TODO broadcast manifest to Kademlia
+   *
+   * @param manifest Block manifest
+   * @param receipt This block manifest's receipt
+   */
+  def onUploaded(manifest: BlockManifest, receipt: Receipt): F[Unit] =
```
folex: Why do you need receipt here?

node/src/main/scala/fluence/node/workers/WorkerBlockManifests.scala
```diff
@@ -0,0 +1,83 @@
...
+    lastManifestRef.tryTake >> lastManifestRef.put(manifest)
```
folex: If we were to make block uploading concurrent, this could deadlock. And it is possible to upload empty blocks in parallel with non-empty ones, at least could be implemented some day. 

Also, if there is some concurrency, manifest of height N could be replaced by manifest of height N-1.

Maybe use Ref's modify with height checking here? Or a queue with element skipping like in ControlSignals with vmHashQueue and receiptQueue

node/src/main/scala/fluence/node/workers/WorkerBlockManifests.scala
```diff
@@ -0,0 +1,83 @@
...
+ */
+class WorkerBlockManifests[F[_]: Monad](
+  val receiptStorage: ReceiptStorage[F],
+  lastManifestRef: MVar[F, BlockManifest]
+) {
+
+  /**
+   * Returns the last known manifest, blocks until there's any
+   */
+  def lastManifest: F[BlockManifest] =
```
folex: Can you please describe in doc who will call this method? Currently I don't fully understand how it will be used. For example, will it be retried (e.g., via backoff)? Will it be called through network?

node/src/main/scala/fluence/node/workers/WorkerBlockManifests.scala
```diff
@@ -0,0 +1,83 @@
...
+   * TODO broadcast manifest to Kademlia
+   *
+   * @param manifest Block manifest
+   * @param receipt This block manifest's receipt
+   */
+  def onUploaded(manifest: BlockManifest, receipt: Receipt): F[Unit] =
+    lastManifestRef.tryTake >> lastManifestRef.put(manifest)
+}
+
+object WorkerBlockManifests {
```
folex: Maybe rename `WorkerBlockManifests` => `WorkerBlockManifestsServices`?

node/src/main/scala/fluence/node/workers/WorkerBlockManifests.scala
```diff
@@ -0,0 +1,83 @@
...
+   * Returns the last known manifest, blocks until there's any
+   */
+  def lastManifest: F[BlockManifest] =
+    lastManifestRef.read
+
+  /**
+   * Returns the last known manifest, if any
+   */
+  def lastManifestOpt: F[Option[BlockManifest]] =
+    lastManifestRef.tryTake.flatMap {
```
folex: Why not use Ref?

node/src/main/scala/fluence/node/workers/WorkersHttp.scala
```diff
@@ -113,6 +113,24 @@ object WorkersHttp {
             }
         }
 
+      case GET -> Root / LongVar(appId) / "lastManifest" ⇒
```
folex: How this endpoint would be used? Some service will poll it? Maybe block until there's a receipt?

node/src/test/scala/fluence/node/BlockUploadingSpec.scala
```diff
@@ -141,16 +141,19 @@ class BlockUploadingSpec extends WordSpec with Matchers with Eventually with Opt
 
           override def status(timeout: FiniteDuration): IO[WorkerStatus] =
             IO.raiseError(new NotImplementedError("def status worker status"))
+
+          override def blockManifests: WorkerBlockManifests[IO] =
+            new WorkerBlockManifests[IO](receiptStorage(appId), MVar.empty[IO, BlockManifest].unsafeRunSync())
```
folex: can you please initialize it above, and send here via flatMap? So everything is created in IO

node/testkit/src/test/scala/fluence/node/BlockUploadingIntegrationSpec.scala
```diff
@@ -162,12 +162,15 @@ class BlockUploadingIntegrationSpec extends WordSpec with Eventually with Matche
 
       override def status(timeout: FiniteDuration): IO[WorkerStatus] =
         IO.raiseError(new NotImplementedError("def status worker status"))
+
+      override def blockManifests: WorkerBlockManifests[IO] =
+        new WorkerBlockManifests[IO](receiptStorage(appId), MVar.empty[IO, BlockManifest].unsafeRunSync)
```
folex: can you please initialize it above, and give here via flatMap?

node/src/main/scala/fluence/node/workers/WorkersHttp.scala
alari: Currently, that's for testing purposes only. Later on, I plan to use the same endpoint to access Kademlia-based interim storage for the last manifest. Thence the endpoint will be used by the client to check that manifest is uploaded and stored in the interim storage.

I don't think it worth polishing inside this PR.

node/src/main/scala/fluence/node/workers/WorkerBlockManifests.scala
alari: The choice is: either use ref of option and fill it with `None` initially or use `MVar` which can hold no value naturally. I chose `MVar` for now, but not sure it will last for a long time.

node/src/main/scala/fluence/node/workers/WorkerBlockManifests.scala
alari: It's too verbose, that's why this convention isn't used in other worker services.

I think we could move `BlockUploading` somewhere inside (as it's a part of a worker's lifecycle, not node's) and figure out a better name for both later.

node/src/main/scala/fluence/node/workers/WorkerBlockManifests.scala
alari: Please note that this method is idempotent, it doesn't empty the `MVar`. It blocks only until the very first value is inside. So there's no problem to use backoff, call this method via a network etc.

I don't think doc should cover the usage of basic accessors. All things that matter (blocking until there's any value) are described.

node/src/main/scala/fluence/node/workers/WorkerServices.scala
alari: I agree that we should find out a better name. But `services` suffix seems too verbose and not convenient.

node/src/main/scala/fluence/node/workers/WorkerBlockManifests.scala
alari: Switched to Ref to avoid potential deadlocks.

node/src/main/scala/fluence/node/workers/WorkerBlockManifests.scala
alari: Switched to Ref. I'm planning to add height checking later, but don't want to overblow this PR.

node/src/main/scala/fluence/node/workers/tendermint/BlockUploading.scala
```diff
@@ -133,21 +132,20 @@ class BlockUploading[F[_]: ConcurrentEffect: Timer: ContextShift](
...
       case Left(empties) => fs2.Stream.emits(empties.toList.takeRight(1))
       case Right(block)  => fs2.Stream.eval(upload(block))
-    }.map(_ -> ReceiptType.New)
+    }
 
     // Receipts from storage; last one will be treated differently, see AbciService for details
     val storedTypedReceipts =
-      storedReceipts.dropLast.map(_._2 -> ReceiptType.Stored) ++
-        storedReceipts.takeRight(1).map(_._2 -> ReceiptType.LastStored)
+      storedReceipts.dropLast.map(_._2) ++
```
folex: You can now use `storedReceipts` right away, there's no need to separate it into prefix and last elements

node/src/main/scala/fluence/node/workers/tendermint/BlockUploading.scala
```diff
@@ -133,21 +132,20 @@ class BlockUploading[F[_]: ConcurrentEffect: Timer: ContextShift](
...
+        storedReceipts.takeRight(1).map(_._2)
 
     // Send receipts to the state machine (worker)
-    val receipts = (storedTypedReceipts ++ newReceipts).evalMap(sendReceipt _ tupled)
+    val receipts = (storedTypedReceipts ++ newReceipts).evalMap(sendReceipt _)
```
folex: you can omit `_` here, I think

PR #762, opened 2019-07-31
Kademlia DHT as a ReceiptStorage backend

alari:


kademlia/dht/src/main/scala/fluence/kad/dht/DhtLocalStore.scala
```diff
@@ -0,0 +1,117 @@
...
+      // Ignore all errors while getting old value
+      oldValueOpt ← EitherT.right(
+        store
+          .get(key)
+          .value
+          .map(_.toOption.flatten)
+      )
+
+      // If the value for the key was previously stored, combine it with the new one
+      newValue = oldValueOpt.fold(value)(_ |+| value)
```
folex: Why always combine values?

kademlia/dht/src/main/scala/fluence/kad/dht/DhtLocalStore.scala
```diff
@@ -0,0 +1,117 @@
...
+      dhtMetadata = DhtValueMetadata(timestamp)
+
+      _ ← store.put(key, newValue).leftMap(DhtLocalStoreError(_))
+
+      _ ← metadata.put(key, dhtMetadata).leftMap(DhtLocalStoreError(_))
```
folex: What if `store.put` succeeded, but `metadata.put` hasn't?

kademlia/dht/src/main/scala/fluence/kad/dht/DhtLocalStore.scala
folex: Maybe call it `append` instead of `store`? `key: K, value: V` looks like a KV storage

node/src/main/scala/fluence/node/workers/tendermint/BlockUploading.scala
```diff
@@ -86,64 +87,79 @@ class BlockUploading[F[_]: ConcurrentEffect: Timer: ContextShift](
...
-    val newReceipts = grouped.flatMap {
-      // Emit receipts for the empty blocks
-      case Left(empties) => fs2.Stream.emits(empties.toList.takeRight(1))
-      case Right(block)  => fs2.Stream.eval(upload(block))
-    }
-
-    // Send receipts to the state machine (worker)
-    val receipts = (storedReceipts.map(_._2) ++ newReceipts).evalMap(sendReceipt)
+  )(implicit backoff: Backoff[EffectError], F: Applicative[F], log: Log[F]): Resource[F, Unit] =
+    Resource.liftF((Ref.of[F, Long](0), Deferred[F, Long]).tupled).flatMap {
```
folex: Maybe pass these as function arguments? Like `lastManifestReceipt: MVar[F, Option[Receipt]]`. This should lower indentation level, and make it easier to read.

node/src/main/scala/fluence/node/workers/tendermint/DhtReceiptStorage.scala
```diff
@@ -0,0 +1,90 @@
...
+  /**
+   * Retrieves a chain of receipts, starting at block height `from`, until `to`.
+   *
+   * @param from height of the first receipt in the resulting chain
+   * @param to   height of the last receipt in the resulting chain
+   * @return chain of receipts
+   */
+  override def retrieve(from: Option[Long], to: Option[Long])(implicit log: Log[F]): fs2.Stream[F, (Long, Receipt)] =
+    fs2.Stream
+      .iterate(0l)(_ + 1)
```
folex: Shouldn't it be `iterate(from.getOrElse(0L))`?

kademlia/dht/src/main/scala/fluence/kad/dht/Dht.scala
```diff
@@ -0,0 +1,142 @@
...
+   * @param results Number of results to collect and merge
+   * @param maxCalls Maximum number of network calls
+   */
+  def retrieve(key: Key, results: Int, maxCalls: Int)(implicit log: Log[F]): EitherT[F, DhtError, V] =
+    EitherT(
+      kad.callIterative(key, n ⇒ rpc(n.contact).retrieve(key), results, maxCalls).map {
+        case sq if sq.isEmpty ⇒
+          Left(DhtValueNotFound(key))
+        case sq ⇒
+          Right(sq.map(_._2).reduce(_ |+| _))
```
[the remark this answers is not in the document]
alari: Because DHT is all about assigning values to hashed keys. If we ask for a value, we get a single value. Reducing here could have different meaning indeed, for example:
- For two integer values, get max
- For two objects, drop the old one, pick the fresh (e.g. network was reorganized, and some nodes have stale value)
- Value type could be naturally combinable, for example, consider `type V = Set[Any]`; in this case, reducing is not to "scalar" value

So this reducing does not imply any limitations.

Currently, we use it just to choose the receipt with greater height.

kademlia/dht/src/main/scala/fluence/kad/dht/Dht.scala
```diff
@@ -0,0 +1,142 @@
...
+   *
+   * @param key Key
+   * @param value Value
+   * @param replication Replication factor: how many remote nodes should store the value
+   * @param maxCalls The upper bound for an overall number of calls made, see [[Kademlia.callIterative]] for details
+   */
+  def store(key: Key, value: V, replication: Int, maxCalls: Int)(implicit log: Log[F]): EitherT[F, DhtError, Unit] =
+    EitherT(
+      kad
+        .callIterative(key, n ⇒ rpc(n.contact).store(key, value), replication, maxCalls, isIdempotentFn = false)
```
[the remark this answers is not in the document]
alari: Maybe `idempotent` is the wrong word here. What I mean is that the function call _changes_ the state of the node it's called upon (it consumes storage). That's why I say: don't make more calls with this function then it's necessary. It disables the `callIterative` internal optimization with exhaustive parallel requests.

kademlia/dht/src/main/scala/fluence/kad/dht/DhtLocalStore.scala
alari: It _is_ a KV storage. It stores a single value for a key. In case there was another value stored, the choice is described in terms of `Semigroup` algebra.

kademlia/dht/src/main/scala/fluence/kad/dht/DhtLocalStore.scala
alari: In this case:
- RPC PUT call will fail, caller will try to store the value on another node
- Subsequent RPC GET call will resolve with the value
- The value will never be republished, nor will it be garbage-collected

node/src/main/scala/fluence/node/workers/tendermint/BlockUploading.scala
alari: There's no need to leak the knowledge about the internals (how exactly the last block height is derived). I'm planning to refactor the whole method later, so let's keep it as is.

kademlia/dht/src/main/scala/fluence/kad/dht/Dht.scala
```diff
@@ -0,0 +1,142 @@
...
+    retrieve(key, conf.retrieveResults, conf.maxRetrieveCalls)
+      .map[Option[V]](Some(_))
+      .leftFlatMap[Option[V], KVReadError] {
+        case _: DhtValueNotFound ⇒ EitherT.rightT(None)
+        case err ⇒ EitherT.leftT(IOExceptionError("Unable to retrieve value from DHT", err))
+      }
+
+  override def put(key: Key, value: V)(implicit log: Log[F]): EitherT[F, KVWriteError, Unit] =
+    store(key, value, conf.replicationFactor, conf.maxStoreCalls)
+      .leftMap(err ⇒ IOExceptionError("Unable to store value into DHT", err))
```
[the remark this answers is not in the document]
alari: I'm not sure it's useful. In some Kademlia implementations, they even hide `Key` completely, as it's just like forcing users to know what is the hash of each key in a hashtable.

kademlia/dht/src/main/scala/fluence/kad/dht/DhtLocalStore.scala
alari: So this rule prevents us from replacing a `receipt(height=10)` with a `receipt(height=9)`. This kind of publishing calls is very likely to happen during network reorganizations when some nodes stop receiving updates but eventually decide to republish obsolete values they already have.

kademlia/dht/src/main/scala/fluence/kad/dht/Dht.scala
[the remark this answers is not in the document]
folex: I mean, if you would send several `put`s in async fashion, and then got a few replies `unable to store`, it wouldn't tell you much. I'm looking at this from UX pov.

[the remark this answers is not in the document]
alari: The key in the string message won't tell you more, it's better to handle the error on the call site. Anyway, I have added `key` to the error message.

kademlia/dht/src/main/scala/fluence/kad/dht/Dht.scala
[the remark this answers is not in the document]
folex: What I mean is that idempotency here is dependent on the Semigroup properties. For example, adding to a Set is idempotent (given there's no removal), storing receipt of higher height–too, but e.g., appending to a List–isn't.

`isIdempotent` enables some optimizations because it relaxes requirements–anything idempotent could be simply repeated, so I just thought maybe it should be implementation dependent.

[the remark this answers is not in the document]
alari: I should rename the flag to smth like `isMutating` or please suggest a better name.

kademlia/dht/src/main/scala/fluence/kad/dht/DhtLocalStore.scala
folex: Maybe add this as a comment? 

Also, maybe write metadata first, so the data will be garbage collected at some point?

PR #768, opened 2019-08-03
Gas metering support

### LINKED ISSUE #165: Implement gas counter
**Motivation**
Since we allow users to execute arbitrary WASM code in our ecosystem we should provide a solution for halting problem. So we need to have some Ethereum gas alternative. As in Ethereum gas is a unit that measures the amount of computational effort that it will take to execute certain WASM instructions. So we also need a way to count it during wasm code execution.

**Proposed changes**
There are two possible ways to implement gas counter:
- add instructions for increment gas counter after each each generated JVM instructions correspond to one WASM instruction;
- add instructions for increment gas counter after each JVM basic blocks;
The second variant is more complicated in implementation but faster.

michaelvoronov:
Resolves #165


vm/src/main/scala/fluence/vm/WasmVm.scala
```diff
@@ -164,22 +169,63 @@ object WasmVm {
...
-            config.invokeFunctionName,
-            memoryHasher
-          )
-
-        } yield acc + (wasmModule.getName → wasmModule)
-    }
+  ): EitherT[F, ApplyError, (MainWasmModule, EnvModule, Seq[WasmModule])] =
+    for {
+
+      rawEnvModule ← EitherT.fromEither[F](
```
alari: You can use `EitherT.cond` instead.

vm/src/main/scala/fluence/vm/WasmVm.scala
```diff
@@ -164,22 +169,63 @@ object WasmVm {
...
+        config.envModuleConfig.setSpentGasFunction
+      )
+
+      (mainModule, sideModules) ← Traverse[List]
+        .foldLeftM[EitherT[F, ApplyError, ?], Compiled, (Option[MainWasmModule], List[WasmModule])](
+          scriptCxt.getModules.toList,
+          (None, Nil)
+        ) {
+          // the main module almost always has non set module name
+          case (acc, moduleDescription) if Option(moduleDescription.getName) == config.mainModuleConfig.name ⇒
```
alari: This equation looks very confusing. Is it supposed that you compare None with None there?

vm/src/main/scala/fluence/vm/config/VmConfig.scala
```diff
@@ -22,13 +22,26 @@ import pureconfig.generic.ProductHint
 
 import scala.util.control.NoStackTrace
 
+case class MainModuleConfig(
+  name: Option[String],
```
alari: Maybe remove main module name from config and always expect it to be empty? Is there any benefit in having this configurable?

vm/src/main/scala/fluence/vm/wasm/module/EnvModule.scala
```diff
@@ -0,0 +1,113 @@
...
+ * This module could be used for gas metering.
+ *
+ * @param instance a instance of Wasm Module compiled by Asmble
+ * @param spentGasFunction a function returns a spent gas count
+ * @param setSpentGasFunction a function sets spent gas count
+ */
+class EnvModule(
+  private val instance: ModuleInstance,
+  private val spentGasFunction: WasmFunction,
+  private val setSpentGasFunction: WasmFunction
```
alari: What does `setSpentGasFunction` do? Does it set spent gas? Or does it add to it? How should the gas counter be reset to 0 after function invocation?

vm/src/main/scala/fluence/vm/WasmVm.scala
michaelvoronov: Yes, it is. I will add some info to the comment above.

vm/src/main/scala/fluence/vm/wasm/module/EnvModule.scala
michaelvoronov: Rename it to `clearState`, since env module will return either gas and eic, and they could be named as State.

vm/src/main/scala/fluence/vm/config/VmConfig.scala
michaelvoronov: Wasm, Rust, AS are too dynamic and, potentially, they could add the name section by default to modules.

vm/src/main/scala/fluence/vm/WasmVm.scala
```diff
@@ -156,22 +161,61 @@ object WasmVm {
...
+      rawEnvModule ← EitherT.cond[F](
+        scriptCxt.getRegistrations.containsKey(config.envModuleConfig.name),
+        scriptCxt.getRegistrations.get(config.envModuleConfig.name),
+        NoSuchModuleError(
+          s"Asmble doesn't provide the environment module with name=${config.envModuleConfig.name} (perhaps you are using the old version)"
+        ): ApplyError
+      )
+
+      envModule ← EnvModule[F](
+        rawEnvModule.asInstanceOf[Native],
```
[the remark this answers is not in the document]
michaelvoronov: Good point, thank you!

PR #776, opened 2019-08-06
Force bouncycastle to 1.60

folex:
Fixes `java.lang.ClassCastException: org.bouncycastle.jcajce.provider.asymmetric.ec.BCECPrivateKey cannot be cast to org.bouncycastle.jcajce.provider.asymmetric.ec.BCECPrivateKey` error in tests. 

The cause was that web3j uses 1.60, and we were using 1.61.


folex:
Hmmm, this doesn't seem to fix it :(

folex:
Will wait for reply https://github.com/web3j/web3j/issues/539

Issue #771, opened 2019-08-05
Implement EIC counter for normal execution mode


michaelvoronov:
According to our protocol, executed instruction counter (EIC) is used in a dispute and in VM state hash computing. So it needs to track it along with gas (#165) in normal VM execution mode.

michaelvoronov:
Depends on https://github.com/cretz/asmble/issues/30

michaelvoronov:
With the choice of using Wasmer instead of Asmble, it isn't depend anymore and will be added in [this](https://github.com/fluencelabs/fluence/pull/847) PR.

PR #784, opened 2019-08-08
Add EIC metering to wasm-utils

michaelvoronov:
Related to #771.


tools/wasm-utils/src/main.rs
```diff
@@ -16,29 +16,89 @@
...
 use exitfailure::ExitFailure;
 use failure::err_msg;
+use parity_wasm::elements::Module;
 
 const VERSION: &str = env!("CARGO_PKG_VERSION");
 const AUTHORS: &str = env!("CARGO_PKG_AUTHORS");
 const DESCRIPTION: &str = env!("CARGO_PKG_DESCRIPTION");
 
-const MODULE_PATH: &str = "module_path";
+const IN_MODULE_PATH: &str = "in_module_path";
```
folex: maybe just `in-module-path` and `out-module-path`? So it more unix-y: `--in-module-path`, `--out-module-path`

tools/wasm-utils/src/main.rs
```diff
@@ -16,29 +16,89 @@
...
+const OUT_MODULE_PATH: &str = "out_module_path";
 
 fn prepare_wasm_file<'a, 'b>() -> App<'a, 'b> {
-    let arg = &[
-        Arg::with_name(MODULE_PATH)
+    let args = &[
+        Arg::with_name(IN_MODULE_PATH)
             .required(true)
```
folex: let's add `.short("in")`

tools/wasm-utils/src/main.rs
```diff
@@ -16,29 +16,89 @@
...
             .takes_value(true)
             .help("path to the wasm file"),
+        Arg::with_name(OUT_MODULE_PATH)
+            .required(true)
```
folex: let's add `.short("o")` , so it would be `-in /path/to.wasm -o /another/path`

tools/wasm-utils/src/main.rs
```diff
@@ -16,29 +16,89 @@
...
     SubCommand::with_name("prepare")
-        .about("Prepare wasm file to run on the Fluence network")
-        .args(arg)
+        .about("Prepare a wasm file to run on the Fluence network")
+        .args(args)
+}
+
+fn gas_metering<'a, 'b>() -> App<'a, 'b> {
+    let args = &[
+        Arg::with_name(IN_MODULE_PATH)
```
folex: I think you could've made paths args top level, so you don't have to repeat them for each subcommand

michaelvoronov: It will require heap-allocation and has to be thread-safe (lazy_static, ). I think the more suitable variant here is to put them to a separate function.

PR #794, opened 2019-08-12
Wait for response on 'tx already in cache'

folex:
Also more debug logs to debug tx timeouts


node/src/main/scala/fluence/node/workers/WorkerApiImpl.scala
```diff
@@ -82,9 +82,14 @@ class WorkerApiImpl extends WorkerApi {
...
         .broadcastTxSync(tx, id.getOrElse("dontcare"))
         .leftMap(RpcTxAwaitError(_): TxAwaitError)
-      response <- log.scope("sessionId" -> txParsed.head.toString) { implicit log =>
+      _ <- Log.eitherT.debug("TendermintRpc broadcastTxSync is ok.")
+      response <- log.scope("tx.head" -> txParsed.head.toString) { implicit log =>
         for {
-          _ <- checkTxResponse(txBroadcastResponse)
+          _ <- checkTxResponse(txBroadcastResponse).recoverWith {
+            // Transaction was sent twice, but response should be available, so keep waiting
+            case e: TendermintError if e.data.toLowerCase.contains("tx already exists in cache") =>
```
alari: Is there any way to avoid parsing the error message? Or at least not depend on the exact wording?

node/src/main/scala/fluence/node/workers/WorkersHttp.scala
```diff
@@ -152,38 +152,62 @@ object WorkersHttp {
...
+                      case Right(queryResponse) =>
+                        queryResponse match {
+                          case OkResponse(txHead, response) =>
+                            log.debug(s"tx.head: $txHead -> OK $response") >> Ok(response)
+                          case RpcErrorResponse(txHead, r) =>
+                            // TODO ERROR INCONSISTENCY: some errors are returned as 200, others as 500
+                            log.scope("tx.head" -> txHead.toString)(implicit log => rpcErrorToResponse(r))
+                          case TimedOutResponse(txHead, tries) =>
+                            log.warn(s"tx.head: $txHead timed out after $tries") >>
+                              // TODO: ERROR INCONSISTENCY: some errors are returned as 200, others as 500
```
alari: Here response code is neither 200 nor 500. Let's leave this TODO only in places where response code is actually inconsistent.

node/src/main/scala/fluence/node/workers/WorkersHttp.scala
```diff
@@ -152,41 +152,62 @@ object WorkersHttp {
...
+                  .flatMap(tendermintResponseToHttp(appId, _))
+            )
           }
         }
 
       case req @ POST -> Root / LongVar(appId) / "txWaitResponse" :? QueryId(id) ⇒
         LogFactory[F].init("http" -> "txAwaitResponse", "app" -> appId.toString) >>= { implicit log =>
           req.decode[String] { tx ⇒
-            log.info(s"tx: ${tx.takeWhile(_ != '\n')}") >>
+            log.info(s"tx.head: ${tx.takeWhile(_ != '\n')}") >>
```
alari: Please move this `tx.takeWhile(_ != '\n')` to a `val` with a meningful name and reuse it.

node/src/main/scala/fluence/node/workers/subscription/TxWaitError.scala
```diff
@@ -17,12 +17,24 @@
...
 /**
  * Errors for `txAwait` API
  */
 trait TxAwaitError
-case class TendermintResponseError(responseError: String) extends TxAwaitError
+case class TendermintResponseDeserializationError(responseError: String) extends TxAwaitError
 case class RpcTxAwaitError(rpcError: RpcError) extends TxAwaitError
 case class TxParsingError(msg: String, tx: String) extends TxAwaitError
 case class TxInvalidError(msg: String) extends TxAwaitError
+case class TendermintError(code: Int, message: String, data: String) extends TxAwaitError
```
alari: Is it a statemachine error? The name is too generic, what cases does it cover?

node/src/main/scala/fluence/node/workers/WorkerApiImpl.scala
folex: Doesn't seem like there is. The error looks like the following: 
```json
{
  "jsonrpc": "2.0",
  "id": "dontcare",
  "error": {
    "code": -32603,
    "message": "Internal error",
    "data": "Tx already exists in cache"
  }
}
```

`code` is the same for all `Internal error` errors.

node/src/main/scala/fluence/node/workers/WorkersHttp.scala
folex: Inconsistency isn't within exact error codes, the problem is that we return some errors as data, wrapped in JSON, and others as HTTP errors.

PR #803, opened 2019-08-14
Give 100 mb per wasm module in frun

folex:


vm/frun/src/main/resources/reference.conf
```diff
@@ -6,7 +6,7 @@ fluence.vm.client {
 
   # The maximum number of memory pages when a module doesn't say
   # Each page contains 65536 bytes of data, `65536*64=4MB`
-  defaultMaxMemPages: 64
+  defaultMaxMemPages: 1600
```
michaelvoronov: Maybe update the comments above (also above the second parameter)?

PR #809, opened 2019-08-16
Ability to provide unique module with name as the main

michaelvoronov:


vm/src/main/scala/fluence/vm/WasmVm.scala
```diff
@@ -194,8 +194,9 @@ object WasmVm {
           (None, Nil)
         ) {
           // the main module almost always doesn't have name section (in config it is represented by None)
+          // also if there is only one module provided, it is considered as the main regardless of its name.
           case ((None, sideModules), moduleDescription)
-              if Option(moduleDescription.getName) == config.mainModuleConfig.name ⇒
+              if Option(moduleDescription.getName) == config.mainModuleConfig.name || scriptCxt.getModules.size() == 1 ⇒
```
alari: In case there's a lot of modules registered within `scriptCxt` (btw: type, should be `ctx`), they will be counted every time -- at least if `size()` function is not cached.

PR #821, opened 2019-08-28
Worker and ControlRpc status updates

alari:
Return worker status from `/status` endpoint, tendermint block from `/status/tendermint` endpoint;

Pass ControlStatus case class via ControlRpc.


node/src/main/scala/fluence/node/workers/WorkersHttp.scala
```diff
@@ -155,9 +158,18 @@ object WorkersHttp {
           withWorker(appId)(w => workerApi.query(w, data, path, id).flatMap(tendermintResponseToHttp(appId, _)))
         }
 
-      case GET -> Root / LongVar(appId) / "status" ⇒
+      case GET -> Root / LongVar(appId) / "status" :? QueryWait(wait) ⇒
```
folex: What is wait? Let's add some doc

node/src/main/scala/fluence/node/workers/WorkersHttp.scala
```diff
@@ -155,9 +158,18 @@ object WorkersHttp {
...
         LogFactory[F].init("http" -> "status", "app" -> appId.toString) >>= { implicit log =>
-          withWorker(appId)(w => workerApi.status(w).flatMap(tendermintResponseToHttp(appId, _)))
+          withWorker(appId)(
+            _.services
+              .status(wait.filter(_ < 10).fold(1.second)(_.seconds))
```
folex: Why 10? What's happening here? Maybe a comment would help

statemachine/control/src/main/scala/fluence/statemachine/control/ControlStatus.scala
```diff
@@ -0,0 +1,34 @@
...
+import io.circe.{Decoder, Encoder}
+import io.circe.generic.semiauto._
+
+/**
+ * Statemachine's status, fetched via Control RPC.
+ *
+ * @param expectEth Whether this statemachine expects to get Ethereum blocks or not
+ */
+case class ControlStatus(
+  expectEth: Boolean
```
folex: `expectsEth` would be better, I think

node/src/main/scala/fluence/node/workers/WorkersHttp.scala
alari: Tbh I have no reason either for 10 nor for 1 second. Just two constants that "should be long enough". But I have added a comment.

PR #838, opened 2019-09-12
Move node & worker builds to Dockerfiles

folex:


build.sbt
```diff
@@ -131,10 +133,10 @@ lazy val `statemachine-docker` = (project in file("statemachine/docker"))
     assemblyMergeStrategy in assembly := SbtCommons.mergeStrategy.value,
     test in assembly                  := {},
     parallelExecution in Test         := false,
-    imageNames in docker              := Seq(ImageName(DockerContainers.Worker)),
-    dockerfile in docker              := DockerContainers.worker(assembly.value, baseDirectory.value)
+    docker                            := { runCmd(s"make worker TAG=v${version.value}") },
+    docker in Test                    := { assembly.value; runCmd("make worker-test") }
```
alari: Let's use dependsOn or similar technique instead of imperative `;`

log/src/main/scala/fluence/log/Log.scala
```diff
@@ -203,8 +203,8 @@ object Log {
   }
   case object Trace extends Level(0, "trace", Console.WHITE)
   case object Debug extends Level(1, "debug", Console.MAGENTA)
-  case object Info extends Level(2, "info ", Console.BLUE)
-  case object Warn extends Level(3, "warn ", Console.RED)
+  case object Info extends Level(2, "info", Console.BLUE)
```
alari: These spaces were kept to provide consistent indentation in logs (all names were 5 chars long). Could you please keep this behaviour in logs rendering either by changing serialization to use object names instead of the `name` attribute or by changing log string builder?

PR #841, opened 2019-09-18
Copy dockerfiles to be compatible with Dockerhub

folex:


tools/docker/Worker.Dockerhub.Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+############## Copy jar from local fs for tests, master-node.jar should be prebuilt
+FROM scratch as test
+COPY . /fluence
+
+############## Aux dynamic stage, could be either test or production
+FROM $environment as build
+
+############## Build final image
+FROM openjdk:8-jre-alpine
+VOLUME /worker
```
alari: Should we update these `worker`-related sections with smth about `statemachine`?

PR #843, opened 2019-09-19
Implement query response for StateMachineHttp

folex:


statemachine/api/src/main/scala/fluence/statemachine/api/query/QueryResponse.scala
```diff
@@ -27,18 +29,69 @@ import scodec.bits.ByteVector
...
+      info <- res.downField("info").as[String]
+      height <- res.downField("height").as[Long]
+      value <- res.downField("value").as[Array[Byte]]
+    } yield QueryResponse(height, value, code, info)
+  }
+}
+
+/*
+{
+"jsonrpc": "2.0",
```
DieMyst: is it needed here?

folex: Nope, I will delete that before merge. It's here to simplify review.

PR #844, opened 2019-09-19
Use docker proxy instead of docker.sock

folex:


Makefile
```diff
@@ -45,5 +47,25 @@ endif
...
+push:            push-node push-worker
+# Build containers from existing jars and publish
+push-local:      node-test worker-test; $(MAKE) push
+# Build containers in docker build environment and publish
+push-clean:      node worker; $(MAKE) push
+push-node:       ;docker push $(NODE_IMG)
+push-worker:     ;docker push $(WORK_IMG)
+
+# Build jars
+jars:            ;sbt ";node/assembly ;statemachine-docker/assembly"
```
alari: You can simply do `sbt node/assembly statemachine-docker/assembly` for a sequence of SBT tasks.

tools/deploy/netdata/netdata.yml
```diff
@@ -25,3 +25,15 @@ services:
         limits:
           cpus: '0.2'
           memory: 500M
+
+  sockguard:
+    container_name: sockguard
+    image: buildkite/sockguard
+    command: -filename /proxy/docker.sock -debug -allow-bind $FLUENCE_STORAGE
```
alari: What's the reason to leave `-debug` option here?

Makefile
```diff
@@ -45,5 +47,25 @@ endif
...
+push:            push-node push-worker
+# Build containers from existing jars and publish
+push-local:      node-test worker-test; $(MAKE) push
+# Build containers in docker build environment and publish
+push-clean:      node worker; $(MAKE) push
+push-node:       ;docker push $(NODE_IMG)
+push-worker:     ;docker push $(WORK_IMG)
+
+# Build jars
+jars:            ;sbt "node/assembly statemachine-docker/assembly"
```
alari: I'm not sure if it works correctly with quotes. You can either do `sbt "command and arguments"`, or "sbt command1 command2". See https://www.scala-sbt.org/1.x/docs/Running.html#Batch+mode

PR #847, opened 2019-09-23
Move Virtual machine to Wasmer

### LINKED ISSUE #771: Implement EIC counter for normal execution mode

michaelvoronov:
Resolves #771 #855.
Related to #854.


vm/src/main/scala/fluence/vm/WasmVm.scala
```diff
@@ -80,163 +65,39 @@ trait WasmVm {
...
     conf: ⇒ Config = ConfigFactory.load()
-  ): EitherT[F, ApplyError, WasmVm] =
+  ): EitherT[F, InitializationError, WasmVm] =
     for {
       // reading config
       config ← VmConfig.readT[F](configNamespace, conf)
 
-      _ ← Log.eitherT[F, ApplyError].info("WasmVm: configs read...")
+      _ ← Log.eitherT[F, InitializationError].info("WasmVm: configs read...")
+      vmRunnerInvoker = new FrankAdapter()
```
folex: I think you should delay this action as it side effects via `@nativeLoader`:
```scala
      vmRunnerInvoker <- EitherT.right(Sync[F].delay(new FrankAdapter()))
      initializationResult <-  EitherT.right(Sync[F].delay(vmRunnerInvoker.initialize(inFiles.head, config)))
```

michaelvoronov: Yes, you are right. Missed it.

PR #862, opened 2019-10-04
Preallocate Wasm memory

michaelvoronov:


vm/frank/src/vm/prepare.rs
```diff
@@ -0,0 +1,85 @@
...
+
+use crate::vm::config::Config;
+use crate::vm::errors::InitializationError;
+use parity_wasm::elements::{MemorySection, MemoryType, ResizableLimits};
+
+struct ModulePreparator {
+    module: elements::Module,
+}
+
+impl<'a> ModulePreparator {
```
DieMyst: docs about what this class is for

DieMyst: or do it `private`, if it used only in `prepare_module` method

michaelvoronov: It's already private. In rust it needs to explicitly declare `pub` to make smth public. I think that comment here is excess because it would too obvious:)

DieMyst: yeah, forgot about this :)

Issue #854, opened 2019-09-30
Tracking issue: change execution environment to Wasmer


michaelvoronov:
With some Asmble drawbacks it seems more reasonable to use Wasmer as the execution environment, but currently Wasmer doesn't provide some features of Asmble and previous virtual machine like memory merkalization and static module linkage out of the box. Since that, we decided to integrate it step-by-step. This issue is a beacon issue for this process.

- [x] initial Wasmer support ([PR 847](https://github.com/fluencelabs/fluence/pull/847)):
   - [x] able to work with one module and pass all tests that use one module ([PR 847](https://github.com/fluencelabs/fluence/pull/847))
   - [x] able to pass config from Scala part and return state back ([PR 847](https://github.com/fluencelabs/fluence/pull/847))
   - [x] `expects_eth` support ([PR 847](https://github.com/fluencelabs/fluence/pull/847))
   - [x] Scala part refactored to support Wasmer ([PR 847](https://github.com/fluencelabs/fluence/pull/847))
   - [x] support gas and eic metering ([PR 847](https://github.com/fluencelabs/fluence/pull/847))
   - [x] able to integrate as a jni library ([branch](https://github.com/fluencelabs/wasmer/compare/master...fluencelabs:clif_jni_hardering))
   - [x] optimize logging ([PR 861](https://github.com/fluencelabs/fluence/pull/861))
   - [ ] able to use singlepass compiler backend and pass our tests
   - [x] preallocates all memory on the startup ([PR 862](https://github.com/fluencelabs/fluence/pull/862))
   - [x] able to save global Instance with found func ([PR 863](https://github.com/fluencelabs/fluence/pull/863))
   - [ ] optimize parameters passing and returning result back
- [x] Decouple Frank to a separate project
- [ ] Support of multi modules
- [ ] Memory merkalization

michaelvoronov:
Singlepass compiler is now blocked by https://github.com/wasmerio/wasmer/issues/842

PR #865, opened 2019-10-07
Move Wasmer to singlepass compiler backend

michaelvoronov:
Related to #854.


michaelvoronov:
This PR is depends on signal refactoring in Wasmer and now it isn't so important.

PR #867, opened 2019-10-07
SubscriptionStorage spec

DieMyst:


worker/responder/src/test/scala/fluence/worker/responder/SubscriptionStorageSpec.scala
```diff
@@ -0,0 +1,52 @@
...
+    val tx = "123"
+    val txData = Tx.Data(tx.getBytes())
+    val key = SubscriptionKey.generate("subscriptionId", txData)
+    (storage, key, txData)
+  }
+
+  "subscription storage" should {
+    "return false on adding subscription twice" in {
+      val (storage, key, txData) = createStorageAndSubscription()
+      storage.addSubscription(key, txData).unsafeRunSync() shouldBe true
```
folex: Let's keep whole test in a single IO execution. Running it this way could be slower, and bring unexpected results.

worker/responder/src/test/scala/fluence/worker/responder/SubscriptionStorageSpec.scala
```diff
@@ -0,0 +1,80 @@
...
+      } yield ()).unsafeRunSync()
+    }
+
+    "return stream after adding subscription and stream" in {
+      (for {
+        (storage, key, txData) <- createStorageAndSubscription()
+        add1Result <- storage.addSubscription(key, txData)
+        _ = add1Result shouldBe true
+        subs <- storage.getSubscriptions
+      } yield subs.get(key).isDefined shouldBe true).unsafeRunSync()
```
folex: `isDefined shouldBe true` => `should be defined`

worker/responder/src/test/scala/fluence/worker/responder/SubscriptionStorageSpec.scala
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    "delete subscription on delete operation" in {
+      (for {
+        (storage, key, txData) <- createStorageAndSubscription()
+        add1Result <- storage.addSubscription(key, txData)
+        _ = add1Result shouldBe true
+        _ <- storage.deleteSubscription(key)
+        subs <- storage.getSubscriptions
+      } yield subs.get(key).isEmpty shouldBe true).unsafeRunSync()
```
folex: it is better to write `isEmpty shouldBe true` => `should be empty` because error and intention both will be more clear

PR #901, opened 2020-06-05
Integrate metrics (+ bugfix)

folex:
- Simplify ClientCommand in rust libp2p client (hope to get rid of json commands in future)
- Fix announced service removal on peer disconnection: force removal from DHT
- Add `run_multiple_nodes`
- Fix `MemoryStore::remove`: 
a) Previously removed the whole MultiRecord, now remove only provider, if record isn't expired
b) Remove the whole record if it's expired
- Add delays to `Bootstrapper` behaviour, it was spamming a lot. Still need to make delays exponential, I think
- Integrate metrics, add function to start /metrics prometheus endpoint


client/rust-libp2p/src/main.rs
```diff
@@ -39,16 +41,22 @@ use futures::{
 use libp2p::PeerId;
 use parity_multiaddr::Multiaddr;
 use std::error::Error;
-use std::time::Duration;
+// use std::time::Duration;
```
michaelvoronov, line 44: Maybe delete this?

server/src/behaviour/bootstrapper.rs
```diff
@@ -20,28 +20,28 @@ use libp2p::swarm::NetworkBehaviourEventProcess;
 
 impl NetworkBehaviourEventProcess<BootstrapperEvent> for ServerBehaviour {
     fn inject_event(&mut self, event: BootstrapperEvent) {
+        // TODO: do not reconnect to boostraps all the time, make it stop after a few minutes after node was started
+        //       первые 5 минут реконнектишься, потом перестаешь делать это
```
michaelvoronov, line 24: Cmt

server/src/function/services.rs
```diff
@@ -143,10 +143,12 @@ impl FunctionRouter {
         }
     }
 
-    // Removes all names that resolve to an address containing `resolvee`
+    /// Removes all names that resolve to an address containing `resolvee`
     pub(super) fn remove_halted_names(&mut self, resolvee: &PeerId) {
         use Protocol::*;
 
+        let mut removed = vec![];
```
michaelvoronov, line 150: Can we do this with collect without a mutable variable?

folex: https://github.com/rust-lang/rust/issues/59618 Currently – no. Gotta wait until drain_filter is implemented in std HashMap

PR #910, opened 2020-06-15
Route by hashtag

folex:
1. Local services are now addressed via `#`: `fluence:/peer/QmPeer#identify`
2. `/service` renamed to `/providers`: `fluence:/service/IPFS`
3. Services are now published to DHT by `#`. This means that calling 
```
FunctionCall {
  target: fluence:/#provide
  arguments: { service_id: IPFS },
  reply_to: /peer/QmPeer/client/QmClient/signature/QmSig,
  ...
}
```
Will publish to DHT the following: `#IPFS -> [ /peer/QmPee/client/QmClient/signature/QmSig ]`

4. Same conversion happens when resolving providers through DHT. A request like
```
FunctionCall {
  target: fluence:/providers/IPFS,
  ...
}
```
Will result in searching for `#IPFS` on DHT, and then forwarding call, using `#IPFS`:
```
target: fluence:/peer/QmPeer/client/QmClient#IPFS
```

### DIFF
--- a/client/fluence-ipfs/src/lib.rs
+++ b/client/fluence-ipfs/src/lib.rs
@@ -42,12 +42,12 @@ use uuid::Uuid;
 
 const IPFS_SERVICE_ID: &str = "IPFS.multiaddr";
 #[rustfmt::skip]
-static IPFS_SERVICE: Lazy<Protocol> = Lazy::new(|| Protocol::Service(IPFS_SERVICE_ID.to_string()));
+static IPFS_SERVICE: Lazy<Protocol> = Lazy::new(|| Protocol::Providers(IPFS_SERVICE_ID.to_string()));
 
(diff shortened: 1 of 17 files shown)


client/fluence-ipfs/src/lib.rs
```diff
@@ -41,15 +41,16 @@ use std::time::Duration;
 use uuid::Uuid;
 
 const IPFS_SERVICE_ID: &str = "IPFS.multiaddr";
-static IPFS_SERVICE: Lazy<Protocol> = Lazy::new(|| Protocol::Service(IPFS_SERVICE_ID.to_string()));
+static IPFS_SERVICE: Lazy<Protocol> =
```
michaelvoronov, line 44: What about `#[rustfmt:skip]`?

PR #916, opened 2020-06-25
Integrate FluenceFaaS

folex:
- Add `get_interface` builtin
- Forward calls to FluenceFaaS if there's a matching `call.module` with a matching `call.function`. Currently exact match is required
- `FluenceConfig` now deserialized straight from TOML, cli arguments override them if passed
- Started refactoring error handling: return Result instead of ad-hoc `send_error_on_call`. Could be better.


faas-api/src/call.rs
```diff
@@ -34,10 +35,28 @@ pub struct FunctionCall {
...
+    /// Set target to a call, and return the call
     pub fn with_target(mut self, target: Address) -> Self {
         self.target = Some(target);
         self
     }
+
+    /// Build a "reply" call from given arguments
+    pub fn reply<O>(target: Address, sender: Address, arguments: serde_json::Value, name: O) -> Self
+    where
+        O: Into<Option<String>>,
```
michaelvoronov, line 47: Looks like it should be `Option<Into<String>>`

folex: No, the idea is to allow both `None` and `"Hello".into()`. 

It is possible to make it
```
O: Into<Option<S>>,
S: Into<String>
```

But then compiler can't derive types on its own, so you need to specify them, and it becomes cumbersome.

michaelvoronov: Hmm, with `Option<Into<String>>` it will be possible to write`Some("Hello")`. But both variants look good for me.

.circleci/config.yml
```diff
@@ -12,13 +12,14 @@ jobs:
...
-          rustup toolchain install stable
+          rustup toolchain install nightly
+          rustup default nightly
           rustup component add rustfmt
           rustup component add clippy
-          cargo fmt --all -- --check --color always
-          cargo build -v --all-features
-          cargo test -v --all-features
-          cargo clippy -v
+          cargo +nightly fmt --all -- --check --color always
```
michaelvoronov, line 19: Do we need `+nightly` having `rustup default nightly`?

server/src/function/execution.rs
```diff
@@ -95,124 +121,85 @@ impl FunctionRouter {
...
-            }
-            Some(pk) => {
-                #[rustfmt::skip]
-                log::error!("unsupported public key: expected ed25519, got {:?} {:?}", pk, call);
-                self.send_error_on_call(call, "unsupported public key: expected ed25519".into());
-                return;
-            }
+        let public_key = peer_id
+            .as_public_key()
+            .ok_or(call.clone().error(MissingPublicKey))
```
michaelvoronov, line 171: ok_or_else

server/src/function/execution.rs
```diff
@@ -95,124 +121,85 @@ impl FunctionRouter {
...
-        // Check reply_to is defined
-        let reply_to = if let Some(reply_to) = call.reply_to.clone() {
-            reply_to
-        } else {
-            log::error!("reply_to undefined on {:?}", call);
-            self.send_error_on_call(call, "reply_to is undefined".into());
-            return;
-        };
+        let reply_to = call.reply_to.clone();
+        let reply_to = reply_to.ok_or(call.clone().error(MissingReplyTo))?;
```
michaelvoronov, line 166: ok_or_else

server/src/function/execution.rs
```diff
@@ -262,4 +235,34 @@ impl FunctionRouter {
...
+    /// Send reply on a given call with given msg_id and data in arguments
+    /// If data or msg_id is empty, it's not included in the arguments
+    fn reply_with<T: Serialize>(
+        &mut self,
+        call: FunctionCall,
+        msg_id: Option<String>,
+        data: (&str, T),
+    ) -> Result<(), CallError<'static>> {
+        let reply_to = call.reply_to.clone();
+        let reply_to = reply_to.ok_or(call.clone().error(MissingReplyTo))?;
```
michaelvoronov, line 248: Looks like redundant clone.

server/src/function/execution.rs
```diff
@@ -42,41 +51,58 @@ impl FunctionRouter {
...
+                error: format!("expected array of interface values: got {:?}", other),
+            })),
+        }?;
+
+        let result = self
+            .faas
+            .call_module(&module, &function, arguments)
+            .map_err(|e| call.clone().error(e))?;
+
+        // Handle empty result manually
```
michaelvoronov, line 97: I think it would be good to add why we are handling empty result manually.

server/src/function/services.rs
```diff
@@ -17,31 +17,60 @@
...
     // ###
 
     /// Execute call locally: on builtin service or forward to provided name
     /// `ttl` – time to live (akin to ICMP ttl), if `0`, execute and drop, don't forward  
-    pub(super) fn pass_to_local_service(&mut self, module: &str, call: FunctionCall) {
+    pub(super) fn execute_locally<'a>(
+        &mut self,
+        module: &'a str,
+        call: FunctionCall,
+    ) -> CallResult<'a, ()> {
```
michaelvoronov, line 38: Is it neccessary CallResult to have the same lifetime as a module name?

server/src/function/services.rs
```diff
@@ -17,31 +17,60 @@
...
+
+    /// Find a matching module with a matching function, and return their names
+    fn find_in_faas(
+        &mut self,
+        module: &str,
+        function: Option<&str>,
+    ) -> Result<Option<(String, String)>, CallErrorKind<'static>> {
+        let interface = self.faas.get_interface();
+        let module = ok_get!(interface.modules.iter().find(|m| m.name == module));
+        let function = function.ok_or(MissingFunctionName {
```
michaelvoronov, line 64: ok_or_else

server/src/function/services.rs
```diff
@@ -17,31 +17,60 @@
...
+            module: module.name.to_string(),
+        })?;
+        let function = module.functions.iter().find(|f| f.name == function).ok_or(
```
michaelvoronov, line 67: What do you think about returning `HashMap` from Faas instead of `Vec`? I think it wouldn't break the exisiting code, but will provide fast search. I can add it if it make sense in your opinion.

michaelvoronov: ok_or_else

server/src/function/services.rs
folex: Yes, `module` is returned within error

server/src/function/services.rs
folex: I think that's reasonable, let's return HashMap

PR #920, opened 2020-07-03
filter certs with one trust

DieMyst:


trust-graph/src/trust_graph.rs
```diff
@@ -276,6 +279,7 @@ mod tests {
     use crate::key_pair::KeyPair;
     use crate::misc::current_time;
     use failure::_core::time::Duration;
+    use std::str::FromStr;
```
michaelvoronov, line 282: If `FromStr` is needed only in `test_chain_from_root_to_another_root`, it is better to move this there.

PR #922, opened 2020-07-06
Update to libp2p 0.20 and fix certificate chain filtering 

folex:


trust-graph/src/trust.rs
```diff
@@ -93,18 +93,24 @@ impl Trust {
             return Err("Trust in chain is expired.".to_string());
         }
 
-        let msg = Self::metadata_bytes(&trust.issued_for, trust.expires_at, trust.issued_at);
+        let msg: &[u8] =
+            &Self::metadata_bytes(&trust.issued_for, trust.expires_at, trust.issued_at);
+        // println!(
```
michaelvoronov, line 98: Do we need these logs?

PR #923, opened 2020-07-14
Create services from context

folex:


faas-api/src/call.rs
```diff
@@ -27,11 +27,33 @@ pub struct FunctionCall {
...
     pub arguments: serde_json::Value, //TODO: make it Option<String>?
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub name: Option<String>,
     pub sender: Address,
+    /// List of modules to load when executing function
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub context: Vec<String>,
+}
+
+impl Default for FunctionCall {
```
michaelvoronov, line 43: It seems that just `#[derive(Debug, ...)]` is enough.

michaelvoronov:
Looks good, I imagined working with the pool and wakers like here.

faas-api/src/call.rs
folex: Custom default is needed to make `arguments` an empty object (`{}`) by default. This is because serde doesn't treat `null` as empty, and thus doesn't invoke `#[serde(default)]` strategy. The alternative is to implement custom de/serialize functions for `arguments` field, not sure if it's worth it

PR #928, opened 2020-07-22
Change "provide" API

folex:


folex:
> It seems that all random methods are needed only for tests. What do you think about introducing a new trait `Random` or several impl blocks for `Address`, `Protocol`, and `FunctionCall` where one would be the as current and the second would contain only `random` function?

I believe `random` methods are used across several crates, so this trait should be implemented almost in the same place, except maybe in a different file. Is that what you're suggesting?

michaelvoronov:
Yes, they should probably be in the same place.

PR #936, opened 2020-08-13
Introduce blueprints

folex:


server/src/app_service/blueprint.rs
```diff
@@ -0,0 +1,34 @@
...
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct Blueprint {
+    pub name: String,
+    pub id: String,
+    pub dependencies: Vec<String>,
+}
+
+impl Blueprint {
+    #[allow(dead_code)]
+    pub fn new<S: Into<String>>(name: S, id: String, dependencies: Vec<String>) -> Self {
```
michaelvoronov, line 27: Maybe make id also `Into<String>`?

server/src/app_service/files.rs
```diff
@@ -0,0 +1,46 @@
...
+}
+
+/// Returns true if file is named like a blueprint would be
+pub(super) fn is_blueprint(name: &str) -> bool {
+    name.ends_with("_blueprint.toml")
+}
+
+/// If name ends with .wasm, returns that name without .wasm. Returns None otherwise
+pub(super) fn extract_module_name(name: &str) -> Option<String> {
+    name.strip_suffix(".wasm").map(|s| s.to_string())
```
michaelvoronov, line 45: Suggest to use `Path::file_name` otherwise.

server/src/app_service/blueprint.rs
folex: actually I think there's no need for generics here at all.

PR #940, opened 2020-08-20
Client blueprint, service incapsulation

DieMyst:


client/fluence-js/src/fluenceClient.ts
```diff
@@ -303,6 +321,28 @@ export class FluenceClient {
         }
     }
 
+    async getAvailableBlueprints(peerId?: string): Promise<Blueprint[]> {
+        let resp;
+        if (peerId) {
+            resp = await this.sendCallWaitResponse(createPeerAddress(peerId), {}, "get_available_blueprints");
+        } else {
+            resp = await this.callPeer("get_available_blueprints", {}, undefined, peerId);
```
folex, line 329: peerId isn't defined here, so I think you could pass `undefined` to make intention clear
```suggestion
            resp = await this.callPeer("get_available_blueprints", {}, undefined, undefined);
```

folex (approved):

PR #941, opened 2020-08-22
Integrate call parameters

michaelvoronov:
Adds support of call parameters accessible by Wasm side.

Linked PRs:
https://github.com/fluencelabs/fce/pull/21
https://github.com/fluencelabs/rust-sdk/pull/3

### DIFF
--- a/faas-api/src/call.rs
+++ b/faas-api/src/call.rs
@@ -15,6 +15,7 @@
  */
 
 use crate::Address;
+use fluence_app_service::CallParameters;
 use serde::{Deserialize, Serialize};
 use uuid::Uuid;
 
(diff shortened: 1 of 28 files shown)


faas-api/src/call.rs
```diff
@@ -32,6 +33,7 @@ pub struct FunctionCall {
         skip_serializing_if = "serde_json::Value::is_null"
     )]
     pub arguments: serde_json::Value, //TODO: make it Option<String>?
+    pub call_parameters: Option<CallParameters>,
```
folex, line 36: I think that application id should be a part of the address, and call id is already passed via `uuid` field. I think that maybe call parameters should be derived from `call.arguments`

folex (approved):

PR #947, opened 2020-09-18
Fire client

DieMyst:


folex (approved):
maybe rename fireClient => callClient?

ultimately I'd also move address creation to a separate set of methods, so you first create needed addresses, and then call `callProvider`/`fireClient` with much fewer arguments

PR #949, opened 2020-09-24
Client particle

DieMyst:


client/fluence-js/src/particle.ts
```diff
@@ -0,0 +1,87 @@
...
+    // sign upper fields
+    signature: string,
+    data: object,
+    action: "Particle"
+}
+
+/**
+ * Copies a particle and stringify it.
+ */
+export function particleToString(call: Particle): string {
```
folex, line 36: I'd call that `stringifyParticle`, so it's clearer that it uses JSON.stringify

client/fluence-js/src/particle.ts
```diff
@@ -0,0 +1,87 @@
...
+export interface Particle {
+    id: string,
+    init_peer_id: string,
+    timestamp: number,
+    ttl: number,
+    script: string,
+    // sign upper fields
+    signature: string,
+    data: object,
+    action: "Particle"
```
folex, line 30: I'm not sure particle should contain `action`. `action` comes from `ProtocolMessage`, it's on the different level from particle

client/fluence-js/src/particle.ts
```diff
@@ -0,0 +1,87 @@
...
+        signature: json.signature,
+        data: json.data,
+        action: "Particle"
+    }
+}
+
+/**
+ * Sign a particle with a private key from peerId.
+ */
+export async function signParticle(peerId: PeerId,
```
folex, line 61: I'd separate that into two functions: `canonicalBytes(particle: Particle): Buffer` and `sign(privKey, Buffer): promise<string>`. This way, you can use `canonicalBytes` in `checkSignature` that you may need later.

client/fluence-js/src/preparedParticle.ts
```diff
@@ -14,22 +14,24 @@
...
-    return true;
+    script: string;
+    client: FluenceClient;
 
+    constructor(script: string, client: FluenceClient) {
+        this.script = script;
+        this.client = client;
+    }
+
+    async send(data: object, ttl?: number): Promise<Particle> {
```
folex, line 33: Why `PreparedParticle` knows how to send itself on the network? `ActiveObject` or something? Seems like mixed responsibilities.

I think `PreparedParticle` should only know how to convert itself into `Particle`, and the resulting particle should be sent through a single mechanism for sending all particles.

client/fluence-js/src/subscriptions.ts
```diff
@@ -14,29 +14,45 @@
  * limitations under the License.
  */
 
-import {FunctionCall} from "./functionCall";
-import {Address} from "./address";
+import {Particle} from "./particle";
 
 export class Subscriptions {
```
folex, line 19: I think all subscription logic will be replaced by `Map<string, Stepper>`, so not sure if `subscriptions.ts` is needed

client/fluence-js/src/trust/trust_graph.ts
```diff
@@ -50,9 +52,10 @@ export class TrustGraph {
 
     // Get certificates that stores in Kademlia neighbourhood by `peerId` key.
     async getCertificates(peerId: string): Promise<Certificate[]> {
-        let resp = await this.client.callPeer("certificates", {
+        let resp: any = {};
+        /*let resp = await this.client.callPeer("certificates", {
```
folex, line 56: Maybe remove that?

client/fluence-js/src/trust/trust_graph.ts
```diff
@@ -34,10 +35,11 @@ export class TrustGraph {
             certsStr.push(await certificateToString(cert));
         }
 
-        let response = await this.client.callPeer("add_certificates", {
+        /*let response = await this.client.callPeer("add_certificates", {
```
folex, line 38: Maybe remove that?

folex (changes requested):

folex (approved):

PR #950, opened 2020-09-29
AquamarineVM integration: semi-working "create" builtin service

folex:


crates/config-utils/src/config.rs
```diff
@@ -0,0 +1,54 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+use libp2p_core::{identity::ed25519::Keypair, PeerId, PublicKey};
+use std::path::{Path, PathBuf};
+
+pub fn abs_path(path: PathBuf) -> PathBuf {
```
michaelvoronov, line 20: Suggest to rename it to `to_abs_path`

crates/test-utils/src/misc.rs
```diff
@@ -327,8 +349,20 @@ pub fn remove_dir(dir: &PathBuf) {
     std::fs::remove_dir_all(&dir).unwrap_or_else(|_| panic!("remove dir {:?}", dir))
 }
 
-pub fn put_aquamarine(mut tmp: PathBuf) {
-    tmp.push("aquamarine.wasm");
-    std::fs::write(&tmp, AQUAMARINE)
+pub fn abs_path(path: PathBuf) -> PathBuf {
```
michaelvoronov, line 352: It is already the same function in `crates/config-utils/src/config.rs`

crates/test-utils/src/misc.rs
```diff
@@ -327,8 +349,20 @@ pub fn remove_dir(dir: &PathBuf) {
...
+    match std::env::current_dir().ok() {
+        Some(c) => c.join(path),
+        None => path,
+    }
+}
+
+pub fn put_aquamarine(tmp: PathBuf, file_name: Option<String>) {
+    let file_name = file_name.unwrap_or("aquamarine.wasm".to_string());
```
michaelvoronov, line 360: Lets use const for that, moreover we can reexport this name from AquamarineVM, because it also uses such const.

particle-actors/src/plumber.rs
```diff
@@ -212,19 +161,26 @@ impl Plumber {
         }
     }
 
-    fn create_actor(config: ActorConfig, particle: Particle) -> Fut {
-        Box::pin(task::spawn_blocking(move || Actor::new(config, particle)))
+    fn create_actor(config: ActorConfig, particle: Particle, services: Fabric) -> Fut {
+        Box::pin(task::spawn_blocking(move || {
```
michaelvoronov, line 165: Maybe you will found more readable to use `.boxed()` instead of `Box::pin`

particle-services/src/files.rs
```diff
@@ -0,0 +1,62 @@
+/*
```
michaelvoronov, line 1: Maybe move these functions to the utils crate?

particle-services/src/files.rs
```diff
@@ -0,0 +1,62 @@
...
+
+use crate::blueprint::Blueprint;
+use std::path::{Path, PathBuf};
+
+/// Calculates filename of the config for a wasm module
+pub(super) fn module_config_name<S: AsRef<str>>(module: S) -> String {
+    format!("{}_config.toml", module.as_ref())
+}
+
+/// Calculates filename of wasm module
```
michaelvoronov, line 26: A little bit confusing with comment and name of this function

particle-services/src/files.rs
```diff
@@ -0,0 +1,62 @@
...
+pub(super) fn module_file_name<S: AsRef<str>>(module: S) -> String {
+    format!("{}.wasm", module.as_ref())
+}
+
+/// Calculates filename of the blueprint
+pub(super) fn blueprint_file_name(blueprint: &Blueprint) -> String {
+    blueprint_fname(blueprint.id.as_str())
+}
+
+pub(super) fn blueprint_fname(id: &str) -> String {
```
michaelvoronov, line 36: `module_file_name` is generic, but this receives `&str`, suggest to be consistent.

michaelvoronov (approved):

crates/test-utils/src/misc.rs
folex: I'll move that to const, but I'd prefer to avoid depending on aquamarine-vm from test-utils

particle-services/src/files.rs
folex: Nah, it's only useful to particle-services

particle-services/src/files.rs
folex: Why? It calculates a name of the wasm module file, given a name of the module. For example, it's `ipfs` in config, and `ipfs.wasm` on filesystem. That function then encapsulates knowledge about that transformation.

folex: I'd put that as a comment to clarify:
```rust
/// Calculates the name of a wasm module file, given a name of the module.
/// For example, it's ipfs in config, and ipfs.wasm on filesystem.
/// That function then encapsulates the knowledge about that transformation.
pub(super) fn module_file_name<S: AsRef<str>>(module: S) -> String {
    format!("{}.wasm", module.as_ref())
}
```

particle-services/src/files.rs
folex: I guess that happened because before `module_file_name` was used with different types of arguments, and `blueprint_fname` only with `&str`.

folex: I'll shrink `module_file_name` to receive only `&str`

PR #951, opened 2020-09-30
Stepper wasm integration

DieMyst:


client/fluence-js/src/call_service.ts
```diff
@@ -0,0 +1,44 @@
...
+registerService(logger);
+
+export function call_service(service_id: string, fn_name: string, args: string): string {
+    let service = services.get(service_id)
+    if (service) {
+        return service.call(fn_name, args)
+    } else {
+        return `Error. There is no service ${service_id}`
+    }
+}
```
folex, line 44: Hm, is it ok to have global state in the middle of a library? What would happen if someone would instantiate several `fluence-js` clients, will they share that map?

client/fluence-js/src/fluenceClient.ts
```diff
@@ -20,6 +20,9 @@ import * as PeerId from "peer-id";
 import Multiaddr from "multiaddr"
 import {FluenceConnection} from "./fluenceConnection";
 import {Subscriptions} from "./subscriptions";
+import * as stepper from "../stepper";
+
+const WASM = stepper.loadWasm();
```
folex, line 25: Global state

folex (commented):
I'm not sure how to deal with the global state. I understand it comes from the fact that wasm module's access to "host functions" like `call_service` is global, but this seems like an undesired effect. 

Having a page-global state in the library is gonna complicate things, forbid running several clients at once, and I'm sure there will be more serious consequences.

client/fluence-js/src/call_service.ts
```diff
@@ -1,4 +1,11 @@
-class Service {
+import {getService, registerService} from "./globalState";
+
+interface CallServiceResult {
+    ret_code: number,
```
folex, line 4: Maybe make this a `enum`? So we can agree on error codes meaning

client/fluence-js/src/globalState.ts
```diff
@@ -0,0 +1,12 @@
...
+// TODO put state with wasm file in each created FluenceClient
+let services: Map<string, Service> = new Map();
+
+export function registerService(service: Service) {
+    services.set(service.serviceId, service)
+}
+
+export function getService(serviceId: string): Service | undefined {
+    return services.get(serviceId)
+}
```
folex, line 12: \n

client/fluence-js/src/stepperOutcome.ts
```diff
@@ -1,4 +1,5 @@
 interface StepperOutcome {
+    ret_code: number,
```
folex, line 2: enum?

folex (approved):

PR #952, opened 2020-10-01
Integrate stepper with app services and builtin services

folex:


crates/test-utils/src/misc.rs
```diff
@@ -47,6 +46,8 @@ pub static TIMEOUT: Duration = Duration::from_secs(15);
 pub static SHORT_TIMEOUT: Duration = Duration::from_millis(100);
 pub static KAD_TIMEOUT: Duration = Duration::from_millis(500);
 
+const AQUAMARINE: &'static str = "aquamarine.wasm";
```
michaelvoronov, line 49: Seems like `'static` isn't needed here

particle-closures/src/host_closures.rs
```diff
@@ -0,0 +1,114 @@
...
+use crate::BuiltinServicesApi;
+
+use particle_actors::{vec1::Vec1, HostImportDescriptor};
+use particle_services::{Args, IType, IValue};
+
+use serde_json::json;
+use std::sync::Arc;
+
+type ClosureDescriptor = Arc<dyn Fn() -> HostImportDescriptor + Send + Sync + 'static>;
+type Closure = Arc<dyn Fn(Args) -> Option<IValue> + Send + Sync + 'static>;
```
michaelvoronov, line 26: Lets rename it to `HostClosure`?

particle-closures/src/host_closures.rs
```diff
@@ -0,0 +1,114 @@
...
+    pub call_service: Closure,
+    pub builtin: Closure,
+}
+
+impl HostClosures {
+    pub fn descriptor(self) -> ClosureDescriptor {
+        Arc::new(move || {
+            let this = self.clone();
+            HostImportDescriptor {
+                host_exported_func: Box::new(move |_, args| this.clone().route(args)),
```
michaelvoronov, line 40: Why clone is needed here?

particle-closures/src/host_closures.rs
```diff
@@ -0,0 +1,114 @@
...
+                argument_types: vec![IType::String, IType::String, IType::String],
+                output_type: Some(IType::Record(0)),
+                error_handler: None,
+            }
+        })
+    }
+
+    fn route(self, args: Vec<IValue>) -> Option<IValue> {
```
michaelvoronov, line 48: Maybe make self passed by ref? In this case cloning above isn't needed at all.

particle-closures/src/host_closures.rs
```diff
@@ -0,0 +1,114 @@
...
+        // route
+        match args.service_id.as_str() {
+            "create" => (self.create_service)(args),
+            s if BuiltinServicesApi::is_builtin(&s) => (self.builtin)(args),
+            _ => (self.call_service)(args),
+        }
+    }
+}
+
+pub fn as_record(v: std::result::Result<IValue, IValue>) -> Option<IValue> {
```
michaelvoronov, line 66: Are there any reasons to duplicate this and other functions below in `vm.rs`?

particle-closures/src/mailbox.rs
```diff
@@ -0,0 +1,139 @@
...
+
+use futures::{channel::mpsc, StreamExt};
+use libp2p::kad::record;
+use std::{
+    collections::HashSet,
+    sync::{mpsc as std_mpsc, Arc},
+    task::{Context, Poll},
+};
+
+pub(super) type Closure = Arc<dyn Fn(Args) -> Option<IValue> + Send + Sync + 'static>;
```
michaelvoronov, line 32: This type's already defined in `host_closures.rs`.

particle-closures/src/mailbox.rs
```diff
@@ -0,0 +1,139 @@
...
+impl Mailbox {
+    pub fn resolve_complete(
+        &mut self,
+        key: record::Key,
+        value: Result<HashSet<Vec<u8>>, ResolveErrorKind>,
+    ) {
+        for cmd in self.waiting.remove(&Key::DHT(key.clone())) {
+            let result = BuiltinCommandResult::DHTResolved(
+                key.clone(),
+                value.clone().map(|v| v.into_iter().collect()),
```
michaelvoronov, line 122: Here you are clonning and converting it to iter on the each step of cycle. Seems like it is possible to make it out of the cycle.

particle-dht/src/resolve.rs
```diff
@@ -14,26 +14,81 @@
...
+
 impl ParticleDHT {
+    pub fn resolve(&mut self, key: Key) {
+        self.kademlia.get_record(&key, Quorum::Majority);
+    }
+
+    pub(super) fn resolved(&mut self, result: GetRecordResult) {
+        match Self::recover_get(result) {
+            Ok(records) => {
+                let key = records.iter().next().unwrap().record.key.clone();
```
michaelvoronov, line 36: Suggest to leave a comment why unwrap is safe here

particle-services/src/app_services.rs
```diff
@@ -0,0 +1,103 @@
...
+use crate::error::ServiceError;
+use crate::error::ServiceError::MissingBlueprintId;
+use crate::vm::{as_record, create_vm};
+
+use fluence_app_service::{AppService, IValue};
+
+use parking_lot::{Mutex, RwLock};
+use std::{collections::HashMap, sync::Arc};
+
+type Closure = Arc<dyn Fn(Args) -> Option<IValue> + Send + Sync + 'static>;
```
michaelvoronov, line 28: This type's already defined in `host_closures.rs`.

particle-services/src/app_services.rs
```diff
@@ -0,0 +1,103 @@
...
+
+        Arc::new(move |args| {
+            let call = || -> Result<Vec<IValue>, ServiceError> {
+                let services = services.read();
+                let vm = services
+                    .get(&args.service_id)
+                    .ok_or(ServiceError::NoSuchInstance(args.service_id))?;
+                let result = vm
+                    .lock()
+                    .call("facade".to_string(), args.fname, args.args, <_>::default())
```
michaelvoronov, line 88: What does "facade" mean here? Or it is a stub?

michaelvoronov (approved):

particle-services/src/app_services.rs
```diff
@@ -0,0 +1,103 @@
...
+impl ParticleAppServices {
+    pub fn new(config: ServicesConfig) -> Self {
+        Self {
+            config,
+            services: <_>::default(),
+        }
+    }
+
+    pub fn create_service(&self) -> Closure {
+        let services = self.services.clone();
```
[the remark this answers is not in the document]
folex: It's an `Arc<RwLock<HashMap<String, VM>>>`, it's meant to be cloned and moved into closure:
```
            let result = match make_vm() {
                Ok(vm) => {
                    let vm = Arc::new(Mutex::new(vm));
                    services.write().insert(service_id.clone(), vm);
                    Ok(IValue::String(service_id))
                }
```

particle-server/src/config/fluence_config.rs
```diff
@@ -97,12 +98,15 @@ pub struct ServerConfig {
     #[serde(default = "default_services_basedir")]
     pub services_base_dir: PathBuf,
 
-    #[serde(default = "Vec::new")]
-    pub services_envs: Vec<String>,
+    #[serde(default)]
+    pub services_envs: HashMap<Vec<u8>, Vec<u8>>,
```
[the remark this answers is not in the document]
folex: No, I think it's too minor of a repetition to be DRYed via complex dependencies

particle-closures/src/host_closures.rs
folex: At this point it's just a closure, I think. It's the usage that makes this `Closure` a `HostClosure`

particle-closures/src/host_closures.rs
folex: Because `this` gets cloned from closure on every call. That makes it `Fn`, without clone it would be `FnOnce`

particle-closures/src/host_closures.rs
folex: You can't pass it by ref, it is destructed at the end:

```
        match args.service_id.as_str() {
            "create" => (self.create_service)(args),
            s if BuiltinServicesApi::is_builtin(&s) => (self.builtin)(args),
            _ => (self.call_service)(args),
        }
```

It is possible to pass by ref, and then clone only a single Arc at the very end, is that what you mean?

```
    fn route(&self, args: Vec<IValue>) -> Option<IValue> {
        ...
        match args.service_id.as_str() {
            "create" => (self.create_service.clone())(args),
            s if BuiltinServicesApi::is_builtin(&s) => (self.builtin.clone())(args),
            _ => (self.call_service.clone())(args),
        }
```

particle-closures/src/host_closures.rs
folex: No reason. I'll move them to a separate crate in the next PR

particle-closures/src/mailbox.rs
folex: I'll move them to a separate crate in the next PR

particle-closures/src/mailbox.rs
folex: How is it possible? I have a `Result<HashSet<_>, ResolveErrorKind>`, and need to send that to several channels. I think it is only possible via cloning on each iteration. Do you see any other way around that?

particle-services/src/app_services.rs
folex: It's a stub. In discussions, we agreed that services shouldn't be addressable by module name, only by functions name, because only facade module should be callable

particle-services/src/app_services.rs
```diff
@@ -0,0 +1,103 @@
...
+            let make_vm = || {
+                let blueprint_id = args
+                    .get("blueprint_id")
+                    .and_then(|v| v.as_str())
+                    .ok_or(MissingBlueprintId)?
+                    .to_string();
+
+                let user_id = args
+                    .get("user_id")
+                    .and_then(|v| Some(v.as_str()?.to_string()));
```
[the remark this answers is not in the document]
folex: Unfortunately, you can't do that because map expects you to return String, and `?` expects you to return Option

particle-services/src/app_services.rs
michaelvoronov: Oy, I think we should integrate this to AppService instead. But it is a story for nxt PR's.

particle-closures/src/mailbox.rs
michaelvoronov: Sorry, I wasn't clear, I meant

```rust
        let dht_value = value.map(|v| v.into_iter().collect());
        for cmd in self.waiting.remove(&Key::DHT(key.clone())) {
            let result = BuiltinCommandResult::DHTResolved(
                key.clone(),
                dht_value.clone(),
            );
            cmd.outlet.send(result.clone()).expect("resolve_complete")
        }
```

In this approach iterator is created only once, and then cloned.

particle-services/src/app_services.rs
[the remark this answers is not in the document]
michaelvoronov: Hmm, yes, what about?

```rust
                let user_id = args
                    .get("user_id")
                    .and_then(|v| v.as_str())
                    .map(|v| v.to_string());
```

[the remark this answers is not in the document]
folex: Well, that is longer and bigger by one closure. It is possible to write `.and_then(|v| v.as_str()?.to_string().into());`, if `Some(..)` is what you dislike about that code

particle-closures/src/host_closures.rs
folex: I see your point now, I'll change it to be &self

particle-services/src/app_services.rs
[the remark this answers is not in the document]
folex: I see your point, there's an additional allocation in my approach: `Some(v)` needs to write tag next to `v`, so it needs to allocate it

particle-services/src/app_services.rs
folex: Yep, I think so, too

particle-services/src/app_services.rs
```diff
@@ -0,0 +1,103 @@
...
+                    .lock()
+                    .call("facade".to_string(), args.fname, args.args, <_>::default())
+                    .map_err(ServiceError::Engine)?;
+
+                Ok(result)
+            };
+
+            let result = match call() {
+                // AppService always returns a single element
+                Ok(result) => Ok(result.into_iter().next().expect("must be defined")),
```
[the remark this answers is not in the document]
folex: I'll fix that in the next PR

PR #953, opened 2020-10-06
Implement service creation 

folex:


crates/ivalue-utils/src/lib.rs
```diff
@@ -0,0 +1,119 @@
...
+)]
+
+pub use wasmer_wit::types::InterfaceType as IType;
+pub use wasmer_wit::values::InterfaceValue as IValue;
+pub use wasmer_wit::vec1;
+
+use serde_json::{json, Value};
+use vec1::Vec1;
+
+pub fn into_str(v: &IValue) -> Option<&str> {
```
michaelvoronov, line 35: `as_str` is more convenient with Rust style guide. Prefix `into` says that value should be moved.

crates/ivalue-utils/src/lib.rs
```diff
@@ -0,0 +1,119 @@
...
+pub fn into_string(v: IValue) -> Option<String> {
+    if let IValue::String(s) = v {
+        Some(s)
+    } else {
+        None
+    }
+}
+
+/// Converts result of call_service into `IValue::Record`
+pub fn as_record_opt(v: std::result::Result<Option<Value>, Value>) -> Option<IValue> {
```
michaelvoronov, line 52: `into_record_opt`

crates/ivalue-utils/src/lib.rs
```diff
@@ -0,0 +1,119 @@
...
+    match v {
+        Ok(None) => unit(),
+        Ok(Some(v)) => ok(v),
+        Err(e) => error(e),
+    }
+}
+
+/// Converts result of call_service into `IValue::Record`
+pub fn as_record(v: std::result::Result<Value, Value>) -> Option<IValue> {
```
michaelvoronov, line 61: `into_record`

crates/json-utils/src/lib.rs
```diff
@@ -0,0 +1,42 @@
...
+    unused_mut,
+    unused_variables,
+    unused_unsafe,
+    unreachable_patterns
+)]
+
+use serde_json::Value;
+use std::fmt::Debug;
+
+pub fn into_string(v: Value) -> Option<String> {
```
michaelvoronov, line 31: Suggest to call this Value as `JValue` everywhere for easy understanding of code.

michaelvoronov (approved):

crates/ivalue-utils/src/lib.rs
folex: ok, I'll rename it in the next PR if you don't mind

crates/ivalue-utils/src/lib.rs
folex: ok, I'll rename it in the next PR if you don't mind

crates/json-utils/src/lib.rs
folex: ok, I'll rename it in the next PR if you don't mind

folex: ok, I'll rename it in the next PR if you don't mind

PR #954, opened 2020-10-08
Implement add_provider & get_providers

folex:
It still requires `get_neighbourhood` to be really useful


particle-providers/src/providers.rs
```diff
@@ -0,0 +1,154 @@
...
+}
+
+impl ProviderRepository {
+    /// Creates a closure that
+    /// takes `key` as a parameter, and returns an array of `Provider` for that key
+    pub fn get_providers(&self) -> Closure {
+        let providers = self.providers.clone();
+        closure_opt(move |mut args| {
+            let key: String = Args::next("key", &mut args)?;
+            println!("get_providers {:?}", key);
```
michaelvoronov, line 84: Should it print to logs every time? What about log crate instead?

particle-providers/src/providers.rs
```diff
@@ -0,0 +1,154 @@
...
+        let providers = self.providers.clone();
+        closure_opt(move |mut args| {
+            let key: String = Args::next("key", &mut args)?;
+            let provider: Provider = Args::next("provider", &mut args)?;
+
+            println!("add_providers {:?} {:?}", key, provider);
+
+            let providers_r = read_map(&providers)?;
+            let empty = providers_r.get(&key).is_none();
+            drop(providers_r);
```
michaelvoronov, line 107: Maybe leave a comment why drop here is needed? Another approach would be
```rust
let empty = {
    let providers_r = read_map(&providers)?;
    providers_r.get(&key).is_none()
};
```

particle-providers/src/providers.rs
```diff
@@ -0,0 +1,154 @@
...
+    providers.read().map_err(|_| Poisoned)
+}
+
+fn write_map(
+    providers: &ProviderMap,
+) -> Result<RwLockWriteGuard<'_, ProviderMapInner>, ProviderError<'static>> {
+    providers.write().map_err(|_| Poisoned)
+}
+
+// fn map_get<'a, 'b>(
```
michaelvoronov, line 135: Is it needed?

particle-providers/src/providers.rs
```diff
@@ -0,0 +1,154 @@
...
+    }
+}
+
+fn read_map<'a>(
+    providers: &'a ProviderMap,
+) -> Result<RwLockReadGuard<'a, ProviderMapInner>, ProviderError<'static>> {
+    providers.read().map_err(|_| Poisoned)
+}
+
+fn write_map(
```
michaelvoronov, line 129: Suggest to rename it to smth like `(get_)map_mut_ref`

particle-providers/src/providers.rs
```diff
@@ -0,0 +1,154 @@
...
+            let providers = ok_get!(providers.get(&key));
+            let mut providers = write_set(&providers, &key)?;
+            providers.insert(provider);
+
+            Ok(None)
+        })
+    }
+}
+
+fn read_map<'a>(
```
michaelvoronov, line 123: Suggest to rename it to smth like `(get_)map_ref`

michaelvoronov (approved):

PR #979, opened 2020-11-19
Doc improvements, script generator

DieMyst:


client/fluence-js/src/aqua/scripts.ts
```diff
@@ -0,0 +1,123 @@
...
+    ${wrap(command, scripts)}
+)`
+    }
+}
+
+/**
+ * Wrap an array of scripts with multiple 'seq' commands
+ * @param scripts
+ */
+export function seq(scripts: string[]): string {
```
folex, line 106: I think it is better to accept two arguments: `seq(left: string, right: string)`

folex (approved):

PR #1008, opened 2020-12-24
Send particle.data in base64 through the wire

folex:


crates/json-utils/src/base64_serde.rs
```diff
@@ -0,0 +1,34 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+
+pub fn serialize<S>(value: &Vec<u8>, serializer: S) -> Result<S::Ok, S::Error>
```
michaelvoronov, line 19: Please add allow for clippy here, because it seems that &Vec<u8> is really needed here.

particle-actors/Cargo.toml
```diff
@@ -21,3 +21,4 @@ async-std = { version = "1.6.5", features = ["unstable"] }
 serde_json = "1.0.57"
 parking_lot = "0.11.0"
 chrono = "0.4.19"
+base64 = "0.13.0"
```
michaelvoronov, line 24: new line

particle-actors/src/actor.rs
```diff
@@ -173,11 +184,9 @@ impl Actor {
 
 fn protocol_error(mut particle: Particle, err: impl Debug) -> ActorEvent {
     let error = format!("{:?}", err);
-    if let Some(map) = particle.data.as_object_mut() {
-        map.insert("protocol!error".to_string(), json!(error));
-    } else {
-        particle.data = json!({"protocol!error": error, "data": particle.data})
-    }
+    particle.data = json!({"protocol!error": error, "data": base64::encode(particle.data)})
+        .to_string()
```
michaelvoronov, line 188: Can you please add a comment why conversion to string is needed here.

particle-actors/src/invoke.rs
```diff
@@ -77,14 +93,21 @@ fn parse_peer_id(s: &str) -> Result<PeerId, FieldError> {
...
+) -> Result<(Vec<u8>, Vec<PeerId>), ExecutionError> {
     let outcome = outcome.map_err(|err| ExecutionError::AquamarineError(err))?;
-    let data = serde_json::from_slice(outcome.data.as_slice()).map_err(|err| {
-        ExecutionError::InvalidResultField {
-            field: "data",
-            error: FieldError::InvalidJson(err),
-        }
-    })?;
+
+    if outcome.ret_code != 0 {
```
michaelvoronov, line 99: What do you think maybe it's better to add a new error type such as StepperError(StepperOutcome) to AquamarineVM?

michaelvoronov (approved):

particle-actors/src/invoke.rs
folex: I think it is a good idea, yes. But then I think it should return `Result<AquamarineOutcome, AquamarineError>`, and hide `ret_code` and `error_message` altogether.

michaelvoronov: Yes, that's I meant

PR #1015, opened 2021-01-17
update Signature use

boneyard93501:


trust-graph/src/key_pair.rs
```diff
@@ -19,6 +19,8 @@ use ed25519_dalek::SignatureError;
 use libp2p_core::identity::error::DecodingError;
 use std::fmt;
 
+// switch to 
+// use ed25519_dalek::Signature; ?
 pub type Signature = Vec<u8>;
```
michaelvoronov, line 24: I think that it is better to switch to ed25519_dalek::Signature on this PR if it doesn't require too many modifications. This is due to changing the signature of verify function. At the moment &Signature means &Vec, but passing &Vec is anti-pattern for Rust, it is better to pass slices instead. But the verify signature would be more expressive if we use Signature instead of raw slice. And on the other hand ed25519_dalek::Signature is a slice with fixed lengths.

DieMyst: I already do my stuff in another repo. So, I'll do it there to avoid confuse

folex:
Since trust-graph has moved to a separate repository, I'll close this one. @DieMyst will continue work in another repo.

PR #1017, opened 2021-01-21
Introduce ConnectionPool & localize libp2p

folex:


particle-node/src/network_api.rs
```diff
@@ -0,0 +1,175 @@
...
+/// This structure is just a composition of Kademlia and ConnectionPool.
+/// It exists solely for code conciseness (i.e. avoid tuples);
+/// there's no architectural motivation behind
+pub struct Connectivity {
+    pub kademlia: KademliaApi,
+    pub connection_pool: ConnectionPoolApi,
+}
+
+impl Connectivity {
+    /// Perform effects that Aquamarine (through [[StepperPoolApi]]) instructed us to
```
michaelvoronov, line 118: What is StepperPoolApi? I can't find in code, seems that it should be StepperEffects instead?

particle-node/src/network_api.rs
```diff
@@ -0,0 +1,175 @@
...
+    pub async fn execute_effects(&self, effects: StepperEffects) {
```
michaelvoronov, line 119: Again, what about introduction a new trait to abstract over stepper? I think that it would allow to make all entities inside networl_api.rs independent on other system parts and make an architecture border over it.

particle-node/src/network_api.rs
```diff
@@ -0,0 +1,175 @@
...
+            // forward particle
+            self.connection_pool.send(contact, particle).await;
+        }
+    }
+
+    /// Discover a peer via Kademlia
+    pub async fn discover_peer(&self, target: PeerId) -> Result<Option<Contact>, KademliaError> {
+        // discover contact addresses through Kademlia
+        let addresses = self.kademlia.discover_peer(target).await?;
+        if !addresses.is_empty() {
```
michaelvoronov, line 157: Suggest not to use the negation here, but rearrange if branches.

michaelvoronov (approved):

michaelvoronov:
There are also some warnings that could be found by clippy

particle-node/src/network_api.rs
folex: It's an old name for `Aquamarine`

PR #1020, opened 2021-01-26
Remove Bootstrapper, express bootstrap logic in async/await

folex:


connection-pool/src/behaviour.rs
```diff
@@ -131,12 +138,7 @@ impl ConnectionPoolBehaviour {
     }
 
     pub fn get_contact(&self, peer_id: PeerId, outlet: OneshotOutlet<Option<Contact>>) {
```
michaelvoronov, line 140: I think that this function deserves a comment because "get" could be understanded in different ways. Or maybe call it "send_contact_to" to resolve name clashing.

connection-pool/src/behaviour.rs
```diff
@@ -243,6 +256,15 @@ impl ConnectionPoolBehaviour {
             )))
         }
     }
+
+    fn get_contact0(&self, peer_id: PeerId) -> Option<Contact> {
```
michaelvoronov, line 260: Maybe get_contact_impl?

particle-node/src/network_api.rs
```diff
@@ -65,22 +74,91 @@ impl NetworkApi {
...
+    ) {
+        let events = pool.lifecycle_events();
+        let disconnections = {
+            let bootstrap_nodes = bootstrap_nodes.clone();
+            events
+                .filter_map(move |e| {
+                    if let LifecycleEvent::Disconnected(Contact { addresses, .. }) = e {
+                        let addresses = addresses.into_iter();
+                        let addresses = addresses.filter(|addr| bootstrap_nodes.contains(addr));
+                        let addresses = iter(addresses.collect::<Vec<_>>());
```
michaelvoronov, line 107: Is it possible not to collect here and return just filtered addresses?

michaelvoronov (approved):

particle-node/src/network_api.rs
folex: Nope, we need to collect to avoid keeping references to `bootstrap_nodes`, I haven't managed to avoid collect here.

But I found an unnecessary clone because of your comment!

connection-pool/src/behaviour.rs
folex: It is derived from the API of `ConnectionPoolT`, so I think it is better to leave it the same. Here `outlet` should be considered an `out` argument. I would say `outlet` works here as a return type.

PR #1022, opened 2021-01-28
Simplistic script storage

folex:


crates/server-config/src/defaults.rs
```diff
@@ -70,5 +70,9 @@ pub fn default_particle_queue_buffer_size() -> usize {
 }
 
 pub fn default_particle_processor_parallelism() -> usize {
-    num_cpus::get()
+    num_cpus::get() * 4
```
michaelvoronov, line 73: Why * 4? Usually it *2 + 1

crates/server-config/src/defaults.rs
```diff
@@ -70,5 +70,9 @@ pub fn default_particle_queue_buffer_size() -> usize {
...
+}
+
+pub fn default_script_storage_interval() -> Duration {
+    Duration::from_secs(3)
```
michaelvoronov, line 77: Maybe make it const?

michaelvoronov (approved):

crates/server-config/src/defaults.rs
folex: That's not a number of threads, it's a number of particles processed concurrently. I think it could be * 128 without much trouble. Let's see how `*4` would behave. I don't have a better idea right now.

crates/server-config/src/defaults.rs
folex: It's called on config deserialization on node start. So I think it is fine being that way.

PR #1024, opened 2021-02-02
Script storage: enable setting custom interval

folex:


script-storage/src/script_storage.rs
```diff
@@ -49,11 +47,26 @@ impl Borrow<String> for ScriptId {
 struct Script {
     src: String,
     failures: u8,
+    /// Interval at which to execute this script.
```
michaelvoronov, line 50: in which?

michaelvoronov (approved):

PR #1025, opened 2021-02-02
Add (script add) and restrict (script remove) to owners

folex:


crates/test-utils/src/connected_client.rs
```diff
@@ -175,4 +176,18 @@ impl ConnectedClient {
         let particle = self.receive();
         read_args(particle, &self.peer_id)
     }
+
+    pub fn wait_particle_args(&mut self, particle_id: String) -> anyhow::Result<Vec<JValue>> {
```
michaelvoronov, line 180: Some comment here would be great

michaelvoronov (approved):

PR #1035, opened 2021-02-15
Migrate modules

folex:


particle-modules/src/modules.rs
```diff
@@ -72,6 +77,29 @@ impl ModuleRepository {
...
     }
 
+    /// check that module file name is equal to module hash
+    /// if not, rename module and config files
+    fn maybe_migrate_module(path: &Path, hash: &ModuleHash, modules_dir: &Path) {
+        use eyre::eyre;
+
+        let migrated: eyre::Result<_> = try {
+            let file_name = extract_module_file_name(&path).ok_or(eyre!("none"))?;
+            if file_name != hash.to_hex().as_ref() {
```
michaelvoronov, line 87: You could invert the comparison here to decrease identation level of instructions inside this `if`.

michaelvoronov (approved):

particle-modules/src/modules.rs
folex: How?

PR #1058, opened 2021-03-26
Configure timeouts in tests

folex:


folex:
Will be merged as a part of #1059

PR #1078, opened 2021-04-28
srv remove

folex:


particle-services/src/app_services.rs
```diff
@@ -122,6 +122,31 @@ impl ParticleAppServices {
...
+            match services.read().get(&service_id) {
+                Some(service) if service.owner_id != particle_params.init_user_id => {
+                    Err(ServiceError::Forbidden {
+                        user: particle_params.init_user_id,
+                        function: "remove_service",
+                        reason: "only creator can remove service",
+                    })
+                }
+                None => Err(ServiceError::NoSuchService(service_id.clone())),
+                _ => Ok(()),
```
justprosh, line 141: I think `Some(service) => Ok(())` is more readable

justprosh (approved):

justprosh:
Maybe add test to check deletion non-existed service

PR #1089, opened 2021-05-18
fix timestamp_ms execution problem, add test

justprosh:
Previously `("peer" "timestamp_ms")` execution failed with:
```
[2021-05-18T21:43:24.932946Z INFO  aquamarine::particle_executor] Executing particle 4d8cb775-3d73-4f42-84d4-fd2fd095a312
thread 'blocking-58' panicked at 'called `Result::unwrap()` on an `Err` value: Error("u128 is not supported", line: 0, column: 0)', /home/circleci/project/particle-closures/src/host_closures.rs:139:53
```

### DIFF
--- a/particle-node/tests/builtin.rs
+++ b/particle-node/tests/builtin.rs
@@ -200,7 +200,7 @@ fn timestamp_ms() {
         },
     );
 
-    let info = client.receive_args().wrap_err("receive args").unwrap();
-    let info = info.into_iter().next().unwrap();
-    let _: u64 = serde_json::from_value(info).unwrap();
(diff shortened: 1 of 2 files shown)


particle-node/tests/builtin.rs
```diff
@@ -178,3 +178,29 @@ fn non_owner_remove_service() {
         panic!("incorrect args: expected two arrays, got: {:?}", args)
     }
 }
+
+#[test]
+fn timestamp_ms() {
```
folex, line 183: Can you plz also add test on timestamp_sec?

folex (approved):

PR #1092, opened 2021-05-27
base58 and sha256 builtins

folex:
### base58 utils:
```
op string_to_b58     : string     => utf8 bytes => base58
op string_from_b58   : base58     => utf8 bytes => string
op bytes_to_b58      : byte array => base58
op bytes_from_b58    : base58     => byte array
```

### sha256
`op sha256_string(s: string, digest_only = false, as_bytes = false)`

parameters explanation:
- `s` – string to hash
- `digest_only` boolean – if set to true, return only SHA256 digest, otherwise (by default) – full multihash
- `as_bytes` boolean - if set to true, return result as array of bytes, otherwise (by default) – as base58 string


particle-node/tests/builtin.rs
```diff
@@ -403,3 +407,132 @@ fn timestamp_sec() {
...
+        result[1],
+        json!(bs58::encode(sha_256.digest()).into_string())
+    );
+    // multihash as byte array
+    assert_eq!(result[2], json!(sha_256.to_bytes()));
+    // sha256 digest as byte array
+    assert_eq!(result[3], json!(sha_256.digest()));
+}
+
+fn call_builtin(
```
justprosh, line 512: call_script?

particle-node/tests/builtin.rs
```diff
@@ -403,3 +407,132 @@ fn timestamp_sec() {
...
+
+    let bytes: Vec<_> = (1..32).map(|i| (200 + i) as u8).collect();
+    let b58_string = bs58::encode(&bytes).into_string();
+    let args = hashmap! {
+        "b58_string" => json!(b58_string),
+        "bytes" => json!(bytes),
+    };
+
+    let result = call_builtin(script, args, "b58_string_out bytes_out identity_bytes");
+    // let array = into_array(result).expect("must be an array");
```
justprosh, line 457: dead code?

justprosh (approved):

PR #1103, opened 2021-06-03
Add ops noop, concat, array; change op identity to return scalar

folex:
`(op noop)` – always returns empty string (unit)
`(op array)` – accepts any number of arguments and wraps them in a single array
`(op concat)` – accepts any number of arrays and concatenates them
`(op identity)` – accepts at most 1 arguments and returns it; if no arguments, empty string (unit) is returned

### Types
```scala
Op.noop() -> ()
Op.array(args: any[]) -> any[]
Op.concat(args: any[][]) -> any[]
Op.identity(a: Option<any>) -> Option<any>
```

### DIFF
--- a/particle-closures/src/host_closures.rs
+++ b/particle-closures/src/host_closures.rs
@@ -170,7 +170,7 @@ impl<C: Clone + Send + Sync + 'static + AsRef<KademliaApi> + AsRef<ConnectionPoo
 
             ("op", "noop")                    => unit(),
             ("op", "array")                   => ok(Array(args.function_args)),
-            ("op", "identity")                => wrap_opt(Ok(args.function_args.into_iter().next())),
+            ("op", "identity")                => wrap_opt(self.identity(args.function_args)),
             ("op", "concat")                  => wrap(self.concat(args.function_args)),
             ("op", "string_to_b58")           => wrap(self.string_to_b58(args.function_args)),
             ("op", "string_from_b58")         => wrap(self.string_from_b58(args.function_args)),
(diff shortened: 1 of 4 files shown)


particle-node/tests/builtin.rs
```diff
@@ -223,7 +223,8 @@ fn non_owner_remove_service() {
 
     use serde_json::Value::{Array, String};
 
-    let args = client2.receive_args().unwrap();
+    println!("will receive");
```
justprosh, line 226: ???

justprosh (approved):

particle-node/tests/builtin.rs
folex: oops :)

particle-closures/src/host_closures.rs
```diff
@@ -379,6 +380,26 @@ impl<C: Clone + Send + Sync + 'static + AsRef<KademliaApi> + AsRef<ConnectionPoo
...
 
+    /// Flattens an array of arrays
+    fn concat(&self, args: Vec<serde_json::Value>) -> Result<JValue, JError> {
+        let flattened: Vec<JValue> = args
+            .into_iter()
+            .enumerate()
+            .map(|(i, mut v)| match v.take() {
+                JValue::Array(array) => Ok(array),
+                _ => Err(JError::new(format!(
+                    "all arguments of 'concat' must be arrays: argument #{} wasn't",
```
justprosh, line 391:
```suggestion
                    "all arguments of 'concat' must be arrays: argument #{} is not",
```
Maybe

justprosh (approved):

PR #1107, opened 2021-06-04
blueprint hash: sort deps without facade module

justprosh:


particle-modules/src/modules.rs
```diff
@@ -405,12 +405,14 @@ fn hash_dependencies(deps: Vec<Dependency>) -> Result<Hash> {
             }),
         })?;
 
+    let facade = deps.pop().expect("dependency list is empty!");
```
folex, line 408: plz return `Err(EmptyDependenciesList)` here

particle-modules/src/modules.rs
```diff
@@ -405,12 +405,14 @@ fn hash_dependencies(deps: Vec<Dependency>) -> Result<Hash> {
...
     deps.sort_by(|a, b| a.as_bytes().cmp(&b.as_bytes()));
 
     for d in deps.iter() {
         hasher.update(d.as_bytes());
     }
 
+    hasher.update(facade.as_bytes());
```
folex, line 415:
```suggestion
    let deps = deps.iter().chain(iter::once(facade));
    for d in deps.iter() {
        hasher.update(d.as_bytes());
    }
    
```

folex (changes requested):

folex (approved):

PR #1126, opened 2021-06-14
set rust toolchain

dzmitry-lahoda:
so other can easy build and switch to this repo


dzmitry-lahoda:
i want to try out https://github.com/fluencelabs/fluence/pull/1110

rust-toolchain
```diff
@@ -0,0 +1 @@
+nightly-2021-04-24
```
folex, line 1: I'd put just `nightly` here

.github/workflows/rust_ci.yml
```diff
@@ -22,6 +22,7 @@ jobs:
         uses: actions-rs/toolchain@v1
         with:
           profile: minimal
+          # keep in sync rust-toolchain (you may cat into here as option)
```
folex, line 25:
```suggestion
```

rust-toolchain
folex, line 1: I wonder if it's possible to move `rust-toolchain` to `.rustc` or `.cargo` to avoid cluttering project root with files

folex (changes requested):
Specific version in CI is needed because on nightly channel not all components are released at once. So if using just `nightly` (i.e. latest nightly), it's possible that `rustfmt` will be missing, for example, and build will fail.

This is inevitable for latest nightly, so we're using specific version only as a mean to avoid that problem.

rust-toolchain
dzmitry-lahoda: it will not work. the version must be exactly same as in ci

dzmitry-lahoda: for thoose who upgrades can use https://rust-lang.github.io/rustup/overrides.html - so the problem of version is the only problem of upgrader, not of others

dzmitry-lahoda: can use override per folder
per package

dzmitry-lahoda: i do not think so https://rust-lang.github.io/rustup/overrides.html

folex: > it will not work. the version must be exactly same as in ci for master

what wouldn't work?

dzmitry-lahoda: i have newer nightly that this, and it fails to compule dht (becayse of try feature v2). similar as i recall was for this repo. anyway - nightly must be either latest always, or if you do not update - must be precise release to allow other peoples collaborate on code.

folex: Ah, I see. Node doesn't compile on higher versions. That's a very unusual situation, I'll fix that ASAP

folex: I've fixed the problem with `try_feature` in #1127. Now you can safely use `nightly` in `rust-toolchain` file

folex: @dzmitry-lahoda, let's change to `nightly` and commit

folex (changes requested):
In aqua-dht repo it turned out ([1](https://app.circleci.com/pipelines/github/fluencelabs/aqua-dht/148/workflows/72d608b3-9d6e-419d-92c4-91e8fb4940f6/jobs/228), [2](https://github.com/fluencelabs/aqua-dht/runs/2846799062)) that we also need to add
```
rustup override set nightly-2021-06-14-x86_64-unknown-linux-gnu
```

to override `rust-toolchain` file for CI

```

dzmitry-lahoda:
@folex are you sure you are not complicating things? may be make master same as CI? if somebody wants to use new feature, have a branch and wait for CI is ready.

dzmitry-lahoda:
Nightly already `alfa`, so you are trying to use alfa-alfa-unknown.

folex:
> @folex are you sure you are not complicating things? may be make master same as CI? if somebody wants to use new feature, have a branch and wait for CI is ready.

I'm thinking about developers who don't have installed `nightly-2021-06-14`, but have installed `nightly`. I see no major reason to force them to install `nightly-2021-06-14`.

If it not random unavailability of components in latest `nightly`, I'd use it in CI.

dzmitry-lahoda:
as a other developer i prefer to install exact version which works, and ci proves that. it like depending on exact p2p crate instead on random latest.

folex:
> as a other developer i prefer to install exact version which works, and ci proves that. it like depending on exact p2p crate instead on random latest.

I think you have a good point that I haven't thought about initially. Since we're using unstable features like `try_trait_v2`, build on `nightly` might brake, and confuse developers who try to build node from sources. 🤔

dzmitry-lahoda:
actually i did not wanted nodes, i wanted tool (rust client) - so I added new crate client to fluence fork.

dzmitry-lahoda:
so what should be done?

folex (approved):

PR #1140, opened 2021-06-29
Added additional funcs to made possible run air with Rust

silvestrpredko:


crates/local-vm/src/local_vm.rs
```diff
@@ -96,6 +96,9 @@ pub fn make_call_service_closure(
                 service_out.lock().extend(args.function_args);
                 ivalue_utils::unit()
             }
+            ("errorHandlingSrv", "error") => {
+                ivalue_utils::unit()
```
folex, line 100: I think that we need to at least log error here; At best - fail `receive_args` with that error.

silvestrpredko: @folex I have added a simple log, is it enough?

folex: Nope, you need to log the passed arguments. Like this:
```
log::warn!("caught an error: {:?}", args.function_args);
```

folex (commented):
What about variable loading? 

If you just pass `hashmap!` with data, local_vm would insert `load` instructions before `getDataSrv` instructions and the script will be invalid.

You need something like this:
https://github.com/fluencelabs/fluence/pull/1110/files#diff-f0a2f5fe61a4b414b315b9d6011ff0a5e79423a03e3b94195cc7c3fae0039c2fR142

silvestrpredko:
@folex You mean

> What about variable loading?
> 
> If you just pass `hashmap!` with data, local_vm would insert `load` instructions before `getDataSrv` instructions and the script will be invalid.
> 
> You need something like this:
> https://github.com/fluencelabs/fluence/pull/1110/files#diff-f0a2f5fe61a4b414b315b9d6011ff0a5e79423a03e3b94195cc7c3fae0039c2fR142

You mean I need to load variables inside make_call_service_closure?

folex:
Superseded by #1143

PR #1147, opened 2021-07-02
Stop node if some builtins is failed to load

justprosh:


crates/builtins-deployer/src/builtins_deployer.rs
```diff
@@ -432,11 +432,15 @@ impl BuiltinsDeployer {
...
         failed
-            .into_iter()
-            .map(|err| log::error!("builtin load failed: {:#}", err))
+            .iter()
+            .map(|err| log::error!("failed to load some builtins: {:#}", err))
             .for_each(drop);
 
-        Ok(successful)
+        return if !failed.is_empty() {
+            Err(eyre!("some builtins load failed"))
```
folex, line 440: Let's collect builtin names into `failed` and show it in the error. `Something has failed` isn't very helpful.

Or, if you're sure that errors in `failed` will contain builtin names already, reword the error like `failed to load builtins from disk: {:#}", err`

folex (changes requested):

folex (approved):

PR #1161, opened 2021-07-17
save builtins autodeploy key between restarts

justprosh:


particle-node/src/main.rs
```diff
@@ -117,8 +116,11 @@ fn start_fluence(config: ResolvedConfig) -> eyre::Result<impl Stoppable> {
     log::info!("node public key = {}", bs58_key_pair);
     log::info!("node server peer id = {}", to_peer_id(&key_pair.into()));
 
-    let startup_management_key_pair = Keypair::generate_ed25519();
-    let startup_peer_id = to_peer_id(&startup_management_key_pair);
+    // TODO: add to_string() for KeyFormat in fluence-identity
+    let startup_keypair =
+        load_key_pair(default_builtins_keypair_path(), "ed25519".to_string(), true)
```
folex, line 121: I think the path should come from the `config`

folex (changes requested):

folex (approved):

PR #1164, opened 2021-07-20
Added configuration for Particle TTL

silvestrpredko:


crates/local-vm/src/local_vm.rs
```diff
@@ -198,7 +198,7 @@ pub fn make_particle(
         id,
         init_peer_id: peer_id,
         timestamp: now_ms() as u64,
-        ttl: PARTICLE_TTL,
+        ttl: particle_ttl.map_or_else(|| { PARTICLE_TTL }, |ttl| { ttl.as_millis() as u32 }),
```
folex, line 201: I think `map_or` should be enough, no reason for closures here

folex (approved):
One more parameter to `connect_with_*`, and we should move to `ClientConfig`, ie:
```rust
struct ClientConfig {
  node_address: Multiaddr,
  key_pair: Option<Keypair>,
  timeout: ...,
  particle_ttl : ...,
}
fn connect(config: ClientConfig)
```

PR #1166, opened 2021-07-27
autodeploy: add particle_ttl to config

justprosh:


deploy/Config.toml
```diff
@@ -6,6 +6,8 @@ path = "/node.key"
 format = "ed25519"
 generate_on_absence = true
 
+autodeploy_particle_ttl = "60s"
```
folex, line 9: why? default value is the same

crates/server-config/src/node_config.rs
```diff
@@ -28,6 +28,11 @@ pub struct NodeConfig {
     #[derivative(Debug = "ignore")]
     pub builtins_key_pair: KeyPair,
 
+    /// Particle ttl for autodeploy
+    #[serde(default)]
```
folex, line 32: how ttl could be none? let's move sane default here

crates/builtins-deployer/src/builtins_deployer.rs
```diff
@@ -179,6 +181,7 @@ impl BuiltinsDeployer {
         node_peer_id: PeerId,
         node_api: AquamarineApi,
         base_dir: PathBuf,
+        particle_ttl: Option<Duration>,
```
folex, line 184: why option?

folex (changes requested):

folex (approved):

PR #1223, opened 2021-11-12
Builtins: add Peer.timeout

folex:


particle-builtins/src/builtins.rs
```diff
@@ -102,6 +102,7 @@ where
             ("peer", "is_connected")          => wrap(self.is_connected(args).await),
             ("peer", "connect")               => wrap(self.connect(args).await),
             ("peer", "get_contact")           => self.get_contact(args).await,
+            ("peer", "timeout")               => self.timeout(args).await,
```
justprosh, line 105: maybe "timeout_sec"?
like `timestamp_sec` and `timestamp_ms`

justprosh (approved):

particle-builtins/src/builtins.rs
folex: It's gonna be documented in the `aqua-lib`:
```
    -- Blocks for the given number of milliseconds. Meant to be used inside `par` blocks.
    -- If `message` is passed, it's returned after duration has passed
    timeout(duration_ms: u64, message: ?string) -> ?string
```

I've changed it to `ms`, btw.

PR #1262, opened 2022-03-25
Scheduled scripts: derive particle_id from script_id

folex:
Also:
- Add debug logs to understand why failed scripts aren't removed
- Express `oneshot` logic explicitly with `executions` and `times` state variables


particle-node/tests/script_storage.rs
```diff
@@ -610,6 +610,7 @@ fn add_script_random_delay() {
     let res = client.wait_particle_args(now_id).unwrap().pop().unwrap();
     let now = res.as_u64().unwrap();
 
+    // client.timeout = Duration::from_secs(120);
```
justprosh, line 613:
```suggestion
```

justprosh (approved):

PR #1271, opened 2022-04-25
Log connection failures

folex:
As @mxinden has suggested in https://github.com/libp2p/rust-libp2p/issues/2602#issuecomment-1107441110, it makes sense to log dial and listening errors. This way, we can know who causes errors like recent EOF-bug.

### DIFF
--- a/connection-pool/src/behaviour.rs
+++ b/connection-pool/src/behaviour.rs
@@ -408,12 +408,30 @@ impl NetworkBehaviour for ConnectionPoolBehaviour {
         _handler: Self::ConnectionHandler,
         error: &DialError,
     ) {
+        log::warn!(
+            "Error dialing peer {}: {:?}",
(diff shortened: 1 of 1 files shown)


justprosh (approved):

connection-pool/src/behaviour.rs
```diff
@@ -408,6 +408,11 @@ impl NetworkBehaviour for ConnectionPoolBehaviour {
         _handler: Self::ConnectionHandler,
         error: &DialError,
     ) {
+        log::warn!(
+            "Error dialing peer {}: {:?}",
```
mxinden, line 412: Note that you could as well log the `SwarmEvent` emitted by `Swarm`. I guess just a matter of taste.

https://github.com/libp2p/rust-libp2p/blob/22fbce34d5c2c3c7fccaf7c7ce9f80a92406a641/swarm/src/lib.rs#L185-L201

PR #1273, opened 2022-04-26
Builtins: add Stat.service_memory

folex:
```
data ModuleMemoryStats:
  name: string
  memory_size_bytes: u32
  max_memory_size_bytes: ?u32

service Stat("stat")
   service_memory(service_id: string) => ModuleMemoryStats[]
```

```json
[Object({"max_memory_size_bytes": Null, "memory_size_bytes": Number(1114112), "name": String("tetraplets")})]
```

### DIFF
--- a/particle-node/tests/builtin.rs
+++ b/particle-node/tests/builtin.rs
@@ -1027,6 +1027,42 @@ fn index_by_math() {
     assert_eq!(element[0], json!(4));
 }
 
+#[test]
+fn service_mem() {
+    let swarms = make_swarms(1);
+
+    let mut client = ConnectedClient::connect_to(swarms[0].multiaddr.clone())
+        .wrap_err("connect client")
+        .unwrap();
+
+    let tetraplets_service = create_service(
+        &mut client,
+        "tetraplets",
+        load_module("tests/tetraplets/artifacts", "tetraplets").expect("load module"),
+    );
+
+    client.send_particle(
+        r#"
+        (seq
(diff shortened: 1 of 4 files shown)


justprosh (approved):

kmd-fl (approved):

particle-node/tests/builtin.rs
```diff
@@ -1027,6 +1027,42 @@ fn index_by_math() {
...
+        },
+    );
+
+    use serde_json::Value::Array;
+
+    if let [Array(stats)] = client.receive_args().unwrap().as_slice() {
+        println!("{:?}", stats);
+    } else {
+        panic!("incorrect args: expected single arrays of module memory stats")
+    }
```
mikevoronov, line 1063: looks a bit complex, maybe just assert? Smth like 
```rust
assert_eq!(client.receive_args().unwrap(), json!([stats]))
```

folex: I like the `if let Array` approach. It makes expected types very visible and allows to easily adapt is returning type changes

mikevoronov: Hmm, but if you organize your could smth like that
```
let expected_stats = Stats::(...);
let expected_args = json!([expected_stats]);
assert_eq!(client.receive_args().unwrap(), expected_args);
```

it'd be as visible as in the youir approach?

And btw, you don't check stats itself, just their type.

But all of this is very nit, so let's merge it.

PR #1280, opened 2022-06-17
Replace mutex with async queue for metrics collection

kmd-fl:
Fixes NET-87


kmd-fl:
I haven't properly tested it yet.

folex (approved):

PR #1292, opened 2022-08-03
Use updated marine config with load_from field

ValeryAntopol:


particle-modules/src/files.rs
```diff
@@ -40,14 +40,23 @@ pub fn load_blueprint(bp_dir: &Path, blueprint_id: &str) -> Result<Blueprint> {
 pub fn load_module_descriptor(modules_dir: &Path, module_hash: &Hash) -> Result<ModuleDescriptor> {
     let config = modules_dir.join(module_config_name_hash(module_hash));
     let config = load_config_by_path(&config)?;
-    let mut config: ModuleDescriptor = config
+    // TODO CHECK: Is "." right? Maybe we need different path?
+    let context = ConfigContext {
+        base_path: PathBuf::from("."),
```
folex, line 45: yeah, it's wrong. We have `ServicesConfig::modules_dir`, I believe that's what you should use.

particle-modules/src/files.rs
```diff
@@ -40,14 +40,23 @@ pub fn load_blueprint(bp_dir: &Path, blueprint_id: &str) -> Result<Blueprint> {
...
+    let mut config: ModuleDescriptor = context
+        .wrapped(config)
         .try_into()
         .map_err(|err| ModuleConvertError { err })?;
 
     // TODO HACK: This is required because by default file_name is set to be same as import_name
     //            That behavior is defined in TomlMarineNamedModuleConfig. Would be nice to refactor that behavior.
     config.file_name = module_file_name_hash(module_hash);
 
+    // Node stores its modules in a cernain directory, and it does not need effect of load_from field
```
folex, line 57:
```suggestion
    // Ignore `load_from` because modules are always stored next to config files
```

particle-modules/src/files.rs
```diff
@@ -84,6 +93,8 @@ pub fn add_module(
     // TODO HACK: use custom structure for API; TomlMarineNamedModuleConfig is too powerful and clumsy.
     // Set file_name = ${hash}.wasm
     config.file_name = Some(module_config_name_hash(module_hash));
+    // Node stores its modules in a cernain directory, and it does not need effect of load_from field
+    config.load_from = None;
```
folex, line 97:
```suggestion
    // Ignore `load_from` because modules are always stored next to config files
```

particle-modules/src/files.rs
```diff
@@ -40,14 +40,23 @@ pub fn load_blueprint(bp_dir: &Path, blueprint_id: &str) -> Result<Blueprint> {
...
+        .wrapped(config)
         .try_into()
         .map_err(|err| ModuleConvertError { err })?;
 
     // TODO HACK: This is required because by default file_name is set to be same as import_name
     //            That behavior is defined in TomlMarineNamedModuleConfig. Would be nice to refactor that behavior.
     config.file_name = module_file_name_hash(module_hash);
 
+    // Node stores its modules in a cernain directory, and it does not need effect of load_from field
+    config.load_from = None;
```
folex, line 58: I don't think you need to do that here.

particle-modules/src/files.rs
ValeryAntopol: I think using "." keeps old behaviour with relative paths. So, if i use `ServicesConfig::modules_dir`, then what should i put into `modules_dir` config field? "."?

ValeryAntopol: As discussed in private, using `Option::None` here

particle-modules/src/files.rs
```diff
@@ -40,7 +40,10 @@ pub fn load_blueprint(bp_dir: &Path, blueprint_id: &str) -> Result<Blueprint> {
 pub fn load_module_descriptor(modules_dir: &Path, module_hash: &Hash) -> Result<ModuleDescriptor> {
     let config = modules_dir.join(module_config_name_hash(module_hash));
     let config = load_config_by_path(&config)?;
-    let mut config: ModuleDescriptor = config
+    let context = ConfigContext { base_path: None };
```
folex, line 43: Please add a comment explaining the None

particle-modules/src/files.rs
```diff
@@ -84,6 +87,8 @@ pub fn add_module(
     // TODO HACK: use custom structure for API; TomlMarineNamedModuleConfig is too powerful and clumsy.
     // Set file_name = ${hash}.wasm
     config.file_name = Some(module_config_name_hash(module_hash));
+    // Ignore `load_from` because modules are always stored next to config files
```
folex, line 90: Would be great to elaborate a little on how `load_from` connected to location of the config files

particle-modules/src/files.rs
```diff
@@ -40,7 +40,11 @@ pub fn load_blueprint(bp_dir: &Path, blueprint_id: &str) -> Result<Blueprint> {
 pub fn load_module_descriptor(modules_dir: &Path, module_hash: &Hash) -> Result<ModuleDescriptor> {
     let config = modules_dir.join(module_config_name_hash(module_hash));
     let config = load_config_by_path(&config)?;
-    let mut config: ModuleDescriptor = config
+    // None makes marine use the old behaviour, interpreting all paths as relative to $CURRENT_DIR
```
folex, line 43:
```suggestion
    // `base_path: None` tells Marine to either expected absolute paths to WASM files, or resolve non-absolute paths relative to location of the service configuration file
```

particle-modules/src/files.rs
```diff
@@ -84,6 +88,9 @@ pub fn add_module(
     // TODO HACK: use custom structure for API; TomlMarineNamedModuleConfig is too powerful and clumsy.
     // Set file_name = ${hash}.wasm
     config.file_name = Some(module_config_name_hash(module_hash));
+    // The `load_from` field overrides `modules_dir` for a single module,
+    // so we ignore `load_from` because modules are always stored next to config files
```
folex, line 92:
```suggestion
    // so we set `load_from` to `None`, telling Marine to load modules from the same location of the module config file
```

folex (approved):

PR #1299, opened 2022-09-02
Add Signing service

justprosh:
Closes NET-199


particle-builtins/src/builtins.rs
```diff
@@ -89,13 +96,16 @@ where
         let particles_vault_dir = vault_dir.to_path_buf();
         let management_peer_id = config.management_peer_id;
         let builtins_management_peer_id = config.builtins_management_peer_id;
+        let local_peer_id = config.local_peer_id.clone();
```
folex, line 99: You don't need to clone PeerIds

particle-builtins/src/builtins.rs
```diff
@@ -738,6 +753,51 @@ where
...
+            .get(0)
+            .ok_or_else(|| JError::new(error.to_string()))?
+            .get(0)
+            .ok_or_else(|| JError::new(error.to_string()))?;
+
+        if tetraplet.peer_pk != self.local_peer_id.to_base58()
+            || tetraplet.service_id != "registry"
+            || tetraplet.function_name != "get_record_bytes"
+        {
+            return Err(JError::new(error.to_string()));
```
folex, line 773: ditto

particle-builtins/src/builtins.rs
```diff
@@ -738,6 +753,51 @@ where
...
+        }
+
+        Ok(JValue::Array(
+            self.root_keypair
+                .sign(&data)?
+                .to_vec()
+                .into_iter()
+                .map(|n| json!(n))
```
folex, line 781: Why serialize each byte individually?

particle-builtins/src/builtins.rs
```diff
@@ -738,6 +753,51 @@ where
             }))
         }
     }
+
+    fn sign(&self, args: Args) -> Result<JValue, JError> {
+        let tetraplets = args.tetraplets;
+        let mut args = args.function_args.into_iter();
+        let data: Vec<u8> = Args::next("data", &mut args)?;
```
folex, line 760: shouldn't it be named like `let record_bytes = Args::next("record_bytes", ...)`? What is `data`?

particle-builtins/src/builtins.rs
```diff
@@ -738,6 +753,51 @@ where
...
+
+        let error = "data should be passed from local registry.get_record_bytes()";
```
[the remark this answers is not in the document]
folex: same, what is `data` here? How is it called in user's scenario? Maybe "record"? Gotta describe it in a way that makes sense. Aqua variable name would be the best, but I don't think it's possible.

particle-builtins/src/builtins.rs
justprosh: We have a common API for both Sig services (in Rust and JS) and the Sig service in JS is now more general.
Later we will make Rust one too

particle-builtins/src/builtins.rs
```diff
@@ -738,6 +753,51 @@ where
...
+                .into_iter()
+                .map(|n| json!(n))
+                .collect(),
+        ))
+    }
+
+    fn verify(&self, args: Args) -> Result<JValue, JError> {
+        let mut args = args.function_args.into_iter();
+        let signature: Vec<u8> = Args::next("signature", &mut args)?;
+        let data: Vec<u8> = Args::next("data", &mut args)?;
```
folex, line 789: data => record_bytes

particle-builtins/src/builtins.rs
```diff
@@ -738,6 +753,51 @@ where
...
+        let signature =
+            Signature::from_bytes(self.root_keypair.public().get_key_format(), signature);
+
+        Ok(JValue::Bool(
+            self.root_keypair.public().verify(&data, &signature).is_ok(),
```
folex, line 794: Maybe return an error message together with `false`?

particle-builtins/src/builtins.rs
```diff
@@ -738,6 +753,51 @@ where
...
+        ))
+    }
+
+    fn get_peer_id(&self) -> Result<JValue, JError> {
+        Ok(JValue::String(self.root_keypair.get_peer_id().to_base58()))
```
folex, line 799: There's `self.local_peer_id`

particle-builtins/src/builtins.rs
justprosh: afaiu you can serialize array only of JValues but you can't serialize an array of u8 without converting them to JValue

particle-builtins/src/builtins.rs
justprosh: I don't agree with that. We have API in [aqua-lib](https://github.com/fluencelabs/aqua-lib/blob/main/builtin.aqua#L302) and there is no sense to rename it for Rust peer specifically

particle-builtins/src/builtins.rs
justprosh: I prefer to use root_keypair now because later if we will change the Sig service to sign with different keypairs (like it is done for JS) there will be less room for mistakes

particle-builtins/src/builtins.rs
justprosh: We don't have it in JS and there is no meaningful message for verification failure

particle-builtins/src/builtins.rs
folex: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=22d7cf6c19c5f189ab6cbe9d62c09523

particle-builtins/src/builtins.rs
folex: There are two cases where verification fails:
1. Signature bytes are of incorrect format
2. Signing equation is incorrect (ie signature forged or made for another data)

It's rather hard to tell one from the other when debugging cryptography code. One mistake in signature storage, and you always get this pesky `false` without ever knowing that the signature format is incorrect.

I don't know why, but even though you use `Signature::from_bytes`, according ed25519-dalek's code it's still possible to get a signature format error. See:
```rust
impl TryFrom<&ed25519::Signature> for InternalSignature {
    type Error = SignatureError;

    fn try_from(sig: &ed25519::Signature) -> Result<InternalSignature, SignatureError> {
        InternalSignature::from_bytes(sig.as_bytes())
    }
}
```

So I insist on having an error message along the result. Let's not allow previous decisions on API to be a handicap forever. At least without a very good reason.

particle-builtins/src/builtins.rs
folex: Well, then why add `local_peer_id` field at all? You either add it and use in both `verify` and `get_peer_id`, or don't add anywhere and derive it from `root_keypair` always.

particle-builtins/src/builtins.rs
justprosh: Okay. 

@coder11 I will file an issue for Sig.verify return type change

justprosh: DXJ-107

particle-node/tests/builtin.rs
```diff
@@ -1203,6 +1204,117 @@ fn service_stats_uninitialized() {
...
+    {
+        let err_msg = "Local service error, ret_code is 1, error message is '\"data should be passed from local registry.get_record_bytes()\"'";
+        assert_eq!(host_error, err_msg);
+        assert_eq!(srv_error, err_msg);
+        assert_eq!(func_error, err_msg);
+    } else {
+        panic!("incorrect args: expected three arguments")
+    }
+}
+
```
folex, line 1317: Would be nice to have a test on incorrect signature as well.

particle-builtins/src/builtins.rs
justprosh: local_peer_id is used for sign data that comes only from local services

justprosh: it's just different meanings

justprosh: `root_keypair `is used in all methods, `local_peer_id` — only in sign for permissions

folex (approved):

PR #1306, opened 2022-09-07
Move from forked libp2p to crates.io

folex:


particle-node/src/lib.rs
```diff
@@ -18,7 +18,7 @@
 #![feature(ip)]
 #![recursion_limit = "512"]
 #![warn(rust_2018_idioms)]
-#![deny(
+#![allow(
```
justprosh, line 21: ???? 😀

justprosh (approved):

PR #1324, opened 2022-09-30
Gather some service metrics for aliased services separately

kmd-fl:


aquamarine/src/particle_functions.rs
```diff
@@ -161,11 +161,7 @@ impl<F: ParticleFunctionStatic> Functions<F> {
             block_on(async move {
                 let outcome = builtins.call(args, params).await;
                 // record whether call was handled by builtin or not. needed for stats.
-                let call_kind = if outcome.is_defined() {
-                    FunctionKind::Builtin
-                } else {
-                    FunctionKind::Service
-                };
```
folex, line 168: You're not using `FunctionKind::Builtin` anymore?

crates/peer-metrics/src/particle_executor.rs
```diff
@@ -12,6 +12,7 @@ use crate::execution_time_buckets;
 pub enum FunctionKind {
     Builtin,
     Service,
+    Function,
```
folex, line 15: `FunctionKind::Function` is kinda hard to understand. What about `FunctionKind::ParticleFunction`?

particle-services/src/app_services.rs
```diff
@@ -235,15 +199,14 @@ impl ParticleAppServices {
         }
         let service = self.services.write().remove(&service_id).unwrap();
         let mut aliases = self.aliases.write();
+        let service_type = ServiceType::Service(service.aliases.first().cloned());
```
folex, line 202: Is order of aliases preserved between node restarts?

crates/peer-metrics/src/services_metrics/backend.rs
```diff
@@ -84,8 +85,8 @@ impl ServicesMetricsBackend {
                     msg = inlet.select_next_some() => {
                         match msg {
                             // save data to the map
-                            ServiceMetricsMsg::Memory { service_id, memory_stat } => {
-                                Self::observe_service_mem(&mut services_memory_stats, service_id, memory_stat);
+                            ServiceMetricsMsg::Memory { service_id, service_type, memory_stat } => {
+                                Self::observe_service_mem(&mut services_memory_stats, service_id, service_type, memory_stat);
                             },
                             ServiceMetricsMsg::CallStats { service_id, function_name, stats } => {
                                 builtin_metrics.update(service_id, function_name, stats);
```
folex, line 92: Do you record call stats for aliased services anywhere? If not, then why not?

aquamarine/src/particle_functions.rs
folex: If not, then does it make sense to have these metrics by `FunctionKindLabel` then?
```
service_call_time_sec: Family<FunctionKindLabel, Histogram>,
service_call_success: Family<FunctionKindLabel, Counter>,
service_call_failure: Family<FunctionKindLabel, Counter>,
```
Maybe stop recording these metrics, and record everything via `CallStat`/`ServicesMetricsBuiltin`?

particle-builtins/src/builtins.rs
```diff
@@ -113,8 +113,26 @@ where
...
+                if let Some(metrics) = self.services.metrics.as_ref() {
+                    metrics.observe_builtins(
+                        !matches!(result, FunctionOutcome::Err { .. }),
+                        end as f64,
+                    );
+                }
+                result
+            }
+        }
+    }
```
folex, line 132: Nice!

particle-builtins/src/builtins.rs
```diff
@@ -196,7 +214,7 @@ where
             ("sig", "verify")      => wrap(self.verify(args)),
             ("sig", "get_peer_id") => wrap(self.get_peer_id()),
 
-            _                      => self.call_service(args, particle),
+            _                      => FunctionOutcome::NotDefined { args, params: particle },
```
folex, line 217: 👍

folex (changes requested):
Everything looks good! But we also need to collect call stats for aliased services.

I think that `service_call_*` stats from `ParticleExecutorMetrics` should be merged with `CallStats` and collected for both aliased and non-aliased services in a same manner.

particle-services/src/app_services.rs
kmd-fl: It seems so. Toml is responsible for de/serialization, and we don't break the order in the process. Additionally, I checked and it was fine.

folex (approved):

PR #1327, opened 2022-10-04
Add --secret-key, move from base58 to base64, store secret key instead of keypair

folex:
fixes NET-66 NET-164 NET-202


justprosh (approved):
lgtm but don't forget to rename PR according to conventional commits

PR #1343, opened 2022-11-17
Draft: basic spells instalation

kmd-fl:
Need to manually place the spell service to `.fluence/v1/spell_bundle` or any path configured with `spell_base_dir` option.

```aqua
service Spell("spell"):
  list() -> []string
  install(script: string) -> string
  remove(script: string)
```

### DIFF
--- a/particle-builtins/src/builtins.rs
+++ b/particle-builtins/src/builtins.rs
@@ -31,7 +31,7 @@ use libp2p::{core::Multiaddr, kad::kbucket::Key, kad::K_VALUE, PeerId};
 use multihash::{Code, MultihashDigest, MultihashGeneric};
 use parking_lot::{Mutex, RwLock};
 use serde::Deserialize;
-use serde_json::{json, Value as JValue};
+use serde_json::{json, Value as JValue, Value};
 use JValue::Array;
 
 use connection_pool::{ConnectionPoolApi, ConnectionPoolT};
@@ -44,9 +44,11 @@ use particle_modules::{
 };
 use particle_protocol::Contact;
 use particle_services::{ParticleAppServices, VIRTUAL_PARTICLE_VAULT_PREFIX};
+use particle_services::ServiceError::Forbidden;
 use peer_metrics::ServicesMetrics;
 use script_storage::ScriptStorageApi;
 use server_config::ServicesConfig;
+use uuid_utils::uuid;
 
 use crate::debug::fmt_custom_services;
 use crate::error::HostClosureCallError;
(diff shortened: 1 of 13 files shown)


particle-builtins/src/builtins.rs
```diff
@@ -104,8 +108,28 @@ where
         let management_peer_id = config.management_peer_id;
         let builtins_management_peer_id = config.builtins_management_peer_id;
         let local_peer_id = config.local_peer_id;
+        let spells_base_dir = config.spells_base_dir.clone();
         let services = ParticleAppServices::new(config, modules.clone(), Some(services_metrics));
 
+        let spell_blueprint_id = services.load_spell_service(spells_base_dir).unwrap();
```
folex, line 114: shouldn't this be inside SpellStorage?

particle-services/src/app_services.rs
```diff
@@ -396,6 +399,14 @@ impl ParticleAppServices {
         Ok(self.modules.get_facade_interface(&service.blueprint_id)?)
     }
 
+    pub fn list_service_with_blueprints(&self) -> Vec<(String, String)> {
```
folex, line 402:
```suggestion
    pub fn list_services_with_blueprints(&self) -> Vec<(String, String)> {
```

particle-services/src/app_services.rs
```diff
@@ -533,6 +544,26 @@ impl ParticleAppServices {
     fn create_vault(&self, particle_id: &str) -> Result<(), VaultError> {
         self.vault.create(particle_id)
     }
+
+    pub fn load_spell_service(&self, spells_base_dir: PathBuf) -> Result<String, ModuleError> {
+        log::info!("Spell service: {}", spells_base_dir.display());
```
folex, line 549: I guess this is a debug log, so you will remove that?

particle-services/src/app_services.rs
```diff
@@ -533,6 +544,26 @@ impl ParticleAppServices {
...
+        let spell_cfg_path = spells_base_dir.to_owned().join("Config.toml");
+        let cfg = TomlMarineConfig::load(spell_cfg_path).unwrap();
+        let spell_cfg_prefix = PathBuf::from(spells_base_dir);
+        let mut hashes = Vec::new();
+        for config in cfg.module {
+            let load_from = config
+                .load_from
+                .clone()
+                .unwrap_or(PathBuf::from(config.name.to_owned() + ".wasm"));
```
folex, line 558: there's `module_file_name` for `.wasm` things.

Why do you use `config.name` instead of module hash like everything else?

particle-services/src/app_services.rs
```diff
@@ -533,6 +544,26 @@ impl ParticleAppServices {
     fn create_vault(&self, particle_id: &str) -> Result<(), VaultError> {
         self.vault.create(particle_id)
     }
+
+    pub fn load_spell_service(&self, spells_base_dir: PathBuf) -> Result<String, ModuleError> {
```
folex, line 548: I don't really understand why this function is needed. I would expect you to reuse existing functions for service creation and all that 🤔

`SpellStorage` should copy configs and .wasms to appropriate places, and then you just do `AppServices::create_service` on `spell.install`.

particle-builtins/src/builtins.rs
```diff
@@ -786,6 +815,15 @@ where
         let mut args = args.function_args.into_iter();
         let service_id_or_alias: String = Args::next("service_id_or_alias", &mut args)?;
 
+        if self.spell_storage.has_spell(&service_id_or_alias) {
```
folex, line 818: How would it work with alias? 🤔 I don't see you doing `register_spell` on `add_alias`, so here's how I can remove a spell:
1. Create spell
2. Add alias
3. Remove by alias
4. `has_spell(alias)` returns `false`
5. `services.remove_service(alias)` goes and removes the spell

PR #1353, opened 2022-11-29
Spell event dispatcher

kmd-fl:


connection-pool/src/behaviour.rs
```diff
@@ -115,6 +116,7 @@ pub struct ConnectionPoolBehaviour {
     events: VecDeque<SwarmEventType>,
     waker: Option<Waker>,
     pub(super) protocol_config: ProtocolConfig,
+    send_events: Outlet<PeerEvent>,
```
folex, line 119: I think `subscribers: Vec<Outlet<LifecycleEvent>>` should be enough for the task 🤔

connection-pool/src/behaviour.rs
```diff
@@ -464,6 +478,12 @@ impl NetworkBehaviour for ConnectionPoolBehaviour {
         _: <Self::ConnectionHandler as IntoConnectionHandler>::Handler,
         remaining_established: usize,
     ) {
+        if let Err(e) = self.send_events.unbounded_send(PeerEvent::Disconnect {
+            peer_id: peer_id.clone(),
+        }) {
+            log::warn!("Failed to send peer event: {}", e);
+        }
+
```
folex, line 486: That's gonna fire a lot of Disconnected events for each peer if they have several multiaddresses. There's `remove_contact` that is called only when a peer is entirely disconnected. There are a lot of edge cases.

But that doesn't really matter, LifecycleEvents should be enough and they're already implemented.

crates/events-dispatcher/src/api.rs
```diff
@@ -0,0 +1,118 @@
...
+#[derive(Debug)]
+pub struct Listener {
+    pub id: ListenerId,
+}
+
+#[derive(Debug)]
+pub struct ListenerEvent {
+    pub id: ListenerId,
+    pub event: Event,
+}
```
folex, line 44:
```suggestion
pub type SpellId = String;

#[derive(Debug)]
pub struct Spell {
    pub id: SpellId,
}

#[derive(Debug)]
pub struct SpellCast {
    pub id: SpellId,
    pub trigger: Trigger,
}
```

crates/events-dispatcher/src/api.rs
```diff
@@ -0,0 +1,118 @@
...
+
+#[derive(Clone, Debug)]
+pub enum Event {
+    Timer,
+    Peer(PeerEvent),
+}
+
+#[derive(Clone, Debug)]
+pub enum PeerEvent {
```
folex, line 53:
```suggestion
pub enum ConnectionPoolEvent {
```

crates/events-dispatcher/src/api.rs
```diff
@@ -0,0 +1,118 @@
...
+
+#[derive(PartialEq, Eq, Hash, Debug, Clone, Deserialize)]
+pub enum PeerEventType {
+    Connect,
+    Disconnect,
+}
+
+pub enum Command {
+    /// Subscribe a listener with a specified ID to a list of events
+    Add(Listener, ListenerConfig),
```
folex, line 75:
```suggestion
    Subscribe(Spell, TriggerConfig),
```

crates/events-dispatcher/src/dispatcher.rs
```diff
@@ -0,0 +1,246 @@
...
+                let next_task_in = if let Some(next_task) = heap.peek() {
+                    next_task
+                        .run_at
+                        .checked_duration_since(now)
+                        .unwrap_or_default()
+                } else {
+                    Duration::MAX
+                };
+
+                async_std::stream::interval(next_task_in).fuse()
```
folex, line 138: I don't really understand what is happening here.

`unwrap_or_default` means that interval could become 0, and timer will start firing madly, sending thread into a busy loop basically.

`Duration::MAX` means that interval would be infinity, and then what about periodic tasks? I would expect interval to be  `timer_resolution`. If you want to optimize, you can calculate the minimum needed interval, and then set lower limit to `timer_resolution` anyway: 
```
max(
    timer_resolution, 
    min(
        now - next_task.run_at, 
        minimum_interval_of_all_spells
    )
)
```

crates/events-dispatcher/src/dispatcher.rs
```diff
@@ -0,0 +1,246 @@
...
+            .collect::<Vec<_>>();
+        let mut sources_channel = stream::select_all(sources);
+
+        let mut heap: BinaryHeap<Scheduled<String>> = BinaryHeap::new();
+        let mut listeners = Listeners::new();
+
+        loop {
+            let now = Instant::now();
+            let mut timer = {
+                let next_task_in = if let Some(next_task) = heap.peek() {
```
folex, line 129: Also I don't understand where `task` name is coming from. It should be something from the spell dictionary, eg `trigger ` or `time_trigger` or `cast` if you prefer

crates/events-dispatcher/src/dispatcher.rs
```diff
@@ -0,0 +1,246 @@
...
+                        }
+                    }
+                },
+                _ = timer.select_next_some() => {
+                    let task = heap.pop().expect("billions of years have gone by already?");
+                    if let Err(e) = send_events.unbounded_send(ListenerEvent { id: task.data.id.clone(), event: Event::Timer }) {
+                        log::warn!("Can't send notification by timer: {:?}", e);
+                    }
+                    heap.push(task.reschedule(now));
+                },
```
folex, line 180: looks like it could be DRYed

crates/events-dispatcher/src/dispatcher.rs
```diff
@@ -0,0 +1,246 @@
...
+                                        let periodic = Periodic { id: listener.id.clone(), period: config.period };
+                                        heap.push(Scheduled::new(periodic, Instant::now()))
+                                    },
+                                    EventConfig::PeerEvent(config) => {
+                                        listeners.add(listener.clone(), config.events);
+                                    },
+                                }
+                            }
+                        },
+                        Command::Remove(listener_id, reply) => {
```
folex, line 158: OneshotOutlet is usually called `out` everywhere (eg ConnectionPoolApi and KademliaApi), let's stick to consistent naming 

```suggestion
                        Command::Remove(listener_id, out) => {
```

crates/events-dispatcher/src/dispatcher.rs
```diff
@@ -0,0 +1,246 @@
...
+                    }
+                    heap.push(task.reschedule(now));
+                },
+
+            }
+        }
+    }
+}
+#[test]
+fn test() {
```
folex, line 187: test name should be meaningful

crates/events-dispatcher/src/dispatcher.rs
```diff
@@ -0,0 +1,246 @@
...
+    let (send1, recv1) = unbounded();
+    task::spawn(async move {
+        loop {
+            task::sleep(std::time::Duration::from_secs(2)).await;
```
folex, line 191: that's too long. Can you eliminate sleeping? Or at least reduce it to almost zero?

crates/events-dispatcher/src/dispatcher.rs
```diff
@@ -0,0 +1,246 @@
...
+                    heap.push(task.reschedule(now));
+                },
+
+            }
+        }
+    }
+}
+#[test]
+fn test() {
+    let (send1, recv1) = unbounded();
```
folex, line 188: what is send1, what is send2?

crates/events-dispatcher/src/dispatcher.rs
kmd-fl: You're probably missing that this `interval` is created once for each loop iteration now, not just once. On each iteration, it will fire when the next periodic task must be executed. This way, we don't even need the concept of `timer_resolution`. We're not polling the time to see if a periodic task is overdue; instead, we're relying on `interval` to wake up when the task is due.

> `unwrap_or_default` means that interval could become 0

Yes. This means that running some periodic task is overdue. In this case, the timer will be `select`'ed immediately, and the task will execute.

> `Duration::MAX` means that interval would be infinity, and then what about periodic tasks?

`MAX` will only be in the case when there are no periodic tasks registered yet. If we process a command to register a new periodic task, then, in the next iteration, the duration will account for this.

crates/events-dispatcher/src/dispatcher.rs
kmd-fl: Didn't really intend for this to be a test, more like a usage example that could also be run to check that everything works. There are even no assertions.

crates/events-dispatcher/src/dispatcher.rs
folex: Can you please document all of these implicit assumptions? Write it in comments in the code

crates/events-dispatcher/src/dispatcher.rs
folex: Usage examples should be given in comments, maybe as doc-tests so the compilation is checked.

But I'm not sure this reads as a documentation, it seems just like some code without any story behind it.

PR #1356, opened 2022-12-05
Add spell event bus

kmd-fl:
Ported improvements from the scheduler before deleting it.

NET-275


crates/spell-event-bus/src/api.rs
```diff
@@ -0,0 +1,141 @@
...
+
+#[derive(Debug)]
+pub struct PeerEventConfig {
+    pub events: Vec<PeerEventType>,
+}
+
+pub type SpellId = String;
+
+#[derive(Debug)]
+pub struct SpellEvent {
```
folex, line 31: Hmm, reads as `event produced by spell`. How about one of those: `SpellRunRequest` / `RunRequest` / `SpellCast` / `CastRequest` / `Trigger`?

I guess given all the `TriggerConfig`s, the `Trigger` would be the most suitable.

crates/spell-event-bus/src/api.rs
```diff
@@ -0,0 +1,141 @@
...
+/// Types of events that are available for subscription.
+#[derive(PartialEq, Eq, Hash, Debug, Clone)]
+pub enum PeerEventType {
+    Connected,
+    Disconnected,
+}
+
+#[derive(Debug)]
+pub(crate) enum Command {
+    /// Subscribe a listener with a specified ID to a list of events
```
folex, line 70:
```suggestion
    /// Subscribe a spell to a list of triggers
```

crates/spell-event-bus/src/api.rs
```diff
@@ -0,0 +1,141 @@
...
+    }
+}
+
+impl SpellEventBusApi {
+    fn send(&self, cmd: Command) -> Result<(), EventBusError> {
+        self.send_cmd_channel.unbounded_send(cmd).map_err(|e| {
+            let reason = if e.is_disconnected() {
+                "disconnected"
+            } else {
+                "full"
```
folex, line 105: How can an unbounded channel be full?

crates/spell-event-bus/src/api.rs
folex: On the other hand, it's an **Event**Bus, right? So maybe `Event` is good.

crates/spell-event-bus/src/bus.rs
```diff
@@ -0,0 +1,299 @@
...
+                        // The timer is triggered only if there are some spells to be awaken.
+                        let scheduled_spell = state.scheduled.pop().expect("billions of years have gone by already?");
+                        Self::notify_spell(&send_events, &scheduled_spell.data.id, Event::Timer)?;
+                        let rescheduled = Scheduled::at(scheduled_spell.data, Instant::now())?;
+                        state.scheduled.push(rescheduled);
+                    },
+                }
+            };
+            if let Err(e) = result {
+                log::error!("{}", e);
```
folex, line 230: That won't do. You have to put some explanation on what happened, and why this error matters, and what it relates to.

Also, it doesn't seem like a critical error, so I would put it to `warn`. Better still, I would route that error back to the client where possible. 

Eg if `Subscribe`, `Unsubscribe` or `UpdateTriggerConfig` failed, it MUST be visible in Aqua/AIR. So such an error MUST be passed back to builtins.

crates/spell-event-bus/src/bus.rs
```diff
@@ -0,0 +1,299 @@
...
+            }
+        }
+    }
+
+    fn notify_spell(
```
folex, line 235: Event Bus _notifies_ spells about _events_ that happened according to _trigger_ config. 3 different words for the same thing.

How about this?
```suggestion
    fn trigger_spell(
```

sorcerer/src/spells.rs
```diff
@@ -52,12 +87,12 @@ pub(crate) fn spell_install(
...
-        service_id.clone(),
-        TimerConfig {
+    let config = api::SpellTriggerConfigs {
+        triggers: vec![api::TriggerConfig::Timer(TimerConfig {
             period: Duration::from_secs(period),
-        },
-    )?;
+        })],
+    };
+    spell_event_bus_api.subscribe(service_id.clone(), config)?;
```
folex, line 95: Will the returned error make sense to the human who calls `spell.install`? Maybe add some context for the error?

sorcerer/src/spells.rs
```diff
@@ -79,6 +115,8 @@ pub(crate) fn spell_remove(
     let mut args = args.function_args.into_iter();
     let spell_id: String = Args::next("spell_id", &mut args)?;
 
+    spell_event_bus_api.unsubscribe(spell_id.clone()).await?;
```
folex, line 118: ditto: context

crates/spell-event-bus/src/bus.rs
```diff
@@ -0,0 +1,299 @@
...
+use futures::stream;
+use futures::stream::BoxStream;
+use futures::{channel::mpsc::unbounded, select, StreamExt};
+use std::cmp::Ordering;
+use std::collections::{BinaryHeap, HashMap};
+use std::time::{Duration, Instant};
+use thiserror::Error;
+
+struct Subscribers {
+    subscribers: HashMap<PeerEventType, Vec<Arc<SpellId>>>,
```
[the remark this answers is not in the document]
justprosh: Oh, it is done to avoid copying

crates/spell-event-bus/src/api.rs
kmd-fl: That's the interface [TrySendError](https://docs.rs/futures/latest/futures/channel/mpsc/struct.TrySendError.html)

crates/spell-event-bus/src/bus.rs
kmd-fl: > what happened, and why this error matters, and what it relates to.

Doesn't the `log` crate add this info automatically? And errors are explained in the `thiserror::Error` impl.

`unsubscribe` right now can fail only when replying to dropped channel, so here we can't do anything else. It seems we need to add a similar channel for `subscribe` to notify about the subscription status.

folex: Without a prefix, it would be impossible to find the line of code that produced that log.

folex: > And errors are explained in the thiserror::Error impl.

Can these errors happen under different circumstances, in different places? If yes, then log must explain where this error happened and how to interpret it.

crates/spell-event-bus/src/api.rs
kmd-fl: Because you can get only one thing from this structure: an error type or content. https://docs.rs/futures/latest/futures/channel/mpsc/struct.TrySendError.html#method.into_inner

crates/spell-event-bus/src/bus.rs
kmd-fl: Well, I think internal errors of the bus shouldn't be available to users, just the fact that smth that don't depend on them went wrong and they should report it. The wrong things that can happen in all these commands are all internal now. The only exception is the timer config, but it should be validated before, and it should be impossible to construct an invalid config.

crates/spell-event-bus/src/bus.rs
```diff
@@ -198,51 +191,53 @@ impl SpellEventBus {
...
-                        let rescheduled = Scheduled::at(scheduled_spell.data, Instant::now())?;
+                        Self::trigger_spell(&send_events, &scheduled_spell.data.id, Event::Timer)?;
+                        let rescheduled = Scheduled::at(scheduled_spell.data, Instant::now()).expect("invalid timer config detected while rescheduling a spell");
                         state.scheduled.push(rescheduled);
                     },
                 }
             };
             if let Err(e) = result {
-                log::error!("{}", e);
+                log::error!("spell-event-bus internal error: {}", e);
```
folex, line 225: Not `log::error`, but `log::warn`, please. 

`log::error` should **only** happen when there's an error that's critical to Rust Peer. Like, something really broken and rust-peer won't ever work.

folex:
```suggestion
                log::warn!("Error in spell event bus run loop: {:?}", e);
```

crates/spell-event-bus/src/bus.rs
```diff
@@ -0,0 +1,294 @@
...
+            };
+
+            let result: Result<(), BusInternalError> = try {
+                select! {
+                    command = recv_cmd_channel.select_next_some() => {
+                        let Command { spell_id, action, reply } = command;
+                        match &action {
+                            Action::Subscribe(config) => {
+                                // TODO: make it possible to construct the config ONLY via `verify :: UserConfig -> Result<SpellTriggerConfigs, _>`.
+                                state.subscribe(spell_id.clone(), &config).expect("invalid timer config detected while subscribing a spell");
```
folex, line 201: I suggest to return `Err` instead of panic here. As I understand it, `expect` would allow anyone to pass invalid config and crash whole EventBus.

Also, this error about invalid timer should be sent back to user. 

You said before that
> The only exception is the timer config, but it should be validated before, and it should be impossible to construct an invalid config.

If this is true, then maybe express it in types? And pass validated config here, thus avoiding validation on this level.

crates/spell-event-bus/src/bus.rs
```diff
@@ -0,0 +1,294 @@
...
+                        for spell_id in state.subscribers(&event.get_type()) {
+                            let event = Event::Peer(event.clone());
+                            Self::trigger_spell(&send_events, spell_id, event)?;
+                        }
+                    },
+                    _ = timer.select_next_some() => {
+                        // The timer is triggered only if there are some spells to be awaken.
+                        let scheduled_spell = state.scheduled.pop().expect("billions of years have gone by already?");
+                        Self::trigger_spell(&send_events, &scheduled_spell.data.id, Event::Timer)?;
+                        let rescheduled = Scheduled::at(scheduled_spell.data, Instant::now()).expect("invalid timer config detected while rescheduling a spell");
```
folex, line 219: Same about validating config. Let's pass already-validated config and avoid validation on this level.

crates/spell-event-bus/src/bus.rs
kmd-fl: This is happening only because we can't prove to the compiler that `now + period` won't overflow. `Instant::now() + Duration` will panic on overflow, `Instant::now().checked_add(Duration)` returns `None` on overflow.

So, by the introduced invariant, it's impossible to give an invalid config, so there's nothing to report back to the user, but it's impossible to express it in types. We can use `+`, but the function won't be total, and implicit unwrapping will be happening nevertheless.  

I believe it's better to make this panic explicit.

sorcerer/src/spells.rs
```diff
@@ -52,13 +89,21 @@ pub(crate) fn spell_install(
...
-        TimerConfig {
-            period: Duration::from_secs(period),
-        },
-    )?;
-    Ok(JValue::String(service_id))
+    if let Err(err) = spell_event_bus_api
+        .subscribe(spell_id.clone(), config.clone())
+        .await
+    {
+        log::error!("can't subscribe a spell {} to triggers {:?} via spell-event-bus-api: {}. Removing created spell service...", spell_id, config, err);
```
folex, line 96:
```suggestion
        log::warn!("can't subscribe a spell {} to triggers {:?} via spell-event-bus-api: {}. Removing created spell service...", spell_id, config, err);
```

sorcerer/src/spells.rs
```diff
@@ -52,13 +89,21 @@ pub(crate) fn spell_install(
...
+
+        spell_storage.unregister_spell(&spell_id);
+        services.remove_service(spell_id, params.init_peer_id)?;
+
+        return Err(JError::new(format!(
+            "can't install a spell due to an internal error"
```
folex, line 102: Could you please return a more descriptive error? Like the log you have above.

This will be useful when someone sends us this error and asks for help.

sorcerer/src/spells.rs
```diff
@@ -70,14 +115,26 @@ pub(crate) fn spell_list(spell_storage: SpellStorage) -> Result<JValue, JError>
...
+    spell_event_bus_api: SpellEventBusApi,
     spell_storage: SpellStorage,
     services: ParticleAppServices,
     args: Args,
     params: ParticleParams,
 ) -> Result<(), JError> {
     let mut args = args.function_args.into_iter();
     let spell_id: String = Args::next("spell_id", &mut args)?;
+    if let Err(err) = spell_event_bus_api.unsubscribe(spell_id.clone()).await {
+        log::error!(
```
folex, line 128:
```suggestion
        log::warn!(
```

sorcerer/src/spells.rs
```diff
@@ -70,14 +115,26 @@ pub(crate) fn spell_list(spell_storage: SpellStorage) -> Result<JValue, JError>
...
+            "can't unsubscribe a spell {} from its triggers via spell-event-bus-api: {}",
+            spell_id,
+            err
+        );
+        return Err(JError::new(format!(
+            "can't remove a spell {} due to an internal error",
```
folex, line 134: More descriptive error.

crates/spell-event-bus/src/bus.rs
folex: One option is to just do nothing on `None`.

crates/spell-event-bus/src/bus.rs
```diff
@@ -198,7 +198,7 @@ impl SpellEventBus {
                         match &action {
                             Action::Subscribe(config) => {
                                 // TODO: make it possible to construct the config ONLY via `verify :: UserConfig -> Result<SpellTriggerConfigs, _>`.
-                                state.subscribe(spell_id.clone(), &config).expect("invalid timer config detected while subscribing a spell");
+                                state.subscribe(spell_id.clone(), &config).unwrap_or(());
```
folex, line 201: Maybe write a log?

folex (approved):

justprosh (changes requested):
Please, add tests for connection pool events

crates/spell-event-bus/src/api.rs
```diff
@@ -0,0 +1,152 @@
...
+pub enum PeerEventType {
+    Connected,
+    Disconnected,
+}
+
+#[derive(Debug)]
+pub(crate) struct Command {
+    pub(crate) spell_id: SpellId,
+    pub(crate) action: Action,
+    pub(crate) reply: OneshotOutlet<()>,
```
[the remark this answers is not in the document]
folex: > Please, add tests for connection pool events

In the next PR. This has been hanging for 7 days already. Where have you been? :)

justprosh (approved):

PR #1357, opened 2022-12-05
feat(spells): read arguments, store results and handle errors

justprosh:
closes NET-296, NET-297, NET-300

### DIFF
--- a/particle-builtins/src/builtins.rs
+++ b/particle-builtins/src/builtins.rs
@@ -56,6 +56,9 @@ use crate::identify::NodeInfo;
 use crate::outcome::{ok, wrap, wrap_unit};
 use crate::{json, math};
 
+// TODO: move to appropriate place
+const DEFAULT_FUNCTION_NAME: &str = "__default__";
+
 #[derive(Derivative)]
 #[derivative(Debug)]
 pub struct Builtins<C> {
@@ -72,6 +75,10 @@ pub struct Builtins<C> {
     pub services: ParticleAppServices,
     pub node_info: NodeInfo,
 
+    /// HashMap = (service_id -> function_name -> service_function)
+    /// You can set function name to `DEFAULT_FUNCTION_NAME`
+    /// and all `function_name` mismatches will be routed to this `service_function`.
+    /// Inside `service_function` `function_name` can be obtained via `args.function_name`
     #[derivative(Debug(format_with = "fmt_custom_services"))]
     pub custom_services: RwLock<HashMap<String, HashMap<String, Mutex<ServiceFunction>>>>,
 
@@ -143,7 +150,10 @@ where
             .custom_services
             .read()
             .get(&args.service_id)
-            .and_then(|fs| fs.get(&args.function_name))
+            .and_then(|fs| {
+                fs.get(&args.function_name)
+                    .or(fs.get(DEFAULT_FUNCTION_NAME))
+            })
         {
             let mut function = function.lock();
             async_std::task::block_on(function(args, particle))
--- a/particle-builtins/src/builtins.rs
+++ b/particle-builtins/src/builtins.rs
@@ -56,8 +56,10 @@ use crate::identify::NodeInfo;
 use crate::outcome::{ok, wrap, wrap_unit};
 use crate::{json, math};
 
-// TODO: move to appropriate place
-const DEFAULT_FUNCTION_NAME: &str = "__default__";
+pub struct CustomService {
+    pub functions: HashMap<String, Mutex<ServiceFunction>>,
+    pub unhandled: Option<Mutex<ServiceFunction>>,
+}
 
 #[derive(Derivative)]
 #[derivative(Debug)]
@@ -80,7 +82,7 @@ pub struct Builtins<C> {
     /// and all `function_name` mismatches will be routed to this `service_function`.
     /// Inside `service_function` `function_name` can be obtained via `args.function_name`
     #[derivative(Debug(format_with = "fmt_custom_services"))]
-    pub custom_services: RwLock<HashMap<String, HashMap<String, Mutex<ServiceFunction>>>>,
+    pub custom_services: RwLock<HashMap<String, CustomService>>,
 
     particles_vault_dir: path::PathBuf,
 }
@@ -151,8 +153,9 @@ where
             .read()
             .get(&args.service_id)
             .and_then(|fs| {
-                fs.get(&args.function_name)
-                    .or(fs.get(DEFAULT_FUNCTION_NAME))
+                fs.functions
+                    .get(&args.function_name)
+                    .or(fs.unhandled.as_ref())
             })
         {
             let mut function = function.lock();
--- a/particle-builtins/src/builtins.rs
+++ b/particle-builtins/src/builtins.rs
@@ -57,7 +57,9 @@ use crate::outcome::{ok, wrap, wrap_unit};
 use crate::{json, math};
 
 pub struct CustomService {
+    /// (function_name -> service function)
     pub functions: HashMap<String, Mutex<ServiceFunction>>,
+    /// if set, all `function_name` mismatches with `custom_service.functions` will be routed to `unhandled`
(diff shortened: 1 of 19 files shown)


particle-node/download_spell.sh
```diff
@@ -7,7 +7,7 @@ SPELL_DIR="${CUR_DIR}/spell"
 TAR="spell.tar.gz"
 
 echo "*** downloading $TAR ***"
-URL="https://github.com/fluencelabs/spell/releases/download/v0.0.3/spell.tar.gz"
+URL="https://github.com/fluencelabs/spell/releases/latest/download/spell.tar.gz"
```
folex, line 10: it is gonna break hard and bite us hard. better specify exact version.

particle-node/tests/spells.rs
```diff
@@ -87,3 +102,52 @@ fn spell_simple_test() {
     assert_eq!(result, script);
     assert_ne!(counter, 0);
 }
+
+#[test]
+fn spell_error_handling_test() {
+    std::fs::remove_file("/tmp/spell.sqlite").ok();
```
folex, line 108: don't need that

particle-node/download_spell.sh
justprosh: oh, it's already in master

justprosh: https://github.com/fluencelabs/rust-peer/blob/master/particle-node/download_spell.sh#L10

sorcerer/src/utils.rs
```diff
@@ -0,0 +1,56 @@
...
+pub(crate) fn process_func_outcome<T>(func_outcome: FunctionOutcome) -> Result<T, JError>
+where
+    T: DeserializeOwned + SpellValueT,
+{
+    match func_outcome {
+        FunctionOutcome::NotDefined { args, .. } => Err(JError::new(format!(
+            "Service with id '{}' not found (function {})",
+            args.service_id, args.function_name
+        ))),
+        FunctionOutcome::Empty => Err(JError::new("Function has not returned any result")),
```
folex, line 32: Maybe add service id and function name to the error?

sorcerer/src/utils.rs
```diff
@@ -0,0 +1,56 @@
...
+        FunctionOutcome::Err(err) => Err(JError::new(err.to_string())),
+        FunctionOutcome::Ok(v) => {
+            let result = serde_json::from_value::<T>(v)?;
+            if result.is_success() {
+                Ok(result)
+            } else {
+                Err(JError::new(result.get_error()))
```
folex, line 39: Ought to add an explanation, like `Result of a function {}.{} cannot be parsed to {}, service, function, std::any::type_name<T>()`

sorcerer/src/utils.rs
```diff
@@ -0,0 +1,56 @@
...
+pub(crate) fn parse_spell_id_from(particle_id: String) -> eyre::Result<String, JError> {
+    if particle_id.starts_with("spell_") {
+        Ok(particle_id
+            .split('_')
+            .collect::<Vec<&str>>()
+            .get(1)
+            .ok_or(JError::new("Invalid particle id format"))?
+            .to_string())
+    } else {
+        Err(JError::new("Invalid particle id format"))
```
folex, line 54:
```suggestion
        Err(JError::new("Invalid spell particle id format"))
```

sorcerer/src/spells.rs
```diff
@@ -86,16 +99,74 @@ pub(crate) fn spell_remove(
...
+    .map(drop)
+    .map_err(|e| {
+        JError::new(format!(
+            "Failed to store error {:?} for spell {}: {}",
+            args.function_args, spell_id, e
+        ))
+    })
+}
+
+pub(crate) fn response_handler(
```
folex, line 150:
```suggestion
pub(crate) fn persist_response(
```

sorcerer/src/spells.rs
```diff
@@ -86,16 +99,74 @@ pub(crate) fn spell_remove(
...
+        vec![json!(key.clone())],
+        Some(params.id),
+        params.init_peer_id,
+        Duration::from_millis(params.ttl as u64),
+    ))
+    .map(|v| json!(v.str))
+    .map_err(|e| JError::new(f!("Failed to get argument {key} for spell {spell_id}: {e}")))
+}
+
+pub(crate) fn error_handler(
```
folex, line 125:
```suggestion
pub(crate) fn store_error(
```

sorcerer/src/utils.rs
```diff
@@ -0,0 +1,56 @@
...
+            if result.is_success() {
+                Ok(result)
+            } else {
+                Err(JError::new(result.get_error()))
+            }
+        }
+    }
+}
+
+pub(crate) fn parse_spell_id_from(particle_id: String) -> eyre::Result<String, JError> {
```
folex, line 45: Why accept `String`? You only need to manipulate indexes (ie slices), so I reckon it should take `&str` and return `Result<&str>`

sorcerer/src/utils.rs
justprosh: ohhh, functionOutcome::Ok has no idea about func and service names

sorcerer/src/utils.rs
justprosh: so i need to pass them to process function because FunctionOutcome::Empty has no idea about it too

particle-builtins/src/builtins.rs
```diff
@@ -56,6 +56,11 @@ use crate::identify::NodeInfo;
 use crate::outcome::{ok, wrap, wrap_unit};
 use crate::{json, math};
 
+pub struct CustomService {
+    pub functions: HashMap<String, Mutex<ServiceFunction>>,
+    pub unhandled: Option<Mutex<ServiceFunction>>,
```
kmd-fl, line 63: Perhaps it would be best to call it `default` or `fallback`, wouldn't it? I think the word `unhandled` might be a little obscure.

particle-builtins/src/builtins.rs
```diff
@@ -143,7 +150,11 @@ where
             .custom_services
             .read()
             .get(&args.service_id)
-            .and_then(|fs| fs.get(&args.function_name))
+            .and_then(|fs| {
+                fs.functions
```
kmd-fl, line 154: Shouldn't it be in `impl CustomService`?

sorcerer/src/spells.rs
```diff
@@ -13,42 +13,65 @@
...
-use serde_json::Value as JValue;
-use serde_json::Value::Array;
 use spell_event_bus::scheduler::api::{SchedulerApi, TimerConfig};
 use spell_storage::SpellStorage;
-use std::time::Duration;
+
+use crate::utils::{parse_spell_id_from, process_func_outcome};
 
 pub(crate) fn spell_install(
+    sargs: Args,
```
kmd-fl, line 30: I'm just curious, why reorder the arguments? I thought the environment arguments should be first and the "variable" arguments should be last.

particle-services/src/app_services.rs
```diff
@@ -335,21 +335,22 @@ impl ParticleAppServices {
 
     pub fn call_function(
         &self,
-        service_id: String,
+        service_id: &str,
```
kmd-fl, line 338: Why should it be a ref?

kmd-fl (approved):

sorcerer/src/spells.rs
justprosh: Every other `builtin` function has args and params first

particle-services/src/app_services.rs
justprosh: because it was handy to me, i can write `call_function("service_id", ...)` without `to_string()`

particle-builtins/src/builtins.rs
justprosh: I was thinking about custom service as data struct, but we can move it later with builtins refactoring

PR #1367, opened 2022-12-12
feat(spell): add full ClockConfig support

kmd-fl:
Config interpretation rules:
* `end_sec = 0`  -- spell execution never ends
* `start_sec = 0` -- clock config considered empty, don't run, don't subscribe
* `period_sec =  0` -- don't subscribe; run once
* `start_sec < now` -- run immediately.

Config is invalid when: 
* `period_sec > 100 years` 
* `end_sec < start_sec`
* `end_sec < now`

if `start_sec` is 0, other fields are ignored.

`now` is the time of config validation on the node.

NET-303


particle-node/tests/spells.rs
```diff
@@ -291,3 +284,211 @@ fn spell_return_test() {
...
+    };
+    client.send_particle(
+        r#"
+        (seq
+            (call relay ("spell" "install") [script data config] spell_id)
+            (call client ("return" "") [spell_id])
+        )"#,
+        data.clone(),
+    );
+    assert!(client.receive_args().is_err());
```
justprosh, line 412: ditto

particle-node/tests/spells.rs
```diff
@@ -291,3 +284,211 @@ fn spell_return_test() {
...
```
justprosh, line 493: ditto

sorcerer/src/spells.rs
```diff
@@ -23,63 +21,9 @@ use fluence_spell_dtos::trigger_config::TriggerConfig;
...
-
-    if triggers.is_empty() {
-        Err(ConfigError::EmptyConfig)
-    } else {
-        Ok(api::SpellTriggerConfigs { triggers })
-    }
-}
+use std::time::Duration;
 
 pub(crate) async fn spell_install(
```
justprosh, line 28: Please, add storing of the trigger config in the spell

crates/spell-event-bus/src/config.rs
```diff
@@ -0,0 +1,115 @@
...
+    InvalidPeriod,
+    #[error("invalid config: end_sec is less than start_sec or in the past")]
+    InvalidEndSec,
+    #[error("config is empty, nothing to do")]
+    EmptyConfig,
+}
+
+// The only way to convert timestamp to instant I found.
+// Fails if the timestamp is in the past or overflow occurred which actually shouldn't happen.
+fn to_instant(timestamp: u64) -> Option<Instant> {
```
justprosh, line 27: You can use [Instant::at](https://docs.rs/datetime/0.4.4/datetime/struct.Instant.html#method.at)

crates/spell-event-bus/src/bus.rs
```diff
@@ -214,11 +225,14 @@ impl SpellEventBus {
...
                         // The timer is triggered only if there are some spells to be awaken.
-                        let scheduled_spell = state.scheduled.pop().expect("billions of years have gone by already?");
-                        Self::trigger_spell(&send_events, &scheduled_spell.data.id, Event::Timer)?;
-                        // We don't expect that timer overflow will happen.
-                        if let Some(rescheduled) = Scheduled::at(scheduled_spell.data, Instant::now()) {
-                            state.scheduled.push(rescheduled);
+                        if let Some(scheduled_spell) = state.scheduled.pop() {
+                            log::trace!("Execute: {:?}", scheduled_spell);
+                            Self::trigger_spell(&send_events, &scheduled_spell.data.id, Event::Timer)?;
+                            // We don't expect that timer overflow will happen.
```
justprosh, line 231: we don't expect but it is better to add `else` with `log::warn`

justprosh (approved):
It will be good to store config in spell in `spell.install` and reschedule spells after restarting [here](https://github.com/fluencelabs/rust-peer/blob/master/sorcerer/src/sorcerer.rs#L85)

crates/spell-event-bus/src/bus.rs
```diff
@@ -59,8 +61,14 @@ struct Scheduled {
 }
 
 impl Scheduled {
+    fn new(data: Periodic, run_at: Instant) -> Self {
+        Self { data, run_at }
+    }
     // schedule to no earlier than now + data.period
     fn at(data: Periodic, now: Instant) -> Option<Scheduled> {
+        if data.end_at.map(|end_at| end_at <= now).unwrap_or(false) {
```
folex, line 69: Add a comment here explaining the condition please

crates/spell-event-bus/src/bus.rs
```diff
@@ -59,8 +61,14 @@ struct Scheduled {
...
+            return None;
```
folex, line 70: eg, does that mean that config is invalid? If yes, maybe return `Err`?

crates/spell-event-bus/src/config.rs
```diff
@@ -0,0 +1,115 @@
...
+}
+
+/// Convert user-friendly config to event-bus-friendly config, validating it in the process.
+pub fn from_user_config(
+    user_config: UserTriggerConfig,
+) -> Result<SpellTriggerConfigs, ConfigError> {
+    let mut triggers = Vec::new();
+    // Process timer config
+    let clock = user_config.clock;
+    if clock.start_sec != 0 {
```
folex, line 40: Add a comment explaining the condition please

crates/spell-event-bus/src/config.rs
```diff
@@ -0,0 +1,115 @@
...
+            Some(start_at)
+        } else {
+            end_at
+        };
+
+        triggers.push(TriggerConfig::Timer(TimerConfig {
+            period: Duration::from_secs(clock.period_sec as u64),
+            start_at,
+            end_at,
+        }));
```
folex, line 70: Maybe move this to a separate function?

crates/spell-event-bus/src/config.rs
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    // Process connections config
+    let mut pool_events = Vec::with_capacity(2);
```
folex, line 74: Why capacity is 2? Can you add a comment explaining that?

crates/spell-event-bus/src/config.rs
```diff
@@ -0,0 +1,115 @@
...
+    // Process timer config
+    let clock = user_config.clock;
+    if clock.start_sec != 0 {
+        if clock.period_sec > MAX_PERIOD_SEC {
+            return Err(ConfigError::InvalidPeriod);
+        }
+
+        let end_at = if clock.end_sec == 0 {
+            None
+        } else if clock.end_sec > clock.start_sec {
```
folex, line 47: please add comments for these conditions

folex (approved):
LGTM. Please make sure that you use doc comments (`///`) on functions and structs, instead of just comments (`//`)

crates/spell-event-bus/src/bus.rs
kmd-fl: `None` means we don't need to reschedule the spell.

crates/spell-event-bus/src/bus.rs
kmd-fl: `None` means we don't need to reschedule the spell.

crates/spell-event-bus/src/bus.rs
kmd-fl: The whole point of introducing the internal invariant that timer overflows can't happen is to avoid this unnecessary error handling that won't affect anyone and won't happen at all.

Maybe here the comment might be a bit misleading because now the `Scheduled::at` returns `None` if the spell shouldn't be rescheduled.

crates/spell-event-bus/src/config.rs
kmd-fl: It's not the same `std::time::Instant` btw, but a different library. Do you know if it's reliable? Need some time to explore.

sorcerer/src/spells.rs
kmd-fl: I think I need to it in a different PR.

particle-node/tests/spells.rs
```diff
@@ -291,3 +284,211 @@ fn spell_return_test() {
...
+            );
+
+            let response = client.receive_args().wrap_err("receive").unwrap();
+            if response[0]["success"].as_bool().unwrap() {
+                counter = response[0]["num"].as_u64().unwrap();
+            }
+        }
+
+        assert!(counter == 0 || counter == 1);
+    }
```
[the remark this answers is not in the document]
kmd-fl: I wasn't sure what's the best solution here. But my reasoning was the following:
1. We can't be sure if a spell will be executed by the time of checking, so it might still be waiting in the queue.
2.  Install and run a spell several times to ensure that the first run wasn't coincidental.

Maybe a better solution is to set some flag in a spell to ensure it was executed and then check that it was executed only once. Right now, we don't have an interface to check which spells are running and which are not, so there is little space for such tests.

crates/spell-event-bus/src/config.rs
justprosh: Sorry, this is not so good lib

particle-node/tests/spells.rs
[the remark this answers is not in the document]
justprosh: For now, your test checks nothing because every assertion can be satisfied with a zero counter. I suggest deploying one and, in for-loop checking, if it was executed. You can add short sleep after spell install

justprosh (approved):

PR #1368, opened 2022-12-13
Add spell rescheduling after restart

kmd-fl:
This can be considered a fast implementation because there are some parts I don't like:
1. Event bus API is available before it can be used because of all API methods, despite being async, wait for the result from the event bus. The result is just a unit that states that the command was processed.
2. Rescheduling spell happens in the `start` methods, merely before the sorcerer starts. Doesn't look good. 

Event bus API is used in sorcerer, so we can't just create it on event-bus start and then pass it to `sorcerer.start(api)`.

So solutions I see are:
1. Start event bus before sorcerer creation in `Node::new`. Doesn't look like a good place for spawning processes, though. 
2. Make rescheduling spells an explicit separate step and call it after the event bus and sorcerer start (somehow preserving all necessary fields of the latter). 
3. Do something else? Maybe we don't want to execute particles before the node is fully initialized for reasons I'm not yet aware of and pursue another approach.

### DIFF
--- a/particle-node/src/node.rs
+++ b/particle-node/src/node.rs
@@ -227,8 +227,6 @@ impl<RT: AquaRuntime> Node<RT> {
             .map(|x| PeerEvent::ConnectionPool(x))
             .boxed()];
 
-        // Unfortunately, API is available before the bus is running. It's used for creating spell builtins.
-        // That means that calling api before the bus is running will block execution.
         let (spell_event_bus, spell_event_bus_api, spell_events_stream) =
             SpellEventBus::new(sources);
 
@@ -394,7 +392,6 @@ impl<RT: AquaRuntime> Node<RT> {
 
             let services_metrics_backend = services_metrics_backend.start();
             let script_storage = script_storage.start();
-            // Spell bus must be run before sorcerer, because sorcerer may send events to the bus
             let spell_event_bus = spell_event_bus.start();
             let sorcerer = sorcerer.start(spell_events_stream);
             let pool = aquavm_pool.start();
--- a/particle-node/src/node.rs
+++ b/particle-node/src/node.rs
@@ -226,6 +226,9 @@ impl<RT: AquaRuntime> Node<RT> {
         let sources = vec![recv_connection_pool_events
             .map(|x| PeerEvent::ConnectionPool(x))
             .boxed()];
+
+        // Unfortunately, API is available before the bus is running. It's used for creating spell builtins.
+        // That means that calling api before the bus is running will block execution.
         let (spell_event_bus, spell_event_bus_api, spell_events_stream) =
             SpellEventBus::new(sources);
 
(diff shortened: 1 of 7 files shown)


sorcerer/src/sorcerer.rs
```diff
@@ -81,13 +85,52 @@ impl Sorcerer {
...
+                        "get_trigger_config",
+                        vec![],
+                        None,
+                        self.node_peer_id,
+                        self.spell_script_particle_ttl,
+                    ),
+                    &spell_id,
+                    "get_trigger_config",
+                )?;
+                if !result.success {
```
justprosh, line 108: It has already checked in [here](https://github.com/fluencelabs/rust-peer/blob/master/sorcerer/src/utils.rs#L44), you can just append error message

justprosh (commented):
In addition to my comment, I noticed that I made a mistake in `process_func_outcome`.
Can you, please, move [this](https://github.com/fluencelabs/rust-peer/blob/master/sorcerer/src/utils.rs#L48) parse error to [map_err here](https://github.com/fluencelabs/rust-peer/blob/master/sorcerer/src/utils.rs#L48) and return only `result.get_error()` instead

sorcerer/src/sorcerer.rs
```diff
@@ -81,13 +85,43 @@ impl Sorcerer {
...
+        }
+    }
+
     pub fn start(self, spell_events_stream: Inlet<TriggerEvent>) -> JoinHandle<()> {
+        // Reschedule spells here because at this point the event bus is ready to accept subscriptions.
+        // TODO: come up with better solution.
+        task::block_on(self.reschedule_spells());
+
         spawn(async {
             spell_events_stream
```
folex, line 126:
```suggestion
        spawn(async {
            self.reschedule_spells().await;
            spell_events_stream
```

particle-node/src/node.rs
```diff
@@ -226,6 +226,9 @@ impl<RT: AquaRuntime> Node<RT> {
         let sources = vec![recv_connection_pool_events
             .map(|x| PeerEvent::ConnectionPool(x))
             .boxed()];
+
+        // Unfortunately, API is available before the bus is running. It's used for creating spell builtins.
+        // That means that calling api before the bus is running will block execution.
```
folex, line 231: fix?

particle-node/src/node.rs
```diff
@@ -391,6 +394,7 @@ impl<RT: AquaRuntime> Node<RT> {
 
             let services_metrics_backend = services_metrics_backend.start();
             let script_storage = script_storage.start();
+            // Spell bus must be run before sorcerer, because sorcerer may send events to the bus
```
folex, line 397: fix

folex (approved):
Let's remove block_on and fix comments and good to go

PR #1369, opened 2022-12-13
Enable spell removal api

kmd-fl:
Now I see only a hack way to implement spells removal. And I think we need to enable it.

We need to somehow detect whether the given id is a spell, but now I don't see how we can share `spell_storage` info with `builtins` to do it the right way.


justprosh (commented):
LGTM, but it will be good to add tests for spell removal via `remove_service` and vice-versa

folex (approved):
Hackity hack!

PR #1376, opened 2022-12-20
chore(spell): add tests for spells

kmd-fl:
* unit tests for spell event bus
* test for connection-pool events

### DIFF
--- a/crates/spell-event-bus/src/bus.rs
+++ b/crates/spell-event-bus/src/bus.rs
@@ -253,7 +253,7 @@ impl SpellEventBus {
     ) -> Result<(), BusInternalError> {
         send_events
             .unbounded_send(TriggerEvent {
-                id: (**id).clone(),
+                spell_id: (**id).clone(),
                 event: event.clone(),
             })
             .map_err(|e| BusInternalError::SendEvent((**id).clone(), event, e.into_send_error()))?;
@@ -268,12 +268,13 @@ mod tests {
     use connection_pool::LifecycleEvent;
     use futures::StreamExt;
     use libp2p::PeerId;
+    use maplit::hashmap;
     use particle_protocol::Contact;
     use std::assert_matches::assert_matches;
     use std::time::Duration;
 
     // Safely call teardown after test.
-    fn save_assert<T>(test: T, teardown: impl FnOnce())
+    fn try_catch<T>(test: T, teardown: impl FnOnce())
     where
         T: FnOnce() + std::panic::UnwindSafe,
     {
@@ -286,6 +287,13 @@ mod tests {
         }
     }
 
+    fn send_connect_event(send: &Outlet<PeerEvent>, peer_id: PeerId) {
+        send.unbounded_send(PeerEvent::ConnectionPool(LifecycleEvent::Connected(
+            Contact::new(peer_id, Vec::new()),
+        )))
+        .unwrap();
+    }
+
     fn emulate_connect(period: Duration) -> (Inlet<PeerEvent>, JoinHandle<()>) {
         let (send, recv) = unbounded();
         let hdl = task::spawn(async move {
@@ -293,7 +301,7 @@ mod tests {
             loop {
                 select! {
                     _ = interval.select_next_some() => {
-                        send.unbounded_send(PeerEvent::ConnectionPool(LifecycleEvent::Connected(Contact::new(PeerId::random(), Vec::new())))).unwrap();
+                        send_connect_event(&send, PeerId::random());
                     }
                 }
             }
@@ -311,28 +319,22 @@ mod tests {
         .unwrap();
     }
 
-    fn subscribe_spell_timer(
-        api: &SpellEventBusApi,
-        spell_id: SpellId,
-        period: Duration,
-        start_at: Instant,
-        end_at: Option<Instant>,
-    ) {
+    fn subscribe_timer(api: &SpellEventBusApi, spell_id: SpellId, config: TimerConfig) {
         task::block_on(api.subscribe(
             spell_id,
             SpellTriggerConfigs {
-                triggers: vec![TriggerConfig::Timer(TimerConfig {
-                    period,
-                    start_at,
-                    end_at,
-                })],
+                triggers: vec![TriggerConfig::Timer(config)],
             },
         ))
         .unwrap();
     }
 
-    fn subscribe_periodic_spell(api: &SpellEventBusApi, spell_id: SpellId, period: Duration) {
-        subscribe_spell_timer(api, spell_id, period, Instant::now(), None)
+    fn subscribe_periodic_endless(api: &SpellEventBusApi, spell_id: SpellId, period: Duration) {
+        subscribe_timer(
+            api,
+            spell_id,
+            TimerConfig::periodic(period, Instant::now(), None),
+        );
     }
 
     #[test]
@@ -341,15 +343,15 @@ mod tests {
         let bus = bus.start();
 
         let spell1_id = "spell1".to_string();
-        subscribe_periodic_spell(&api, spell1_id.clone(), Duration::from_millis(5));
+        subscribe_periodic_endless(&api, spell1_id.clone(), Duration::from_millis(5));
 
         let events =
             task::block_on(async { event_stream.take(5).collect::<Vec<TriggerEvent>>().await });
-        save_assert(
+        try_catch(
             || {
                 assert_eq!(events.len(), 5);
                 for event in events.into_iter() {
-                    assert_eq!(event.id, spell1_id.clone(),);
+                    assert_eq!(event.spell_id, spell1_id.clone(),);
                     assert_matches!(event.event, Event::Timer);
                 }
             },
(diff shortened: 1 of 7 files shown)


crates/spell-event-bus/src/bus.rs
```diff
@@ -264,55 +264,279 @@ impl SpellEventBus {
...
+        let spell1_id = "spell1".to_string();
+        subscribe_periodic_spell(&api, spell1_id.clone(), Duration::from_millis(5));
+
+        let events =
+            task::block_on(async { event_stream.take(5).collect::<Vec<TriggerEvent>>().await });
+        save_assert(
+            || {
+                assert_eq!(events.len(), 5);
+                for event in events.into_iter() {
+                    assert_eq!(event.id, spell1_id.clone(),);
```
folex, line 352: I guess it should be named `event.spell_id`. Otherwise it's hard to grasp why ID of the Event should be equal to `spell_id`

crates/spell-event-bus/src/bus.rs
```diff
@@ -264,55 +264,279 @@ impl SpellEventBus {
...
+
+        let events =
+            task::block_on(async { event_stream.take(10).collect::<Vec<TriggerEvent>>().await });
+        save_assert(
+            || {
+                assert_eq!(events.len(), 10);
+                for event in events.into_iter() {
+                    assert!(spell_ids.contains(&event.id));
+                    assert_matches!(event.event, Event::Timer);
+                }
```
folex, line 386: I guess you should check for completeness. ie, were there events for all 3 spells, or only for a single one?

crates/spell-event-bus/src/bus.rs
```diff
@@ -264,55 +264,279 @@ impl SpellEventBus {
...
+        let (bus, api, event_stream) = SpellEventBus::new(vec![]);
+        let bus = bus.start();
+        let spell1_id = "spell1".to_string();
+        subscribe_spell_timer(
+            &api,
+            spell1_id.clone(),
+            Duration::from_secs(0),
+            Instant::now(),
+            Some(Instant::now()),
+        );
```
folex, line 407: Please write comments explaining why these values mean "oneshot"

crates/spell-event-bus/src/bus.rs
```diff
@@ -264,55 +264,279 @@ impl SpellEventBus {
...
+        let events =
+            task::block_on(async { event_stream.take(5).collect::<Vec<TriggerEvent>>().await });
+
+        let mut counts = HashMap::new();
+        counts.insert(spell1_id.clone(), 0);
+        counts.insert(spell2_id.clone(), 0);
+        for event in events.into_iter() {
+            counts.entry(event.id).and_modify(|e| *e += 1);
+        }
+        save_assert(
```
folex, line 420: `save_assert` confuses me. Sounds like you're trying to save (persist) assert.

crates/spell-event-bus/src/bus.rs
```diff
@@ -264,55 +264,279 @@ impl SpellEventBus {
...
+        let (bus, api, event_stream) = SpellEventBus::new(vec![]);
+        let bus = bus.start();
+
+        let spell1_id = "spell1".to_string();
+        subscribe_periodic_spell(&api, spell1_id.clone(), Duration::from_millis(5));
+
+        let spell2_id = "spell2".to_string();
+        // Period is high enough to avoid triggering the spell.
+        let spell2_period = Duration::from_secs(10);
+        subscribe_spell_timer(
```
folex, line 445: what's the difference between `subscribe_periodic_spell` and `subscribe_spell_timer`?

crates/spell-event-bus/src/bus.rs
```diff
@@ -264,55 +264,279 @@ impl SpellEventBus {
...
+            &api,
+            spell2_id.clone(),
+            spell2_period,
+            Instant::now() + spell2_period,
+            None,
+        );
 
         // let's remove spell2
         task::block_on(async { api.unsubscribe(spell2_id).await }).unwrap();
```
folex, line 454: Would test pass if you remove these lines? If yes, then it doesn't really test `unsubscribe`.

crates/spell-event-bus/src/bus.rs
```diff
@@ -264,55 +264,279 @@ impl SpellEventBus {
...
+                    bus.cancel().await;
+                });
+            },
+        );
+    }
+
+    #[test]
+    fn test_subscribe_connect() {
+        let (recv, hdl) = emulate_connect(Duration::from_millis(5));
+        let (bus, api, event_stream) = SpellEventBus::new(vec![recv.boxed()]);
```
folex, line 479: Why does `SpellEventBus::new` accept `Vec` of sources? I would expect it to accept only a single source, the one from `ConnectionPool`.

Do you foresee more sources?

crates/spell-event-bus/src/bus.rs
```diff
@@ -264,55 +264,279 @@ impl SpellEventBus {
...
+        let events =
+            task::block_on(async { event_stream.take(10).collect::<Vec<TriggerEvent>>().await });
+        save_assert(
+            || {
+                for event in events.into_iter() {
+                    assert!(event.id == spell1_id || event.id == spell2_id);
+                    if event.id == spell1_id {
+                        assert_matches!(
+                            event.event,
+                            Event::Peer(PeerEvent::ConnectionPool(LifecycleEvent::Connected(_)))
```
folex, line 527: Could you please also check that PeerId of the of the Connected event is as you expect?

particle-node/tests/spells.rs
```diff
@@ -58,10 +58,7 @@ fn create_spell(
 
 #[test]
 fn spell_simple_test() {
-    let swarms = make_swarms_with_cfg(1, |mut cfg| {
-        cfg.timer_resolution = Duration::from_millis(20);
-        cfg
-    });
+    let swarms = make_swarms(1);
```
folex, line 61: How come you don't need `timer_resolution` anymore?

particle-node/tests/spells.rs
```diff
@@ -706,3 +666,73 @@ fn spell_remove_service_as_spell() {
...
+        .unwrap();
+
+    let script = r#"(call %init_peer_id% ("peer" "identity") [] x)"#;
+    let mut config = TriggerConfig::default();
+    config.connections.connect = true;
+    let spell_id1 = create_spell(&mut client, script.clone(), config.clone(), hashmap! {});
+
+    let mut config = TriggerConfig::default();
+    config.connections.disconnect = true;
+    let spell_id2 = create_spell(&mut client, script, config.clone(), hashmap! {});
```
folex, line 684: unnecessary `config.clone`?

particle-node/tests/spells.rs
```diff
@@ -706,3 +666,73 @@ fn spell_remove_service_as_spell() {
...
+
+    // This connect should trigger the spell
+    let connect_num = 5;
+    for _ in 0..connect_num {
+        ConnectedClient::connect_to(swarms[0].multiaddr.clone()).unwrap();
+        std::thread::sleep(std::time::Duration::from_millis(100));
```
folex, line 690: Why do you need `sleep` here?

particle-node/tests/spells.rs
```diff
@@ -706,3 +666,73 @@ fn spell_remove_service_as_spell() {
...
+    if let [result] = client
+        .receive_args()
+        .wrap_err("receive")
+        .unwrap()
+        .as_slice()
+    {
+        let counter = result["num"].as_u64().unwrap();
+        assert_eq!(counter, connect_num, "disconnected");
+    }
+}
```
folex, line 738: nice test!

I think there should also be a test where spell sends trigger info to some client.

Like
```
let first_client = ...;
let spell_script = "(call {first_client.peer_id} ("op" "return") [trigger_info])";

// and then the same code as you have
// but instead of checking `counter`, you would catch events like
first_client.listen_for_n(...)
```

particle-node/tests/spells.rs
justprosh: In tests, the default time resolution is still used. I added a custom resolution previously to pass faster, but now it's not needed, because we can specify start time in config

crates/spell-event-bus/src/bus.rs
justprosh: And it's not an assert either, and maybe it can be called smth like `try_catch` or `panic_catch` etc

particle-node/tests/spells.rs
kmd-fl: I'm not sure where `timer_resolution` is used (I saw only in script-storage), but it's not in the spell event bus as we discussed earlier in slack, so I thought it should be deleted.

crates/spell-event-bus/src/bus.rs
kmd-fl: I wonder, should we check `end_sec` if the period is `0` and the spell considered oneshot?

crates/spell-event-bus/src/bus.rs
kmd-fl: I thought we would need blockchain triggers, HTTP triggers.

crates/spell-event-bus/src/bus.rs
kmd-fl: `periodic` is a specialized version of `spell_timer`. Just to avoid lots of text that distracts attention. I'll refactor it a bit for clarity.

crates/spell-event-bus/src/bus.rs
kmd-fl: Yeah, I think this test doesn't test actual unsubscribing. Need to rewrite it.

particle-node/tests/spells.rs
kmd-fl: What's trigger info?

crates/spell-event-bus/src/bus.rs
folex: makes sense

particle-node/tests/spells.rs
folex: Trigger Info is the information about what trigger has caused spell run. It should be in `trigger_mailbox` or something like that.

folex: It seems that we don't have that info anywhere yet. So you can just read something from KV (eg counter) and send it to client.

kmd-fl: Also, I thought we decided to implement this test separately?

folex: Yes, we did! Just was answering the question :)

That test better go in a separate PR.

particle-node/tests/spells.rs
folex: So it seems that `timer_resolution` is now used only by Scheduled Scripts, and never by Spells. So all is good!

particle-node/tests/spells.rs
folex: Okay, I see why. Client connects to a peer, and then `get_u32` script is called immediately. Concurrently, a spell script is executed in reaction to `Connected` event.

Instead of sleep, I see three options:
1. Poll `get_u32` until it gives you desired result. Bad option: code will be complex and hard to edit.
2. Subscribe on `LifecycleEvent` from `ConnectionPool`. Will require to add `ConnectionPoolApi` to `CreatedSwarm`. okay-ish.
3. Instead of Pull approach, do Push approach. Spell should send `counter` to `client` every time a `Connected` event happens. I think this is the best option, because Push is faster and thus suits the task better.

crates/spell-event-bus/src/bus.rs
kmd-fl: Did that in another test.

folex (approved):

PR #1381, opened 2022-12-28
feat(spells): add spell config update [NET-234 NET-315]

kmd-fl:


sorcerer/src/spells.rs
```diff
@@ -143,6 +143,52 @@ pub(crate) async fn spell_remove(
...
+) -> Result<(), JError> {
+    let mut args = args.function_args.into_iter();
+    let spell_id: String = Args::next("spell_id", &mut args)?;
+    let user_config: TriggerConfig = Args::next("config", &mut args)?;
+    let config = api::from_user_config(user_config.clone())?;
+
+    // TODO: proper right management
+
+    // Can't really do proper right management here, so for now this call does the job.
+    // It fails for everyone who's not a spell owner, so only the spell owner can update spell's config.
```
folex, line 161: aaand the spell itself, right?

sorcerer/src/spells.rs
```diff
@@ -143,6 +143,52 @@ pub(crate) async fn spell_remove(
...
+    params: ParticleParams,
+    services: ParticleAppServices,
+    spell_event_bus_api: SpellEventBusApi,
+) -> Result<(), JError> {
+    let mut args = args.function_args.into_iter();
+    let spell_id: String = Args::next("spell_id", &mut args)?;
+    let user_config: TriggerConfig = Args::next("config", &mut args)?;
+    let config = api::from_user_config(user_config.clone())?;
+
+    // TODO: proper right management
```
folex, line 158: right => permission

folex: also would be nice to explain what you want to achieve with permissions here

sorcerer/src/spells.rs
```diff
@@ -143,6 +143,52 @@ pub(crate) async fn spell_remove(
...
+// update_config(spell_id, config)
+pub(crate) async fn spell_update_config(
+    args: Args,
+    params: ParticleParams,
+    services: ParticleAppServices,
+    spell_event_bus_api: SpellEventBusApi,
+) -> Result<(), JError> {
+    let mut args = args.function_args.into_iter();
+    let spell_id: String = Args::next("spell_id", &mut args)?;
+    let user_config: TriggerConfig = Args::next("config", &mut args)?;
```
folex, line 155: I was expecting TriggerConfig to be read from spell service. But this is feasible as well, not sure of consequences

crates/spell-event-bus/src/bus.rs
```diff
@@ -126,10 +131,20 @@ impl SubscribersState {
...
 
-    fn unsubscribe(&mut self, spell_id: &SpellId) {
+    /// Returns true if spell_id was removed from subscribers
+    fn unsubscribe(&mut self, spell_id: &SpellId) -> bool {
+        let prev_len = self.scheduled.len();
         self.scheduled
             .retain(|scheduled| *scheduled.data.id != *spell_id);
-        self.subscribers.remove(spell_id);
+        let new_len = self.scheduled.len();
+        let removed = self.subscribers.remove(spell_id);
```
folex, line 140:
```suggestion
        let removed = self.subscribers.remove(spell_id);
        debug_assert!(removed == (new_len < prev_len))
```

crates/spell-event-bus/src/bus.rs
```diff
@@ -126,10 +131,20 @@ impl SubscribersState {
...
+        prev_len != new_len || removed
+    }
+
+    fn update(&mut self, spell_id: &SpellId, config: &SpellTriggerConfigs) {
+        if self.unsubscribe(spell_id) {
+            self.subscribe(spell_id.clone(), config);
```
folex, line 146: you might miss some events this way 🤔

crates/spell-event-bus/src/bus.rs
```diff
@@ -294,6 +312,13 @@ mod tests {
         .unwrap();
     }
 
+    fn send_disconnect_event(send: &Outlet<PeerEvent>, peer_id: PeerId) {
```
folex, line 315: #[cfg(test)]?

sorcerer/src/spells.rs
```diff
@@ -143,6 +143,52 @@ pub(crate) async fn spell_remove(
...
+        &spell_id,
+        "set_trigger_config",
+    )?;
+
+    if let Err(err) = spell_event_bus_api
+        .update_config(spell_id.clone(), config)
+        .await
+    {
+        log::warn!(
+            "can't update config of the spell {} from its triggers via spell-event-bus-api: {}; the new config is saved",
```
folex, line 180: > the new config is saved

what does that mean?

sorcerer/src/spells.rs
kmd-fl: I don't know, really. Because the check is in the spell service itself, and it's `is_creator`, I think.

sorcerer/src/spells.rs
kmd-fl: Well, I think it's not good to rely on the spell service to check if the update request is allowed.

sorcerer/src/spells.rs
kmd-fl: It means that at this point of execution, we successfully saved the new config to the spell service, but we couldn't resubscribe the spell due to some internal errors.

sorcerer/src/spells.rs
kmd-fl: What do you mean? I thought the updating process is:
1. Read and validate the new config from the user's args.
2. Save the config to the spell service.
3. Resubscribe the spell to new triggers from the config.

Do you mean that after saving the config, we need to read it again from the service?

folex: i was thinking that builtin would trigger a reread from spell service

like we do after restart, same thing

crates/spell-event-bus/src/bus.rs
kmd-fl: Why? The bus is single-threaded, and events are stored in the input channel.

sorcerer/src/spells.rs
folex: write a comment pls

sorcerer/src/spells.rs
folex: so that should be changed then, spell should allow to be self updated

crates/spell-event-bus/src/bus.rs
```diff
@@ -39,10 +39,15 @@ impl Subscribers {
             .unwrap_or_else(|| [].iter())
     }
 
-    fn remove(&mut self, spell_id: &SpellId) {
+    /// Returns true spell_id was removed from subscribers
```
justprosh, line 42:
```suggestion
    /// Returns true if spell_id was removed from subscribers
```

sorcerer/src/spells.rs
justprosh: I like the way that Maria suggested. So spell service can only receive trigger config updates only from spell keypair and via `update_config` builtin

sorcerer/src/spells.rs
justprosh: In my PR there is a function that returns the same associated keypair (and peer id) for the spell owner and the spell itself

sorcerer/src/spells.rs
folex: spell keypair is not creator

justprosh: And who's service creator?

crates/spell-event-bus/src/bus.rs
kmd-fl: Why is it inconsistent? Here `scheduled` contains info about the spells that are scheduled and `subscribers` -- for the spells that are subscribed to peer events. It's two different lists because these spells are treated differently.

sorcerer/src/spells.rs
folex: then reword to clarify:
```
save config to spell service SUCCESS
update trigger subscriptions FAIL: {}

call set_trigger_config to try again
```

kmd-fl: Both for users and as internal error?

folex: yes. but for logs use single line

kmd-fl: Just to clarify, not `set_trigger_config` but `update_config`, because `set_trigger_config` just updates the config in the spell service, it doesn't resubscribe.

folex: yes. btw, sounds like it should be `update_trigger_config`, wdyt?

kmd-fl: It's used as `SpellBuiltin.update_config`. Not sure if there are or will be any other types of configs, no?

folex: we might have different configs in the future. also it makes sense to have consistency between set_trigger_config and update_trigger_config

kmd-fl: I updated the name.

Although, the `set_trigger_config` function is the one of a spell service and `update_trigger_config` is the one of a node. I think we don't need to keep their naming consistency because of their different classes.

folex (approved):

PR #1385, opened 2023-01-11
chore: update spell-dtos to 0.1.4

folex:


sorcerer/src/utils.rs
```diff
@@ -54,7 +54,7 @@ where
             } else {
                 Err(JError::new(format!(
                     "Function {spell_id}.{function_name} executed with error: {}",
-                    result.get_error()
```
kmd-fl, line 57: I wonder, is it a common convention to use `take` to mark that the value is consumed? I thought the `into` prefix is used in rust for. `take` is definitely better than `get`, I just don't understand if it's our invention or not.

folex: more or less https://doc.rust-lang.org/std/option/enum.Option.html?search=take

usually I guess it's `into`, eg `into_error`. But here it sounds weird.

monoid: In Rust, `take` is used for an operation that replaces inner state with some default value, returning the original state (i.e. it is non-consuming).

justprosh (approved):

particle-node/tests/spells.rs
```diff
@@ -755,6 +756,8 @@ fn spell_trigger_connection_pool() {
 
 #[test]
 fn spell_update_config() {
+    enable_logs();
+
```
justprosh, line 760:
```suggestion
```

particle-node/tests/spells.rs
```diff
@@ -813,3 +821,68 @@ fn spell_update_config() {
         assert_eq!(x, "called", "spell must be triggered after config update");
     }
 }
+
+#[test]
+fn spell_set_u32() {
+    enable_logs();
+
```
justprosh, line 828:
```suggestion
```

particle-node/tests/spells.rs
```diff
@@ -18,6 +18,7 @@ use created_swarm::make_swarms;
 use eyre::Context;
 
 use fluence_spell_dtos::trigger_config::TriggerConfig;
+use log_utils::enable_logs;
```
justprosh, line 21:
```suggestion
```

justprosh (changes requested):

spell-storage/Cargo.toml
```diff
@@ -10,9 +10,10 @@ particle-modules = { path = "../particle-modules" }
...
 
 fluence-app-service = { workspace = true }
-fluence-spell-distro = "0.1.1"
+fluence-spell-distro = { workspace = true }
 
 parking_lot = { workspace = true }
 derivative = { workspace = true }
 eyre = { workspace = true }
 toml = { workspace = true }
+log = { workspace = true }
```
justprosh, line 19:
```suggestion
log = { workspace = true }

```

justprosh (approved):

PR #1390, opened 2023-01-13
chore: Move node tests to a separate crate

nahsi:


crates/builtins-tests/Cargo.toml
```diff
@@ -0,0 +1,29 @@
+[package]
+name = "builtins"
```
folex, line 2: I think it's better to name it `builtins-tests`

folex (approved):

PR #1430, opened 2023-01-25
fix(spells): Allow parallel execution of immut service functions [NET-331]

kmd-fl:
Tested: run 10 parallel `spell.install`s, and they worked fine, all spells were created quite fast.


gurinderu (approved):

crates/particle-node-tests/tests/client_api.rs
```diff
@@ -40,20 +40,19 @@ fn call_custom_service() {
...
-    let closure: Box<
-        dyn FnMut(_, _) -> BoxFuture<'static, FunctionOutcome> + 'static + Send + Sync,
-    > = Box::new(move |args, params| {
-        async move {
-            println!("got call!!! {args:?} {params:?}");
-            FunctionOutcome::Ok(json!("hello!"))
-        }
-        .boxed()
-    });
+    let closure: Box<dyn Fn(_, _) -> BoxFuture<'static, FunctionOutcome> + 'static + Send + Sync> =
```
folex, line 43: Maybe keep this one FnMut just to make sure it works?

Might as well add some mutable state to check that it's working :) Like
```
let mut args = vec![];
...
args.push(args);
FunctionOutcome::Ok(json!(format!("hello! {} args so far", args.len())));
```

folex (approved):

crates/particle-node-tests/tests/client_api.rs
kmd-fl: It looks like we already have such an example below 
https://github.com/fluencelabs/rust-peer/blob/4546b0f69e5d0d5132c6679684be7244cd72c085/crates/particle-node-tests/tests/client_api.rs#L57-L70

folex: awesome!

PR #1433, opened 2023-01-26
feat(spells): do json manipulations internally

justprosh:
- spell's `getDataSrv` returns parsed json value, so there is no need to parse it manually
- `spell.install` now accepts any object as `init_data` (previously, it was accepting only stringified ones)
- spell's `response` accepts objects (previously, it was accepting only stringified ones)

### DIFF
--- a/sorcerer/src/spells.rs
+++ b/sorcerer/src/spells.rs
@@ -37,7 +37,7 @@ pub(crate) async fn spell_install(
 ) -> Result<JValue, JError> {
     let mut args = sargs.function_args.clone().into_iter();
     let script: String = Args::next("script", &mut args)?;
-    let init_data: String = Args::next("data", &mut args)?;
+    let init_data: JValue = Args::next("data", &mut args)?;
     let user_config: TriggerConfig = Args::next("config", &mut args)?;
     let config = api::from_user_config(user_config.clone())?;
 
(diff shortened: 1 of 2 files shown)


sorcerer/src/spells.rs
```diff
@@ -66,7 +66,7 @@ pub(crate) async fn spell_install(
         services.call_function(
             &spell_id,
             "set_json_fields",
-            vec![json!(init_data)],
+            vec![json!(init_data.to_string())],
```
folex, line 69: shouldn't it be enough to pass `init_data.to_string()`? If not, why not?

sorcerer/src/spells.rs
```diff
@@ -262,12 +263,12 @@ pub(crate) fn store_response(
...
 ) -> Result<(), JError> {
     let spell_id = parse_spell_id_from(&params.id)?;
-
+    let response: JValue = Args::next("spell_id", &mut args.function_args.into_iter())?;
     process_func_outcome::<UnitValue>(
         services.call_function(
             spell_id,
             "set_json_fields",
-            args.function_args.clone(),
+            vec![json!(response.to_string())],
```
folex, line 271: same thing, why do you serialize it to json twice?

sorcerer/src/spells.rs
justprosh: call_function accepts `function_args: Vec<JValue>`

sorcerer/src/spells.rs
folex: `response.to_string()` seems to satisfy that description

justprosh: yeah, but `call_function` accepts `function_args: Vec<JValue>`

folex (approved):

PR #1434, opened 2023-01-27
feat(builtins): json.obj_pairs, json.puts_pairs to build JSON objects from array of pairs

folex:
`json.obj` and `json.puts` are basically varargs, which gives flexibility, but is impossible to express in well-typed Aqua.

`json.obj` allows to build JSON object from an array of `(string, json)` pairs.

### DIFF
--- a/particle-builtins/src/builtins.rs
+++ b/particle-builtins/src/builtins.rs
@@ -256,8 +256,8 @@ where
             ("json", "puts")       => wrap(json::puts(args)),
             ("json", "parse")      => unary(args, |s: String| -> R<JValue, _> { json::parse(&s) }),
             ("json", "stringify")  => unary(args, |v: JValue| -> R<String, _> { Ok(json::stringify(v)) }),
-            ("json", "obj_array") => unary(args, |vs: Vec<JValue>| -> R<JValue, _> { json::obj_from_array(vs) }),
-            ("json", "puts_array") => binary(args, |obj: JValue, vs: Vec<JValue>| -> R<JValue, _> { json::puts_from_array(obj, vs) }),
+            ("json", "obj_pairs") => unary(args, |vs: Vec<(String, JValue)>| -> R<JValue, _> { json::obj_from_pairs(vs) }),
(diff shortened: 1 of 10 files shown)


particle-builtins/src/builtins.rs
```diff
@@ -255,6 +255,10 @@ where
             ("json", "puts")       => wrap(json::puts(args)),
             ("json", "parse")      => unary(args, |s: String| -> R<JValue, _> { json::parse(&s) }),
             ("json", "stringify")  => unary(args, |v: JValue| -> R<String, _> { Ok(json::stringify(v)) }),
+            ("json", "obj_pairs")  => unary(args, |vs: Vec<(String, JValue)>| -> R<JValue, _> { json::obj_from_pairs(vs) }),
+            ("json", "puts_pairs") => binary(args, |obj: JValue, vs: Vec<(String, JValue)>| -> R<JValue, _> { json::puts_from_pairs(obj, vs) }),
+
+            ("run-console", "print") => wrap_unit(Ok(log::debug!(target: "run-console", "{}", json!(args.function_args)))),
```
kmd-fl, line 261: Do you forget to remove it? Afaik we always remove such builtins after debugging.

folex: nope, it's for support of `run-console` in Aqua. `fluence-js` supports it, don't see a reason why `rust-peer` shouldn't

kmd-fl: Maybe some wrongdoer could try to pollute our logs even more? Could it make more sense to allow `run-console.print` only for management_id?

folex: That wrongdoer would have to enable `debug` for `run-console`, eg pass variable `RUST_LOG=run-console=debug`. 

I think that's a good-enough protection. WDYT?

kmd-fl: Yeah, I didn't notice, sry

folex: no worries! it's good to raise concerns ^_^

justprosh (approved):
lgtm, but it will be nice to have these new methods called at least once in the tests

PR #1452, opened 2023-02-10
fix(spells): do not run spells with passed (ended) Clock trigger

kmd-fl:
Right now _all_ spells will be rerun after the node restart, even the ones that ended before restarting. 

So the new rule is:  _the spell won't be restarted if in its trigger config `end_at <= now` or it's a one-shot spell that started in the past_.

Not sure about the following events:
1. The spell is a one-shot that should be run immediately or was running, but the node was restarted during the process.
2. The periodic spell with `end_at` was added, then the node was shut down until `end_at`, so the spell was never executed.

These events don't seem to be very relevant though.

### DIFF
--- a/sorcerer/src/sorcerer.rs
+++ b/sorcerer/src/sorcerer.rs
@@ -119,10 +119,12 @@ impl Sorcerer {
                     "get_trigger_config",
                 )?;
                 let config = from_user_config(result.config)?;
-                if let Some(config) = config {
+                if let Some(config) = config.and_then(|c| c.into_rescheduled()) {
                     self.spell_event_bus_api
                         .subscribe(spell_id.clone(), config.clone())
                         .await?;
+                } else {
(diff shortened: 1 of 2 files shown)


folex (approved):

sorcerer/src/sorcerer.rs
```diff
@@ -119,10 +119,12 @@ impl Sorcerer {
                     "get_trigger_config",
                 )?;
                 let config = from_user_config(result.config)?;
-                if let Some(config) = config {
+                if let Some(config) = config.and_then(|c| c.into_rescheduled()) {
                     self.spell_event_bus_api
                         .subscribe(spell_id.clone(), config.clone())
                         .await?;
+                } else {
+                    log::warn!("Spell {spell_id} is not rescheduled since its config is either not found or not reschedulable");
```
justprosh, line 127: I think it should be `info`, not `warn`

justprosh (approved):

PR #1477, opened 2023-02-21
fix(spells): allow global aliases for builtin spells

justprosh:
+  fix spell.install for global scope

### DIFF
--- a/particle-services/src/app_services.rs
+++ b/particle-services/src/app_services.rs
@@ -56,17 +56,26 @@ pub struct Service {
     pub service: Mutex<AppService>,
     pub blueprint_id: String,
     pub owner_id: PeerId,
-    pub aliases: Vec<ServiceAlias>,
+    pub worker_aliases: Vec<ServiceAlias>,
     pub worker_id: PeerId,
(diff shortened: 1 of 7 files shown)


sorcerer/src/spells.rs
```diff
@@ -42,7 +42,12 @@ pub(crate) async fn spell_install(
     let config = api::from_user_config(user_config.clone())?;
 
     let worker_id = if key_manager.is_host(params.host_id) {
```
folex, line 44: all these ifs gonna hit us hard :)

particle-services/src/app_services.rs
```diff
@@ -487,7 +487,10 @@ impl ParticleAppServices {
             .get_mut(&service_id)
             .ok_or_else(|| ServiceError::NoSuchService(service_id.clone()))?;
 
-        if service.worker_id != worker_id {
+        if service.worker_id != worker_id
+            && init_peer_id != self.management_peer_id
+            && init_peer_id != self.builtins_management_peer_id
```
folex, line 492: Can you write a small comment on that? And maybe put conditions into 3 variables so it's more readable.

folex (approved):

PR #1486, opened 2023-02-23
fix: builtin redeploy and associated aliases bug

justprosh:


particle-services/src/app_services.rs
```diff
@@ -785,14 +787,14 @@ impl ParticleAppServices {
         })?;
         let stats = service.module_memory_stats();
         let stats = ServiceMemoryStat::new(&stats);
-        let service_type = ServiceType::Service(worker_aliases.first().cloned());
+        let service_type = ServiceType::Service(root_aliases.first().cloned());
```
kmd-fl, line 790: That explains what I saw in grafana! I was very confused about why there was no registry and so one.

kmd-fl: Is the decider in root aliases? It would be good to monitor it via grafana.

kmd-fl (approved):

PR #1499, opened 2023-03-06
feat(worker): add worker.remove [fixes NET-354 NET-376]

justprosh:


crates/particle-node-tests/tests/spells.rs
```diff
@@ -1476,3 +1476,82 @@ fn spell_create_worker_same_deal_id_different_peer() {
...
+    );
+
+    if let [JValue::Array(before), JValue::String(spell_err), JValue::String(srv_err)] = client
+        .receive_args()
+        .wrap_err("receive")
+        .unwrap()
+        .as_slice()
+    {
+        assert_eq!(before.len(), 2);
+        log::info!("RESULT {:?}", before);
```
justprosh, line 1544:
```suggestion
```

sorcerer/src/sorcerer.rs
```diff
@@ -57,11 +58,18 @@ pub struct SpellBuiltin {
 }
 
 impl SpellBuiltin {
-    pub fn new(service_id: &str) -> Self {
+    pub fn new(
+        service_id: &str,
+        funcs: Vec<(&str, ServiceFunction)>,
+        unhandled: Option<ServiceFunction>,
```
kmd-fl, line 64: What again does `unhandled` mean in this context? Quite confusing naming.

sorcerer/src/worker_builins.rs
```diff
@@ -0,0 +1,101 @@
...
+    let mut args = args.function_args.into_iter();
+    let deal_id: Option<String> = Args::next_opt("deal_id", &mut args)?;
+    Ok(JValue::String(
+        key_manager
+            .create_worker(deal_id, params.init_peer_id)?
+            .to_base58(),
+    ))
+}
+
+pub(crate) fn get_worker_peer_id(
```
kmd-fl, line 42: May we add something like `get_or_create`? It's quite uncomfortable to call it via aqua like that and get loads of `xor` errors each time.
```
func get_root_peer_id() -> string:
    worker_id: *string
    try:
        worker_id <<- WorkerRoot.get_peer_id()
    otherwise:
        worker_id <<- WorkerRoot.create()
    <- worker_id!
```

sorcerer/src/spells.rs
```diff
@@ -136,7 +158,7 @@ pub(crate) fn spell_list(spell_storage: SpellStorage) -> Result<JValue, JError>
         spell_storage
             .get_registered_spells()
             .into_iter()
-            .map(JValue::String)
+            .map(|(id, _)| JValue::String(id))
```
kmd-fl, line 161: Could it be more prudent to return a worker id here as well? It was just a list of strings before because there wasn't much data to provide.

crates/particle-node-tests/tests/spells.rs
```diff
@@ -1476,3 +1476,81 @@ fn spell_create_worker_same_deal_id_different_peer() {
     let error_msg = response[0].as_str().unwrap().to_string();
     assert!(error_msg.contains("Worker for deal_id already exists"));
 }
+
+#[test]
+fn create_remove_worker() {
```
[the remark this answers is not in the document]
justprosh: so it can only be removed by peer manager, decider or worker itself

sorcerer/src/spells.rs
justprosh: Will do it in another PR

sorcerer/src/sorcerer.rs
justprosh: All calls of this service_id that are not presented in funcs will be handled with `unhandled` function. We do it i.e. for `getDataSrv` for spell script, functions `-relay-` and `spell_id` are handled by corresponding functions and others (that are simply just arguments in spell function) are handled by `get_spell_arg` and we retrieve smth from spell's KV with provided func name

kmd-fl: Can we rename it to `default` or `fallback` or something like that?

particle-services/src/app_services.rs
```diff
@@ -203,13 +224,36 @@ impl ParticleAppServices {
...
+            if let Err(err) = remove_persisted_service(&self.config.services_dir, srv.id.clone()) {
+                log::warn!(
+                    "Error while removing persisted service for {}: {:?}",
+                    srv.id,
+                    err
+                )
+            }
+
+            if let Some(aliases) = self.aliases.write().get_mut(&srv.worker_id) {
+                for alias in srv.aliases.into_iter() {
```
folex, line 250:
```suggestion
                for alias in srv.aliases {
```

sorcerer/src/spells.rs
folex: @justprosh mind creating an issue for that?

justprosh: for sure

justprosh: NET-401

spell-storage/src/storage.rs
```diff
@@ -6,19 +6,22 @@ use derivative::Derivative;
...
+type WorkerId = PeerId;
+
 #[derive(Derivative)]
 #[derivative(Debug, Clone)]
 pub struct SpellStorage {
     // The blueprint for the latest spell service.
     spell_blueprint_id: String,
     // All currently existing spells
-    registered_spells: Arc<RwLock<HashSet<String>>>,
+    registered_spells: Arc<RwLock<HashSet<(String, WorkerId)>>>,
```
folex, line 24: Maybe make it a HashMap? Or is it possible that the same spell exists on several workers?

sorcerer/src/worker_builins.rs
justprosh: i.e. we can have `worker.exists(deal_id: ?string) -> bool` method to check worker existence and write if/else instead of try-catch

folex: Well, that doesn't change much. It will be still two FFI calls, which is rather expensive.

I think there should be a semaphore-like function, ie current `create`, and a "please just work" function like `create_or_get` for cases where program does not need exclusive access.

justprosh: yeah, we can add it whenever we need this

kmd-fl (approved):

spell-storage/Cargo.toml
```diff
@@ -11,6 +11,7 @@ service-modules = { workspace = true }
 
 fluence-app-service = { workspace = true }
 fluence-spell-distro = { workspace = true }
+libp2p = { workspace = true }
```
folex, line 14: wow, why?

folex: ah, for PeerId. OK. Though usually it is `use fluence_libp2p::PeerId;`, but doesn't matter I guess

justprosh: I think it is because of peer id usage

spell-storage/src/storage.rs
```diff
@@ -1,24 +1,27 @@
...
+type WorkerId = PeerId;
+
 #[derive(Derivative)]
 #[derivative(Debug, Clone)]
 pub struct SpellStorage {
     // The blueprint for the latest spell service.
     spell_blueprint_id: String,
     // All currently existing spells
-    registered_spells: Arc<RwLock<HashSet<String>>>,
+    registered_spells: Arc<RwLock<HashMap<String, WorkerId>>>,
```
folex, line 24: And not it looks like `registered_workers` :D

crates/key-manager/src/key_manager.rs
```diff
@@ -142,14 +157,38 @@ impl KeyManager {
...
+            .ok_or_else(|| KeyManagerError::WorkerNotFound(worker_id))
+            .map(|info| info.deal_id.clone())
+    }
+
+    pub fn remove_worker(&self, worker_id: PeerId) -> Result<(), KeyManagerError> {
+        let deal_id = self.get_deal_id(worker_id)?;
+        remove_keypair(&self.keypairs_dir, &deal_id)?;
+        self.worker_ids.write().remove(&deal_id);
+        self.worker_infos.write().remove(&worker_id);
+        self.worker_keypairs.write().remove(&worker_id);
```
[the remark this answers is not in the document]
justprosh: Ok!

spell-storage/src/storage.rs
```diff
@@ -93,7 +96,7 @@ impl SpellStorage {
     fn restore_spells(
         services: &ParticleAppServices,
         modules: &ModuleRepository,
-    ) -> HashSet<String> {
+    ) -> HashMap<String, PeerId> {
```
folex, line 99:
```suggestion
    ) -> HashMap<String, WorkerId> {
```

folex (approved):

PR #1536, opened 2023-03-28
fix: bug with repeated alias for service [NET-434]

justprosh:
\+ a little bit of refactoring

### DIFF
--- a/particle-services/src/app_service.rs
+++ b/particle-services/src/app_service.rs
@@ -16,10 +16,10 @@
 
 use crate::error::ServiceError;
 use crate::{Result, VIRTUAL_PARTICLE_VAULT_PREFIX};
+
 use fluence_app_service::{
     AppService, AppServiceConfig, MarineConfig, MarineWASIConfig, ModuleDescriptor,
 };
-
 use particle_modules::ModuleRepository;
 use peer_metrics::ServicesMetrics;
 use server_config::ServicesConfig;
@@ -32,7 +32,6 @@ pub fn create_app_service(
     modules: &ModuleRepository,
     blueprint_id: String,
     service_id: String,
-
     metrics: Option<&ServicesMetrics>,
 ) -> Result<AppService> {
     try {
@@ -57,8 +56,8 @@ pub fn create_app_service(
 
         log::debug!("Creating service {}, envs: {:?}", service_id, config.envs);
 
-        let service = AppService::new(modules, service_id.clone(), config.envs)
-            .map_err(ServiceError::Engine)?;
+        let service =
+            AppService::new(modules, service_id, config.envs).map_err(ServiceError::Engine)?;
 
         service
     }
--- a/particle-services/src/app_service.rs
+++ b/particle-services/src/app_service.rs
@@ -34,33 +34,27 @@ pub fn create_app_service(
     service_id: String,
     metrics: Option<&ServicesMetrics>,
 ) -> Result<AppService> {
-    try {
-        let mut modules_config = modules.resolve_blueprint(&blueprint_id)?;
-        modules_config
-            .iter_mut()
-            .for_each(|module| inject_vault(&config.particles_vault_dir, module));
+    let mut modules_config = modules.resolve_blueprint(&blueprint_id)?;
+    modules_config
+        .iter_mut()
+        .for_each(|module| inject_vault(&config.particles_vault_dir, module));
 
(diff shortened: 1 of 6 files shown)


particle-services/src/app_services.rs
```diff
@@ -73,7 +73,9 @@ pub struct Service {
 
 impl Service {
     pub fn remove_alias(&mut self, alias: &str) {
-        self.aliases.retain(|a| a.ne(alias));
+        if let Some(pos) = self.aliases.iter().position(|x| *x == alias) {
+            self.aliases.remove(pos);
```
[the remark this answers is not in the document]
justprosh: Yeah, I think about that, but we must decide which alias we can use in metrics. Now it's just the first element of vec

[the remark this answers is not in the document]
justprosh: Maybe we can collect metrics per-worker?

particle-services/src/app_services.rs
```diff
@@ -553,6 +551,15 @@ impl ParticleAppServices {
                 .cloned()?
         };
 
+        if Some(service_id.clone()) == previous_owner_id {
```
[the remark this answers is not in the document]
justprosh: tried to rewrite it, btw only `is_some_and` exists

kmd-fl (approved):

particle-services/src/app_services.rs
```diff
@@ -73,7 +73,9 @@ pub struct Service {
 
 impl Service {
     pub fn remove_alias(&mut self, alias: &str) {
-        self.aliases.retain(|a| a.ne(alias));
```
folex, line 76: Why retain wasn't enough? 🤔

particle-services/src/app_services.rs
```diff
@@ -553,6 +551,19 @@ impl ParticleAppServices {
                 .cloned()?
         };
 
+        let is_the_same_service = previous_owner_id
+            .clone()
```
folex, line 555: this clone also makes me dizzy

particle-services/src/app_services.rs
```diff
@@ -553,6 +551,19 @@ impl ParticleAppServices {
...
+            .is_some_and(|id| id.eq(&service_id));
+
+        if is_the_same_service {
+            // service already has this alias, nothing to do
+            return Ok(());
+        }
```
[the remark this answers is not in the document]
folex: maybe it's better to rewrite the whole `add_alias`, because it all looks very broken

particle-services/src/app_services.rs
justprosh: because this option is used after that

particle-services/src/app_services.rs
justprosh: It is more than enough, I did add_alias and remove_alias symmetrically. Now add_alias adds 1 alias, and remove_alias removes exactly one alias

particle-services/src/app_services.rs
folex: you don't need to clone to compare references

particle-services/src/app_service.rs
```diff
@@ -34,43 +32,29 @@ pub fn create_app_service(
...
-        log::debug!("Creating service {}, envs: {:?}", service_id, config.envs);
-
-        let service = AppService::new(modules, service_id.clone(), config.envs)
-            .map_err(ServiceError::Engine)?;
+    if let Some(metrics) = metrics.as_ref() {
+        metrics.observe_service_config(config.max_heap_size.as_u64(), &modules_config);
+    }
 
-        // Save created service to disk, so it is recreated on restart
-        let persisted =
```
kmd-fl, line 68: We don't `persist` a service on creation anymore?

justprosh: I moved it outside of this function because this function is a creation of AppService

kmd-fl: oh yeah, didn't see it first

kmd-fl (approved):

particle-services/src/app_services.rs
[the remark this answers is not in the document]
folex: is there any reason to allow several aliases per service?

folex (approved):

particle-services/src/app_services.rs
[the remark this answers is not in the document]
justprosh: I really don't know

PR #1539, opened 2023-03-29
fix: collect separate metrics for root alised services and worker spells [fixes NET-431]

kmd-fl:
Collect separate metrics for:
- aliased services from root scope
- worker spells (aka marked as such in aliases)

### DIFF
--- a/particle-services/src/app_services.rs
+++ b/particle-services/src/app_services.rs
@@ -383,7 +383,17 @@ impl ParticleAppServices {
         //     ));
         // }
 
-        let service_type = ServiceType::Service(service.aliases.first().cloned());
+        // Metrics collection are enables for services with aliases which are installed on root worker or worker spells.
+        let allowed_alias = if worker_id == self.config.local_peer_id {
+            service.aliases.first().cloned()
+        } else {
+            if service.aliases.iter().any(|alias| alias == "worker-spell") {
+                Some("worker-spell".to_string())
+            } else {
+                None
+            }
+        };
+        let service_type = ServiceType::Service(allowed_alias);
 
         // TODO: move particle vault creation to aquamarine::particle_functions
         self.create_vault(&particle.id)?;


particle-services/src/app_services.rs
```diff
@@ -383,7 +383,17 @@ impl ParticleAppServices {
         //     ));
         // }
 
-        let service_type = ServiceType::Service(service.aliases.first().cloned());
+        // Metrics collection are enables for services with aliases which are installed on root worker or worker spells.
+        let allowed_alias = if worker_id == self.config.local_peer_id {
+            service.aliases.first().cloned()
+        } else {
+            if service.aliases.iter().any(|alias| alias == "worker-spell") {
```
justprosh, line 390: `service.aliases.contains(&alias)` will be enough

kmd-fl: It will not https://doc.rust-lang.org/std/vec/struct.Vec.html#method.contains

justprosh: oooh, yeah, because of `&str`, damn

gurinderu (approved):

particle-services/src/app_services.rs
```diff
@@ -425,7 +425,15 @@ impl ParticleAppServices {
         //     ));
         // }
 
-        let service_type = ServiceType::Service(service.aliases.first().cloned());
+        // Metrics collection are enables for services with aliases which are installed on root worker or worker spells.
+        let allowed_alias = if worker_id == self.config.local_peer_id {
+            service.aliases.first().cloned()
+        } else if service.aliases.iter().any(|alias| alias == "worker-spell") {
```
folex, line 431: iterating through aliases on every service call, sounds harsh...

maybe we should use alias that was used to make function call? report metrics with that alias...

kmd-fl: Yeah. On the other hand, it's possible to call a spell or a service by it's id. Also it's totally possible that soon the worker-spell will call itself using `Spell "spell"` and not `Spell "worker-spell"`.

Mb we should carefully consider wether need multiple aliases for a service.

kmd-fl: I decided that checking just the first alias in the list is enough. Pushed the fix.

folex (approved):

PR #1540, opened 2023-03-29
feat: extend peer identify with spell service version and allowed mounted binaries list [fixes NET-429 NET-381]

kmd-fl:
Extend `Peer.identify`
* add list of `allowed_binaries`
* add `spell_version`

Looks like this
```json
{
  "air_version": "0.38.0",
  "allowed_binaries": [
    "/usr/bin/ipfs",
    "/usr/bin/curl"
  ],
  "external_addresses": [],
  "node_version": "0.9.1",
  "spell_version": "0.5.6"
}
```

Spell version is determent as follows:
1. If it's loaded from crate, the crate version is in use.
2. If it's loaded from disk, version is in the format `wasm hashes <hash1[..8]> <hash2[..8]>` with first 8 symbols of each hash.

To do this, I needed to move `Peer.identify` implementation from builtins to the sorcerer. Can't say I like it very much, but it works.

### DIFF
--- a/particle-node/src/node.rs
+++ b/particle-node/src/node.rs
@@ -327,6 +327,12 @@ impl<RT: AquaRuntime> Node<RT> {
             external_addresses,
             node_version: env!("CARGO_PKG_VERSION"),
             air_version: air_interpreter_wasm::VERSION,
+            spell_version: fluence_spell_distro::VERSION,
+            allowed_binaries: services_config
+                .allowed_binaries
+                .iter()
+                .map(|s| s.to_string_lossy().to_string())
+                .collect::<_>(),
         };
 
         Builtins::new(
--- a/particle-node/src/node.rs
+++ b/particle-node/src/node.rs
@@ -249,7 +249,7 @@ impl<RT: AquaRuntime> Node<RT> {
         let (spell_event_bus, spell_event_bus_api, spell_events_receiver) =
             SpellEventBus::new(sources);
 
-        let (sorcerer, mut spell_service_functions, spell_version) = Sorcerer::new(
+        let (sorcerer, mut custom_service_functions, spell_version) = Sorcerer::new(
             builtins.services.clone(),
             builtins.modules.clone(),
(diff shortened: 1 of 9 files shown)


sorcerer/src/sorcerer.rs
```diff
@@ -348,4 +352,20 @@ impl Sorcerer {
             .boxed()
         }))
     }
+
+    fn make_peer_builtin(&self, node_info: NodeInfo) -> (String, CustomService) {
```
folex, line 356: pr is awesome, but I doubt that it is a good idea to hold that closure in Sorcerer.

Sorcerer shouldn't be responsible for Peer.identify, cuz there is no reason for that.

Can you move it to a separate entity? maybe an ad hoc closure even

folex (approved):

sorcerer/src/sorcerer.rs
kmd-fl: I thought it was the idea (to move to sorcerer, I mean) to untangle dependencies.

kmd-fl: Oh, you've said 'like in Sorcerer', not `in Sorcerer' in that thread, sry. 

It can be moved out, but we would need to create a new system/entity for these extended builtins which aren't a part of `Sorcerer` or `Builtins`. I think that would bring more pain tbh and it's better to think about refactoring these two guys first.

particle-node/src/node.rs
```diff
@@ -243,7 +249,7 @@ impl<RT: AquaRuntime> Node<RT> {
         let (spell_event_bus, spell_event_bus_api, spell_events_receiver) =
             SpellEventBus::new(sources);
 
-        let (sorcerer, spell_service_functions) = Sorcerer::new(
+        let (sorcerer, mut spell_service_functions, spell_version) = Sorcerer::new(
```
justprosh, line 252: let's call it `custom_service_functions` because it already has worker, spell and identify builtins

PR #1549, opened 2023-04-03
fix(metrics): collect metrics for custom services [fixes NET-438]

kmd-fl:


folex (approved):

crates/peer-metrics/src/services_metrics/builtin.rs
```diff
@@ -15,7 +15,7 @@ use crate::services_metrics::message::ServiceCallStats;
 type ServiceId = String;
 type Name = String;
 
-/// Store a part of series of numeric observations and some parameters that desribe the series.
+/// Store a part of series of numeric observations and some parameters that describes the series.
```
justprosh, line 18:
```suggestion
/// Store a part of series of numeric observations and some parameters that describe the series.
```
"parameters" is plural

particle-builtins/src/builtins.rs
```diff
@@ -140,12 +140,16 @@ where
     pub async fn call(&self, args: Args, particle: ParticleParams) -> FunctionOutcome {
         let start = Instant::now();
         let result = self.builtins_call(args, particle).await;
+        let result = match result {
+            FunctionOutcome::NotDefined { args, params } => {
+                self.custom_service_call(args, params).await
+            }
+            result => result,
+        };
         let end = start.elapsed().as_secs();
```
justprosh, line 149: So it measures builtin+custom service calls?

justprosh (approved):

particle-builtins/src/builtins.rs
kmd-fl: Yeah, I think the time needed for ensuring the service call isn't in the builtins list isn't significant. 

Probrably, can do smth like
```rust
     let mut start = ...;
     ....
            FunctionOutcome::NotDefined { args, params } => {
                start = Instant::now();
                self.custom_service_call(args, params).await
            }
       ...
```

PR #1552, opened 2023-04-05
feat(metrics): add rust-peer version info in prometheus metrics [fixes NET-422]

kmd-fl:
Add new `rust_peer_build_info` metric.
```
# HELP rust_peer_info_version Rust Peer version.
# TYPE rust_peer_info_version info
rust_peer_version_info{peer-version="0.9.1",air-version="0.38.0",spell-version="0.5.6"} 1
```
These versions are the same as in `Peer.identify`: `peer-version` is from `CARGO_PKG_VERSION`, `air-version` is `air_interpreter_wasm::VERSION`.

~~I think, atm it's enough to add only basic versions. Suggest to extend it when we feel that we need it.~~
UPD: ~~I'll add spell version and builtins versions.~~
UPD: added spell_version

### DIFF
--- a/crates/peer-metrics/src/info.rs
+++ b/crates/peer-metrics/src/info.rs
@@ -13,5 +13,5 @@ pub fn add_info_metrics(
         ("air-version", air_version),
         ("spell-version", spell_version),
     ]);
-    sub_registry.register("version", "Rust Peer versions", info);
+    sub_registry.register("build", "Rust Peer Info", info);
 }
--- a/crates/peer-metrics/src/info.rs
+++ b/crates/peer-metrics/src/info.rs
@@ -3,13 +3,15 @@ use prometheus_client::registry::Registry;
 
 pub fn add_info_metrics(
     registry: &mut Registry,
-    node_version: &'static str,
-    air_version: &'static str,
+    node_version: String,
+    air_version: String,
+    spell_version: String,
 ) {
     let sub_registry = registry.sub_registry_with_prefix("rust_peer_info");
     let info = Info::new(vec![
         ("peer-version", node_version),
         ("air-version", air_version),
+        ("spell-version", spell_version),
     ]);
     sub_registry.register("version", "Rust Peer versions", info);
 }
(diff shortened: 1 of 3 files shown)


crates/peer-metrics/src/info.rs
```diff
@@ -0,0 +1,15 @@
+use prometheus_client::metrics::info::Info;
+use prometheus_client::registry::Registry;
+
+pub fn add_info_metrics(
+    registry: &mut Registry,
+    node_version: &'static str,
+    air_version: &'static str,
+) {
+    let sub_registry = registry.sub_registry_with_prefix("rust_peer_info");
+    let info = Info::new(vec![
```
folex, line 10: Would be also great to add `spell version`, and version for every builtin.

crates/peer-metrics/src/info.rs
```diff
@@ -0,0 +1,15 @@
...
+        ("peer-version", node_version),
+        ("air-version", air_version),
```
folex, line 12:
```suggestion
        ("avm-version", avm_version),
```

crates/peer-metrics/src/info.rs
kmd-fl: Ok, I'll try to find builtins versions

crates/peer-metrics/src/info.rs
kmd-fl: That's how it's called in `Peer.identify`. Should we rename there as well?

folex: argh. Well, let's keep it `air` everywhere then. Too minor a reason to worth a breaking change.

crates/peer-metrics/src/info.rs
kmd-fl: I guess, I'll add `spell_version`, and we'll think about builtins later?

folex: yep

kmd-fl: done it btw

folex (approved):

nahsi:
Let's name it `rust_peer_info`? :)

justprosh (approved):

PR #1565, opened 2023-04-11
feat: add service metrics for spells [fixes NET-441]

kmd-fl:
Add a new service type for spells and spells with aliases.

Extend some new metrics with labels:
* services_count
* creation_time_msec
* removal_time_msec
* creation_count
* removal_count

And add a __hack__ to detect that a service is a spell _fast_, without rechecking the blueprint cache on every call.

Need to design a non-hacky way to distinguish a service and a spell, I think. It will be needed not only for this metrics collection thing but also for rewriting the registry and other built-in services as spells.

### DIFF
--- a/particle-services/src/app_services.rs
+++ b/particle-services/src/app_services.rs
@@ -872,12 +872,7 @@ impl ParticleAppServices {
         } else {
             None
         };
-        // How to detect that the service is a spell?
-        // Afaik there's no way to detect that the service is a spell FAST
-        // We can check blueprint name, but it's not reliable
-        // Need to rework that
 
-        // This approach doesn't work since not only spells calls spell functions
         if service.is_spell {
             ServiceType::Spell(allowed_alias)
         } else {
--- a/particle-services/src/app_services.rs
+++ b/particle-services/src/app_services.rs
@@ -60,10 +60,7 @@ pub enum ServiceType {
 
 impl ServiceType {
     fn is_spell(&self) -> bool {
-        match self {
-            ServiceType::Spell => true,
-            _ => false,
-        }
+        matches!(self, ServiceType::Spell)
     }
 }
 
@@ -442,7 +439,7 @@ impl ParticleAppServices {
         //     ));
         // }
         // Metrics collection are enables for services with aliases which are installed on root worker or worker spells.
-        let service_type = self.get_service_type(&service, &worker_id);
+        let service_type = self.get_service_type(service, &worker_id);
 
         // TODO: move particle vault creation to aquamarine::particle_functions
         self.create_vault(&particle.id)?;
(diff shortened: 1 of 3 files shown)


particle-services/src/app_services.rs
```diff
@@ -324,17 +328,13 @@ impl ParticleAppServices {
             )?;
 
             // tmp hack to forbid spell removal via srv.remove
```
folex, line 330: is it temporary? What's the plan to fix it? @justprosh

particle-services/src/app_services.rs
```diff
@@ -841,26 +829,26 @@ impl ParticleAppServices {
...
         let stats = service.module_memory_stats();
         let stats = ServiceMemoryStat::new(&stats);
-        let allowed_alias = if worker_id == self.config.local_peer_id {
-            aliases.first().cloned()
-        } else {
-            None
-        };
-        let service_type = ServiceType::Service(allowed_alias);
+
+        // Would be nice to determine that we create a spell service, but we don't have a reliable way to detect it yet
```
folex, line 833: Pass it as a boolean flag to `create_service`. I think that would be cleaner.

particle-services/src/app_services.rs
```diff
@@ -870,6 +858,32 @@ impl ParticleAppServices {
...
+            Some("worker-spell".to_string())
+        } else {
+            None
+        };
+        // How to detect that the service is a spell?
+        // Afaik there's no way to detect that the service is a spell FAST
+        // We can check blueprint name, but it's not reliable
+        // Need to rework that
+
+        // This approach doesn't work since not only spells calls spell functions
```
folex: What doesn't work? I don't get it

kmd-fl: oh yeah, this is an outdated comment, I removed it.

particle-services/src/app_services.rs
```diff
@@ -69,6 +69,8 @@ pub struct Service {
     pub service: Mutex<AppService>,
     pub service_id: String,
     pub blueprint_id: String,
+    // temp hack to detect if the service is a spell
```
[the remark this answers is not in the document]
kmd-fl: I don't know if it's a temporary solution, really. I hoped to have `enum ServiceType` in the future, so I guess this flag is temp.

But since this pr isn't about spells and services but about metrics, I decided to use a boolean flag as a compromise.

particle-services/src/app_services.rs
justprosh: I think it is fixed with `service.is_spell` by @kmd-fl

particle-services/src/app_services.rs
kmd-fl: The flag's value always depends on the other arguments, it would be invalid to pass anything else, so it would be strange if the API allowed this.

PR #1566, opened 2023-04-12
fix: unnecessary vault creation [fixes NET-440]

gurinderu:


sorcerer/src/spells.rs
```diff
@@ -298,6 +302,7 @@ pub(crate) fn get_spell_arg(
             Some(params.id.clone()),
             params.init_peer_id,
             Duration::from_millis(params.ttl as u64),
+            true, //TODO: check it
```
justprosh, line 305: why? I think this can be false also

particle-services/src/app_services.rs
```diff
@@ -553,7 +561,7 @@ impl ParticleAppServices {
             signature: vec![],
         };
 
-        self.call_service(args, particle)
```
justprosh, line 556: I think `call_function` should always have `create_vault` set to `false` because there is no way to clear the vault for the particle created here

gurinderu: Agree, fixed

justprosh (approved):

PR #1578, opened 2023-04-17
feat(spells): resolve spell_id by 'spell' and 'self' [NET-419]

justprosh:


sorcerer/src/utils.rs
```diff
@@ -61,16 +61,9 @@ where
...
-            .split('_')
-            .collect::<Vec<&str>>()
-            .get(1)
-            .ok_or(JError::new("Invalid particle id format"))?)
-    } else {
-        Err(JError::new(format!(
-            "Expected spell particle id to start with 'spell_': {particle_id}"
-        )))
-    }
+pub(crate) fn parse_spell_id_from(particle: &ParticleParams) -> Result<String, JError> {
```
kmd-fl, line 64: Why do you pass `ParticleParams` if you use only `particle.id`?

particle-execution/src/particle_params.rs
```diff
@@ -54,4 +54,16 @@ impl ParticleParams {
             signature: signature.clone(),
         }
     }
+
+    pub fn is_spell_particle(particle_id: &str) -> bool {
```
kmd-fl, line 58: Is it possible to use it everywhere where `starts_with("spell_")` check happens? To have a single place with the check. I see several places in the codebase with the check.

particle-services/src/error.rs
```diff
@@ -52,6 +52,8 @@ pub enum ServiceError {
     ForbiddenAliasWorker(PeerId),
     #[error("Cannot add alias '{0}' because there is a service with that id")]
     AliasAsServiceId(String),
+    #[error("Cannot add alias 'spell' because it is reserved")]
```
kmd-fl, line 55: Come to think of it, isn't the "spell" alias also used for spell-controlling functions like `install`, `list`, `update`?  Doesn't it clash? Maybe we should use `"self"` instead of `"spell"`?

justprosh: I don't know, I like that we can control everything about spells with `spell` alias, but we should remember which func names reserved whenever we update spell service
@folex what do you think?

particle-execution/src/particle_params.rs
justprosh: Yeah, forgot to fix it

sorcerer/src/utils.rs
justprosh: idk, it requires less symbols

particle-services/src/error.rs
kmd-fl: Yeah, you right that "spell" has obvious meaning, which "self" hasn't. Still don't know if the clashing is a problem.

Mb the idea of the change in general isn't as good as we think it is. I feel that something is off but I can't tell what exactly.

kmd-fl (approved):

PR #1579, opened 2023-04-18
fix: long arguments [fixes NET-442]

gurinderu:


crates/server-config/src/args.rs
```diff
@@ -59,13 +59,12 @@ pub struct RootKeyPairArgs {
...
 
     #[arg(
-    short('g'),
-    long,
-    value_parser = clap::value_parser!(bool),
-    id = "ROOT_KEY_PAIR_GENERATE",
-    help_heading = "Node keypair",
-    display_order = 13,
-    action =  clap::ArgAction::SetTrue
+        short('g'),
```
kmd-fl, line 62: Mb add a long argument here as well?

gurinderu: Fixed

kmd-fl (approved):

PR #1591, opened 2023-04-25
fix(metrics): provide separate memory metrics for spells

kmd-fl:
Due to the tricky system for collecting and providing to Prometheus memory metrics, spells metrics were sorted into `non-alised-services` service type and incorrectly displayed in Graphana. This PR fixes it.


crates/peer-metrics/src/services_metrics/backend.rs
```diff
@@ -137,11 +137,14 @@ impl ServicesMetricsBackend {
     }
 
     /// Actually send all collected memory memory_metrics to Prometheus.
+    ///
+    /// TODO: this all system looks overcomplicated, need to look into ways to simplify it.
```
folex, line 141: It's very hard to act upon this TODO

kmd-fl: I don't think I'll do it in the nearest future

kmd-fl: I'll also create a linear task

folex (approved):

crates/peer-metrics/src/services_metrics/backend.rs
folex: No, I mean either remove it or rewrite the TODO so that it is clear what the problem is. Right now it is like "I see there's a problem". Instead, it should be "Problems I see here are: X, Y, Z"

PR #1592, opened 2023-04-27
fix(spells): start resubscribed spells scheduling after full node initialization [fixes NET-459]

kmd-fl:


crates/spell-event-bus/src/bus.rs
```diff
@@ -208,6 +210,33 @@ impl SpellEventBus {
         let mut sources_channel = futures::stream::select_all(sources);
 
         let mut state = SubscribersState::new();
+        loop {
+            let result: Result<(), BusInternalError> = try {
+                select! {
+                    Some(command) = self.recv_cmd_channel.recv() => {
```
gurinderu, line 216: Probably it should be inside one loop and one select

folex: whoah, yeah, we need to refactor that

folex: I think something like `FusedStream` might help

folex: `select!` supports conditions in branches, ie

```
<pattern> = <async expression> (, if <precondition>)? => <handler>,
```

So basically you can have a flag that is set in Run and avoid polling `timer` and `sources` if it is not set.

folex (commented):
Awesome! This is a very important PR that will make nox more stable.

Just a few things to change before the merge.

particle-node/src/node.rs
```diff
@@ -489,6 +493,11 @@ impl<RT: AquaRuntime> Node<RT> {
             .await
             .wrap_err("builtins deploy failed")?;
 
+        let result = self.spell_event_bus_api.run_scheduling().await;
```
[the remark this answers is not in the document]
kmd-fl: Fixed, but it does happen only once, there's no stop.

crates/spell-event-bus/src/bus.rs
kmd-fl: Changed on using select conditions. Checked that spells aren't run until `start_scheduling`

particle-node/src/node.rs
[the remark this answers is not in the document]
folex: I mean that `run_scheduling` sounds like something you should call every time you want scheduling to happen.

Like `loop { run_scheduling }`, while `start_scheduling` sounds like a start of a self-sustaining process

folex (approved):

PR #1593, opened 2023-04-28
feat(transport): enable compression

folex:


folex:
support for that feature was removed, and we need further research on how to use Deflate in libp2p. Right now it is not clear.

https://github.com/libp2p/rust-libp2p/pull/3949

PR #1606, opened 2023-05-05
fix(config): aqua-pool-size argument

gurinderu:
## Description

Fix aqua-pool-size argument


kmd-fl:
Do I understand correctly that the problem was in the variable's name and that the simple renaming fixes the issue?

gurinderu:
> Do I understand correctly that the problem was in the variable's name and that the simple renaming fixes the issue?

exactly, var names in deserialized structure should have the same name

kmd-fl (approved):

justprosh (approved):

folex (approved):

PR #1608, opened 2023-05-10
feat(config)!: Introduce tracing config, change allowed_binaries list format

gurinderu:
## Description

- Added tracing config
- Changed a config library because of bad error messaging
- Changed a key separator for envs from _ to __
- Changed a list format (ie FLUENCE__ALLOWED_BINARIES = "/bin/sh, /bin/bash" instead of "[\"/bin/sh\",\"/bin/bash\"]"


kmd-fl:
What does "tracing config" mean? What does it do?

gurinderu:
> What does "tracing config" mean? What does it do? @gurinderu

It is a configuration for the opentelemetry collector

crates/server-config/src/defaults.rs
```diff
@@ -182,7 +182,7 @@ pub fn default_management_peer_id() -> PeerId {
     let public_key = PublicKey::Ed25519(kp.public());
     let peer_id = PeerId::from(public_key);
 
-    println!(
+    log::info!(
```
folex, line 185: yaaay! ^__^

folex (approved):

PR #1618, opened 2023-05-16
fix(network): fixed allow private addrs

gurinderu:


particle-node/src/behaviour/network.rs
```diff
@@ -59,7 +60,10 @@ impl FluenceNetworkBehaviour {
             cfg.connection_pool_metrics,
         );
 
+        let keep_alive = KeepAlive::default();
```
folex, line 63: How will it behave? Why not set `keep_alive(true)` in Ping?

folex: it seems that keep_alive::Behvaiour is intended to use in test code, mostly. 

```
/// Implementation of [`NetworkBehaviour`] that doesn't do anything other than keep all connections alive.
///
/// This is primarily useful for test code. In can however occasionally be useful for production code too.
/// The caveat is that open connections consume system resources and should thus be shutdown when
/// they are not in use. Connections can also fail at any time so really, your application should be
/// designed to establish them when necessary, making the use of this behaviour likely redundant.
```

folex (approved):

PR #1623, opened 2023-06-05
feat(system-service-deployer): introduce new system service deployment system [fixes NET-487]

kmd-fl:
1. New crate `system-service-deployer`. Installs all system services.
2. Move decider and aqua-ipfs config to the node's config. Added default values for easier running.
3. Refactor other small things a bit. Maybe, I needed to do it in a separate commit/pr, sorry.

Atm in the pr there are some after-testing stuff (like crates with "-test" suffix) that I didn't remove yet, but I'll do it.

Some description of the PR.

----------------------

## Service Configuration

Service configuration for `aqua-ipfs` and `decider` is moved to the node's config:
```
[system_services]
enable = [ 
    "aqua-ipfs",
     "registry"
]

[system_services.decider]
decider_period_sec = 1000
worker_period_sec = 123

[system_services.aqua_ipfs]
local_api_multiaddr = "/ip4/127.0.0.1/tcp/5001"
```

**UPD**: added enabled services

We may additionally list the system services we want to be deployed. By default, we enable ALL system services.

**Note** that atm newly disabled services aren't removed yet

Default values that we use in our network are used as default values, so right now, we don't need
to change the config in the nox-distro.

However, configuration for `trust-graph` (fluence certificates) is distributed within the crate, since
we decided that this configuration is part of the `trust-graph` and not of the node.

**Old ENVs are supported now and overrides the configuration file**. 

## System Service Distribution

There's _no unified approach_ for this atm since I don't know how to provide a unified library for it.

I come up with the following approach. These structures are now created at the system service deployer
```rust
// For describing services
struct ServiceDistro {
    modules: HashMap<&'static str, &'static [u8]>,
    config: &'static [u8],
    name: &'static str, // is used as alias
}

// For describing spells
struct SpellDistro {
    name: &'static str, // is used a alias
    air: &'static str, // air script of the spell
    kv: HashMap<&'static str, JValue>, // initial KV of the spell
    trigger_config: TriggerConfig,
}
```

It's a slightly more difficult situation for spells since we need somehow to *provide initial values* from the node's config
in the format for a spell to understand.

Atm for `decider`, we provide a structure with the config to the decider distro crate; the crate takes the values
and returns us `SpellDistro` with *spell-compatible* KV init data.

I also wanted to create a `PackageDistro` which would unite spells and services of one package (like with `decider`),
but it doesn't unify properly because of the need to initialize `aqua-ipfs` and `trust-graph`.

## Running deployer

System services deployment happens _after_ running the main node loop. This is required for subscribing
system spells.

There's no specific reason for it. Maybe we need to move it to the main node loop, to the initialization phase,
where all the node's subsystems are started. This will allow us to stop the node easily if the system service deployer
fails.

**ATM if the deployer wasn't able to deploy everything on the first try, it will stop trying again, and will do nothing** 

I will change it in the following PR.

## Deployment process

Atm we need to install 3 stand-alone services and 1 spell with an aux service:
1. `aqua-ipfs`, requires initialization with local and external api multi addresses of an IPFS node, provided in the node config; note that these values aren't very much in use in our spells, maybe we can remove them
2. `registry`
3. `trust-graph`, requires initialization with the Fluence certificates distributed with the service
4. `decider` and its `connector`, require a load of initial values for the spell provided in the node config; the `connector` itself doesn't need any initialization

Deployment of every service and spell happens similarly:
1. Detect if we need to install/update the service
2. Remove old service/spell
3. Install/Update
4. Initialize if needed

In the code, the 1 and 2 steps are united in one function, `deploy_system_service` and `deploy_system_spell`.

### Find existing services and spells

The rule is simple: two services/spells are the same if they share the same alias.

For services, if the new blueprint is different from the old one, install a new service

For spells, we don't compare blueprints, only scripts, and trigger configs. If the script or the config
is different, we update the spell.

### Remove old service/spell

If we found existing service/sepll and detected that we need to update it, now first we will remove the old instance.

For services, we just plainly try to remove a service.

For spells, first, we try to unsubscribe them from the triggers. We use the same function as for `Spell.remove`. If removing fails,
we try just to unsubscribe the spell and update its trigger config to an empty trigger config to avoid resubscribing on restart.

**ATM we install a new service/spell despite how the removing processes ended** 

I think we will change it in the following PRs.

### Service/Spell Deployment

Service installation doesn't require much:
1. Add modules of the service to the module repo to get a blueprint
2. Create a service
3. Add an alias

For spells:
1. Create a new spell (using the function as in `Spell.install`)
2. Add an alias

### System Service Owner

The owner of system services (aka controller) is the node itself, **services are controlled by** `HOST_PEER_ID`.
Management PeerId is used only for assigning aliases.

This is done with way because `worker_id` and `owner_id` in the current node implementation are considered to be the same entities.

It would be nice to be able to control the system services with Management PeerId.

### Initialization

Initialization happens after installation. It's not a unified process, it requires manual
implementation. For example, `aqua-ipfs` requires to set API multiaddr by calling two functions
`set_local_api_multiaddr` and `set_external_api_multiaddr`; on the other hand, `trust-graph`
wants us to call `set_root` with an address of a root node and also call `insert_cert` for
every provided certificate from the provided array.

Previously, service initialization was implemented via air scripts which were provided by the services.
We removed this for the node's simplification.

One of the approaches to unify this initialization step is to ask for the services to provide
`setup` or `init` functions, so the system deployer could do smth like
```rust
fn initialize_service(service: ServiceDistro, service_config: Option<ServiceSpecificConfig>) {
    let init_data_json = service.init_data(service_config);
	self.call_service(service.name, service.init_function_name, service.init_data_json);
	...
}
```

Cons:
- Required rewriting of the existing services
- Probably, we need to avoid services initialization and leave this possibility only to spells

### Service Calls

During the deployment process, we use
```
const SYSTEM_SERVICE_DEPLOYER_TTL: u64 = 60_000;
```
as TTL for the service calls.

So, ATM if the service is too long to call, the whole initialization will fail. Good to know, right?

# Questions

1. Do we need to stop the node if the system deployer fails?
    - Yes, we want.
3. Do we want to allow service initialization? Do we want to leave this option to spells only (via init data or KV)?
    - Yes, we want to allow system service initialization.
5. Do we want to assign the `HOST_PEER_ID` as `worker_id` for the system services? Can't assign Management PeerId.
    - For now, we will use HOST_PEER_ID as an owner of all system services and spells.
7. Do we want to remove old services? Do we want to remove old spells?
    - Yes, we want to remove old services. We want to try to update old spells.
9. How to update a spell without erasing the old state? How to check if the update is a breaking change for the old state?
     - We will try to implement it in the next PRs.
11. Do we want to be able to re-initialize or re-install services/spells on the changes in the system service config which is a part of the node's config? Some services cannot be re-initialized, only reinstalled (like aqua-ipfs)
     - We want to try when applicable. Try to re-initialize spells KV on each run.


Cargo.toml
```diff
@@ -41,6 +41,7 @@ members = [
     "script-storage",
     "spell-storage",
     "particle-execution",
+    "crates/system-service-deployer"
```
folex, line 44: How about a shorter name?

```suggestion
    "crates/system-services"
```

folex: `system_services=debug` would look saner

crates/server-config/src/defaults.rs
```diff
@@ -204,3 +204,23 @@ pub fn default_max_builtin_metrics_storage_size() -> usize {
...
+pub fn default_deal_network_api_endpoint() -> String {
+    "https://testnet.aurora.dev".to_string()
+}
+
+pub fn default_deal_contract_address_hex() -> String {
+    "0xb497e025D3095A197E30Ca84DEc36a637E649868".to_string()
+}
+
+pub fn default_deal_contract_block_hex() -> String {
+    "0x75f3fbc".to_string()
```
folex, line 225: wdyt about putting "latest" here? 🤔

particle-builtins/src/builtins.rs
```diff
@@ -841,7 +841,7 @@ where
     }
 
     fn list_services(&self, params: ParticleParams) -> JValue {
-        JValue::Array(self.services.list_services(params.host_id))
+        json!(self.services.list_services(params.host_id))
```
folex, line 844: Why? It was statically type-checked before, now it will accept any structure from `list_services`, so it is easier to make a mistake.

particle-node/src/lib.rs
```diff
@@ -20,7 +20,6 @@
 #![recursion_limit = "512"]
 #![warn(rust_2018_idioms)]
 #![deny(
-    dead_code,
```
folex, line 23: why?

particle-node/src/node.rs
```diff
@@ -51,6 +49,7 @@ use server_config::{NetworkConfig, ResolvedConfig, ServicesConfig};
 use sorcerer::Sorcerer;
 use spell_event_bus::api::{PeerEvent, SpellEventBusApi, TriggerEvent};
 use spell_event_bus::bus::SpellEventBus;
+use system_service_deployer::SystemServiceDeployer;
```
folex, line 52:
```suggestion
use system_services::Deployer;
```

particle-node/src/node.rs
```diff
@@ -494,11 +498,11 @@ impl<RT: AquaRuntime> Node<RT> {
...
 
-        let mut builtins_deployer = self.builtins_deployer;
-        builtins_deployer
-            .deploy_builtin_services()
-            .await
-            .wrap_err("builtins deploy failed")?;
+        // Note: need to be after the start of the node to be able to subscribe spells
+        let deployer = self.system_service_deployer;
+        if let Err(e) = deployer.deploy_system_services().await {
+            log::error!("installing system services failed: {}", e);
```
folex, line 504:
```suggestion
            log::error!("deploying system services failed: {}", e);
```

sorcerer/src/lib.rs
```diff
@@ -1,13 +1,14 @@
 #![feature(try_blocks)]
 #![feature(extend_one)]
 pub use sorcerer::Sorcerer;
+pub use spell_builtins::{get_spell_info, spell_install_inner, SpellInfo};
```
folex, line 4: `pub use **_inner` is strange. Why?

sorcerer/src/spell_builtins.rs
```diff
@@ -96,8 +127,8 @@ pub(crate) async fn spell_install(
             "set_script_source_to_file",
             vec![json!(script)],
             None,
-            worker_id,
-            Duration::from_millis(params.ttl as u64),
+            init_peer_id,
```
folex, line 130: why change worker_id to init_peer_id?

sorcerer/src/spell_builtins.rs
```diff
@@ -111,8 +142,8 @@ pub(crate) async fn spell_install(
             "set_json_fields",
             vec![json!(init_data.to_string())],
             None,
-            worker_id,
-            Duration::from_millis(params.ttl as u64),
+            init_peer_id,
```
folex, line 145: ditto

sorcerer/src/spell_builtins.rs
```diff
@@ -126,8 +157,8 @@ pub(crate) async fn spell_install(
             "set_trigger_config",
             vec![json!(user_config)],
             None,
-            worker_id,
-            Duration::from_millis(params.ttl as u64),
+            init_peer_id,
```
folex, line 160: ditto

sorcerer/src/spell_builtins.rs
```diff
@@ -142,27 +173,96 @@ pub(crate) async fn spell_install(
...
+
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub struct SpellInfo {
+    pub script: String,
+    pub trigger_config: TriggerConfig,
+}
+
+pub fn get_spell_info(
+    services: &ParticleAppServices,
+    host_id: PeerId,
```
folex, line 201: shouldn't it accept worker_id?

crates/system-service-deployer/src/lib.rs
```diff
@@ -0,0 +1,527 @@
...
+use sorcerer::{get_spell_info, spell_install_inner};
+use spell_event_bus::api::SpellEventBusApi;
+use spell_storage::SpellStorage;
+use std::collections::HashMap;
+use std::time::Duration;
+
+const SYSTEM_SERVICE_DEPLOYER_TTL: u64 = 60_000;
+
+// TODO: do smth with naming
+enum DeployedService {
```
folex, line 22:
```suggestion
enum SystemService {
```

crates/system-service-deployer/src/lib.rs
```diff
@@ -0,0 +1,527 @@
...
+                    log::warn!("couldn't stop the old spell {}: {err}", spell_distro.name);
+                }
+            }
+            FindSpell::NoUpdate(spell_id) => {
+                return Ok(DeployedService::Existing(spell_id));
+            }
+            FindSpell::NotFound => {}
+        }
+
+        let spell_id = spell_install_inner(
```
folex, line 367: Would be nice to just update spell script and trigger config instead of redeploying, if blueprint has not changed.

crates/system-service-deployer/src/lib.rs
```diff
@@ -0,0 +1,527 @@
...
+            &self.services,
+            &self.spell_storage,
+            &self.spell_event_bus_api,
+            self.root_worker_id,
+            self.root_worker_id,
+            self.root_worker_id,
+            "dummy-particle-id".to_string(),
```
folex, line 374:
```suggestion
            "system-services-deployment".to_string(),
```

crates/system-service-deployer/src/lib.rs
```diff
@@ -0,0 +1,527 @@
...
+                log::error!(
+                    "can't obtain details on a spell {} (will create a new one): {err}",
+                    new_spell.name
+                );
+                return FindSpell::NotFound;
+            }
+        };
+        if spell.service_type != ServiceType::Spell {
+            log::warn!(
+                "alias `{}` already used for a service, but now intended for a spell",
```
folex, line 417: So it will be overwritten?

crates/system-service-deployer/src/lib.rs
```diff
@@ -0,0 +1,527 @@
...
+    // Two spells are the same if
+    // - they have the same alias
+    //
+    // Need to redeploy (stop the old one, create a new one) a spell if
+    // - the script is different
+    // - the trigger config is different
+    fn find_same_spell(&self, new_spell: &SpellDistro) -> FindSpell {
+        let existing_spell =
+            self.services
+                .get_service_info("", self.root_worker_id, new_spell.name.to_string());
```
folex, line 400: why always `root_worker_id`? 🤔 

Aliases are by worker_id which corresponds to deal_id, right? If you look for aliases in the root worker, then you won't find them 🤔

crates/system-service-deployer/src/lib.rs
```diff
@@ -0,0 +1,527 @@
...
+
+        Ok(DeployedService::Created(spell_id))
+    }
+
+    // Two spells are the same if
+    // - they have the same alias
+    //
+    // Need to redeploy (stop the old one, create a new one) a spell if
+    // - the script is different
+    // - the trigger config is different
```
folex, line 396:
```suggestion
    // - the trigger config is different
    // - the blueprint is different
```

crates/system-service-deployer/src/lib.rs
folex: if yes, let's state that in the log

crates/system-service-deployer/src/lib.rs
```diff
@@ -0,0 +1,527 @@
...
+            }
+        }
+        None
+    }
+
+    fn add_modules(&self, service_distro: ServiceDistro) -> eyre::Result<String> {
+        let marine_config: TomlMarineConfig = toml::from_slice(service_distro.config)?;
+        let mut hashes = Vec::new();
+        for config in marine_config.module {
+            let name = config.name.clone();
```
folex, line 504: Why clone here? seems like you only use it as `.as_str`, so a ref should do 🤔

crates/system-service-deployer/src/lib.rs
```diff
@@ -0,0 +1,527 @@
...
+            // TODO: introduce nice errors for this
+            let module = service_distro
+                .modules
+                .get(name.as_str())
+                .ok_or(eyre!(format!(
+                    "there's no module {name} in the given modules map for {}",
```
folex, line 510:
```suggestion
                    "there's no module {name} in the given modules map for system service {}",
```

folex (commented):
Well done!

crates/server-config/src/defaults.rs
```diff
@@ -204,3 +204,23 @@ pub fn default_max_builtin_metrics_storage_size() -> usize {
 pub fn default_allowed_binaries() -> Vec<String> {
     vec!["/usr/bin/curl".to_string(), "/usr/bin/ipfs".to_string()]
 }
+
+pub fn default_ipfs_multiaddr() -> String {
+    "/dns4/ipfs.fluence.dev/tcp/5001".to_string()
+}
+
+pub fn default_spell_period_sec() -> u32 {
+    121
```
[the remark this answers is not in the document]
justprosh: If we set it to 30 min, it will take 30 min in the worst case to deploy services after deal update. Maybe we should trigger worker-spell from decider spell in some cases?

crates/server-config/src/defaults.rs
kmd-fl: afaik, it doesn't work properly on some chains. If set  `from_block` to `latest` and `to_block` to `latest`, the chain will always return an empty list.

Cargo.toml
kmd-fl: Or `ss-deployer`! :upside_down_face:

crates/server-config/src/system_services_config.rs
```diff
@@ -0,0 +1,73 @@
...
+
+#[derive(Clone, Serialize, Deserialize, Debug)]
+pub struct DeciderConfig {
+    #[serde(default = "default_spell_period_sec")]
+    pub decider_period_sec: u32,
+    #[serde(default = "default_spell_period_sec")]
+    pub worker_period_sec: u32,
+    #[serde(default = "default_ipfs_multiaddr")]
+    pub worker_ipfs_multiaddr: String,
+    // FIXME: I think these fields should not have default values, but it's only for now
```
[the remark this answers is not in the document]
kmd-fl: It just feels strange to embed them like this, but I can't explain why, so I'm okay with it

particle-builtins/src/builtins.rs
kmd-fl: I believe that ParticleAppService shouldn't know anything about JSON stuff, so now it returns `Vec<ServiceInfo>` and we need to convert it to JSON.

Previously, here we could only check that `list_services` returned a list of _something_. That's all. We couldn't type-check the content, so I think it's not as valuable as it could seem to be. We can add a check that it's a list additionally if it feels right for you. 

But yeah, it would be nice to be able to check that the public interface corresponds to some spec. I wonder, how we could do it.

particle-node/src/lib.rs
kmd-fl: Oh, I did it for my testing. Will return.

sorcerer/src/lib.rs
kmd-fl: I spent too much time trying to name it properly, so in the end, I named it somehow. Need to come up with a sane name.

My intention was to use the same function as in `Spell.install`. I didn't find a _nice_ way to do it.

sorcerer/src/spell_builtins.rs
kmd-fl: Because `call_function` accepts `init_peer_id` there.

sorcerer/src/spell_builtins.rs
kmd-fl: Yeah, need to rename.

sorcerer/src/spell_builtins.rs
```diff
@@ -142,27 +173,96 @@ pub(crate) async fn spell_install(
...
+        return Err(JError::new(script_value.error));
+    };
+    Ok(SpellInfo {
+        script,
+        trigger_config,
+    })
+}
+
+// It's not in use, do we need it? I just wanted `get_spell_info`, and added this one as well
+pub fn spell_info(
```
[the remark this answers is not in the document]
kmd-fl: As I wrote in the comment, this "new" builtin is implemented because I added a function `get_spell_info`. I really only need `get_spell_info`. 

So the question is "Do we want to leave this "spell_info" as a new builtin?"

folex:
> `[system_services_config.decider]`

Maybe remove `_config` part? Just `[system_services.decider]` would look nicer as a config parameter

sorcerer/src/spell_builtins.rs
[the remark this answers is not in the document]
folex: So there would be two very similar builtins?

sorcerer/src/spell_builtins.rs
folex: But aren't `init_peer_id` and `worker_id` always the same?

sorcerer/src/lib.rs
folex: How about `spell_install`?

crates/server-config/src/defaults.rs
folex: But we do not care, right? Because decider retrieves a block from the chain, and starts to follow events from there

sorcerer/src/lib.rs
kmd-fl: `spell_install` already exists and is a builtin for `Spell.install`.

We may move some functions to other modules and use qualified names to avoid name clashing though.

sorcerer/src/spell_builtins.rs
kmd-fl: I don't know. All those `host_id`, `worker_id`, and `init_peer_id` usage is quite messy.

But yesterday, on the daily, we determined that the spell's owner is a `worker_id`, so it seems that yeah, `init_peer_id` should always be set to `worker_id` for spells.

sorcerer/src/spell_builtins.rs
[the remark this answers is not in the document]
kmd-fl: What do you mean? Atm there are four spell builtins: install, remove, update, and list. I added the new one, `info`, but I don't know if we need it and if we want it to be a part of API.

This new builtin is a just side effect. I added it while doing other stuff. So, by implementing it I offered you a discussion if we need it. That's all.

This `info` builtin returns the spell's script and the spell's trigger config.

sorcerer/src/lib.rs
kmd-fl: Called it `install_spell`.

crates/server-config/src/defaults.rs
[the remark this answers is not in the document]
kmd-fl: As a compromise, set 15 min. Feels like it should be discussed on a product level.

crates/created-swarm/src/swarm.rs
```diff
@@ -262,7 +263,13 @@ impl SwarmConfig {
             builtins_dir: None,
             spell_base_dir: None,
             timer_resolution: default_script_storage_timer_resolution(),
-            allowed_binaries: vec![],
+            allowed_binaries: vec!["/usr/bin/ipfs".to_string(), "/usr/bin/curl".to_string()],
+            disabled_system_services: vec![
```
justprosh, line 267: Maybe it's better to have `enabled_system_services.` `disabled` seems counter-intuitive

crates/system-services/src/lib.rs
```diff
@@ -0,0 +1,612 @@
...
+            spell_event_bus_api,
+            root_worker_id,
+            management_id,
+            config,
+        }
+    }
+
+    pub async fn deploy_system_services(&self) -> Result<(), JError> {
+        /*
+        let deployers: HashMap<_, Box<dyn FnOnce(&Deployer) -> Result<(), JError>>> = maplit::hashmap! {
```
justprosh, line 91: Why is it commented?

crates/system-services/src/lib.rs
```diff
@@ -0,0 +1,612 @@
...
+    fn deploy_trust_graph(&self) -> Result<(), JError> {
+        let service_distro = Self::get_trust_graph_distro();
+        let service_id = match self.deploy_system_service(service_distro)? {
+            ServiceStatus::Existing(id) => {
+                log::info!("found `trust-graph` [{id}] service, no need to redeploy",);
+                return Ok(());
+            }
+            ServiceStatus::Created(id) => id,
+        };
+        log::info!("deployed `trust-graph` [{service_id}] service");
```
justprosh, line 221: This log should be moved in `match`, because sometimes you will see logs like this:
```
found `trust-graph` [{id}] service, no need to redeploy
deployed `trust-graph` [{service_id}] service
```

crates/system-services/src/lib.rs
```diff
@@ -0,0 +1,612 @@
...
+        };
+
+        for (name, deployer) in deployers {
+            if !self.config.disable.contains(&name.to_string()) {
+                deployer()?;
+            }
+        }
+        */
+
+        if !self.config.disable.contains(&"aqua-ipfs".to_string()) {
```
justprosh, line 107: Let's create constants for services aliases

crates/system-services/src/lib.rs
```diff
@@ -0,0 +1,612 @@
...
+        }
+    }
+
+    async fn deploy_new_system_spell(&self, spell_distro: SpellDistro) -> Result<String, JError> {
+        let spell_id = install_spell(
+            &self.services,
+            &self.spell_storage,
+            &self.spell_event_bus_api,
+            self.root_worker_id,
+            "system-services-deployment".to_string(),
```
justprosh, line 426: It should be a constant too

crates/nox-tests/tests/builtin.rs
```diff
@@ -1466,15 +1467,17 @@ async fn service_stats_uninitialized() {
     }
 }
 
+// Since KeyPair isn't in use for builtins anymore, we can't use this check as it is
```
kmd-fl, line 1470: @justprosh here is the test I was talking about yesterday

I'm not sure how to fix this test correctly

kmd-fl:
@folex  btw old services envs **are not supported** 

Smth like `FLUENCE_AQUA_IPFS...` and `FLUENCE_DECIDER_...`. I have no idea what is the full list of them and how exactly they should be integrated with the new layered configs Nick implemented.

Also, I'm not sure how important they are. Should they be implemented as a part of this PR or another

kmd-fl:
Added support for old system services ENVs

crates/nox-tests/tests/builtin.rs
```diff
@@ -1594,9 +1608,11 @@ async fn sign_invalid_tetraplets() {
 
 #[tokio::test]
 async fn sig_verify_invalid_signature() {
-    let kp = KeyPair::generate_ed25519();
-    let swarms =
-        make_swarms_with_builtins(1, "tests/builtins/services".as_ref(), Some(kp), None).await;
+    let swarms = make_swarms_with_cfg(1, |mut cfg| {
+        cfg.enabled_system_services = vec![Registry];
```
folex, line 1612: nice

crates/server-config/src/system_services_config.rs
```diff
@@ -16,15 +16,60 @@
 
 use super::defaults::*;
 use serde::{Deserialize, Serialize};
+use std::fmt::Formatter;
 
-#[derive(Clone, Serialize, Deserialize, Debug, Default)]
+#[non_exhaustive]
```
folex, line 21: why?

crates/system-services/src/lib.rs
```diff
@@ -80,159 +87,137 @@ impl SystemServiceDeployer {
...
+        match key {
+            AquaIpfs => self.deploy_aqua_ipfs(),
+            TrustGraph => self.deploy_trust_graph(),
+            Registry => self.deploy_registry().await,
+            Decider => {
+                self.deploy_connector()?;
+                self.deploy_decider().await
+            }
+            x => {
+                log::error!("installation of a service `{x}` is not implemented",);
```
folex, line 100:
```suggestion
                log::error!("unknown system service `{x}`",);
```

crates/system-services/src/lib.rs
```diff
@@ -80,159 +87,137 @@ impl SystemServiceDeployer {
...
-        self.deploy_connector()
-            .map_err(|e| JError::new(e.to_string()))?;
-        self.deploy_decider().await
+    pub async fn deploy_system_services(&self) -> Result<(), JError> {
+        for service in &self
+            .config
+            .enable
+            .iter()
+            .collect::<std::collections::HashSet<_>>()
+        {
```
folex, line 112:
```suggestion
        let enabled = self.config.enable.iter().collect::<std::collections::HashSet<_>>();
        for service in enabled {
```

crates/system-services/src/lib.rs
```diff
@@ -80,159 +87,137 @@ impl SystemServiceDeployer {
...
-            let spell_id = match deployed {
-                DeployedService::Existing(id) => {
-                    log::info!("found `{spell_name}` [{id}] spell, no need to redeploy");
-                    continue;
-                }
-                DeployedService::Created(id) => id,
-            };
-
-            log::info!("deployed `{spell_name}` [{spell_id}] spell");
+            let _deployed = self.deploy_system_spell(spell_distro).await?;
```
folex, line 169: ditto
```suggestion
            self.deploy_system_spell(spell_distro).await?;
```

crates/system-services/src/lib.rs
```diff
@@ -347,32 +330,64 @@ impl SystemServiceDeployer {
...
-                    log::warn!("couldn't stop the old spell {}: {err}", spell_distro.name);
-                }
+            ServiceUpdateStatus::NeedUpdate(spell_id) => {
+                log::info!("found existing spell `{spell_name}` [{spell_id}]; will redeploy",);
+                self.clean_old_spell(&spell_name, spell_id).await;
             }
-            FindSpell::NoUpdate(spell_id) => {
-                return Ok(DeployedService::Existing(spell_id));
+            ServiceUpdateStatus::NoUpdate(spell_id) => {
+                log::info!("found existing spell `{spell_name}` [{spell_id}]; no need to redeploy",);
```
folex, line 341: ditto 1 and 2

crates/system-services/src/lib.rs
```diff
@@ -347,32 +330,64 @@ impl SystemServiceDeployer {
...
+            let result: Result<_, JError> = try {
+                // Stop the spell to avoid re-subscription
+                self.call_service(&spell_id, "set_trigger_config", vec![json!(empty_config)])?;
+
+                // Unsubscribe spell from execution
+                self.spell_event_bus_api.unsubscribe(spell_id.clone()).await
+            };
+            if let Err(err) = result {
+                log::error!(
+                    "couldn't stop the old spell {spell_name} (will install new spell nevertheless): {err}",
```
folex, line 377: 👍

crates/system-services/src/lib.rs
```diff
@@ -444,28 +458,51 @@ impl SystemServiceDeployer {
...
+        let service_name = service_distro.name.clone();
+        let blueprint_id = self
+            .add_modules(service_distro)
+            .map_err(|e| JError::new(e.to_string()))?;
 
-        if let Some(service_id) = self.find_same_service(service_name.to_string(), &blueprint_id) {
-            return Ok(DeployedService::Existing(service_id));
+        match self.find_same_service(service_name.to_string(), &blueprint_id) {
+            ServiceUpdateStatus::NeedUpdate(service_id) => {
+                log::info!("found existing service {service_name} [{service_id}]; will redeploy");
```
folex, line 485: 1. Not clear to which System Service this `service_name` belongs. I'd add this info.
2. Let's move to DEBUG
3. Maybe use the tracing syntax here to pass fields? `log::debug!(service_id=service_id, service_name=service_name,system_service=system_service_name, "found existing service, will redeploy")`

Not sure if `service_name=service_name` could be shortened to `service_name`. @gurinderu do you know?

crates/system-services/src/lib.rs
```diff
@@ -444,28 +458,51 @@ impl SystemServiceDeployer {
...
+                    false,
+                );
+                if let Err(err) = result {
+                    log::error!(
+                        "couldn't remove the old service {service_name} (will install new service nevertheless): {err}",
+                    );
+                }
+            }
+            ServiceUpdateStatus::NoUpdate(service_id) => {
+                log::info!(
```
folex, line 500: debug

nox/src/node.rs
```diff
@@ -501,7 +501,7 @@ impl<RT: AquaRuntime> Node<RT> {
         // Note: need to be after the start of the node to be able to subscribe spells
         let deployer = self.system_service_deployer;
         if let Err(e) = deployer.deploy_system_services().await {
-            log::error!("installing system services failed: {}", e);
+            log::error!("deploying system services failed: {}", e);
```
folex, line 504: Will it show a list of failed services and corresponding errors? Will it fail on the first service or try all of them first?

folex (approved):

crates/system-services/src/lib.rs
```diff
@@ -347,32 +330,64 @@ impl SystemServiceDeployer {
...
-            FindSpell::NeedUpdate(spell_id) => {
-                let empty_config = TriggerConfig::default();
-                // Stop old spell
-                let result =
-                    self.call_service(&spell_id, "set_trigger_config", vec![json!(empty_config)]);
-                if let Err(err) = result {
-                    log::warn!("couldn't stop the old spell {}: {err}", spell_distro.name);
-                }
+            ServiceUpdateStatus::NeedUpdate(spell_id) => {
+                log::info!("found existing spell `{spell_name}` [{spell_id}]; will redeploy",);
```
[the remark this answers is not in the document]
kmd-fl: Atm we don't have connections between services and spells, so we can't really communicate it in a general manner. (If you meant that it would be nice to have a log like `"found existing spell 'registry-renew' with id ... of 'registry' service that needs update; will redeploy"`).

 _A bit offtop:_ In the future, I want some kind of common lib for every crate that would describe a _package_ that consists of services, spells, and their connections; I want to be able to create some kind of worker for this package to be able to control the package easily.

nox/src/node.rs
kmd-fl: It will fail on the first one.

crates/system-services/src/lib.rs
folex: spans are rather easy to use, you can always try

crates/nox-tests/tests/builtin.rs
kmd-fl: Hope, you'll fix it in the future

nox/src/node.rs
folex: there's a `stop` channel

kmd-fl: Is it okay for now? Can we

folex: `exit_outlet`, that is

kmd-fl: Yeah, I did

It's pretty straightforward, but my brain can't generate smth nice

kmd-fl: I tried but it didn't work for me :thinking:. Maybe I needed to wait longer though.

crates/system-service-deployer/src/lib.rs
kmd-fl: I'll do it in one of the next PRs. Created a task for [this](https://linear.app/fluence/issue/NET-490/update-spell-scripts-instead-of-redeploying-a-spell-in-system-service)

PR #1628, opened 2023-06-15
fix(particle-execution): do not clone particle on every AVM::call

folex:


aquamarine/src/actor.rs
```diff
@@ -154,16 +154,19 @@ where
             let waker = cx.waker().clone();
             // Schedule execution of functions
             self.functions
-                .execute(effects.particle.id.clone(), effects.call_requests, waker);
+                .execute(self.particle.id.clone(), effects.call_requests, waker);
```
kmd-fl, line 157: Why is this change okay?  Isn't it entirely different `particle_id`?

folex: it's the same particle_id, because actor processes only a single particle_id

kmd-fl (approved):

justprosh (approved):

PR #1629, opened 2023-06-19
feat: Migrate Registry to spell

gurinderu:


crates/system-services/Cargo.toml
```diff
@@ -8,7 +8,7 @@ edition = "2021"
 [dependencies]
 aqua-ipfs-distro = "=0.5.14"
 decider-distro = "=0.4.15"
-registry-distro = "=0.8.5"
+registry-distro = { git = "https://github.com/fluencelabs/registry.git", branch = "feature/net-256"}
```
kmd-fl, line 11: Maybe just publish the registry-distro and use the update version?

kmd-fl (approved):
Fine by me. Need to update to the latest system-service-deployer, though. Would be nice to merge this PR asap

crates/server-config/src/defaults.rs
```diff
@@ -224,6 +224,26 @@ pub fn default_decider_spell_period_sec() -> u32 {
     120
 }
 
+// 60 minutes
+pub fn default_registry_spell_period_sec() -> u32 {
+    3600
```
folex, line 229: what is this period for? others are clearly named, but this one i don't understand

gurinderu: it is a period for running a spell

folex: it's not really clear that these 4 values must be connected. ie, `default_registry_spell_period_sec` must always be a smallest common denominator.

crates/system-services/src/deployer.rs
```diff
@@ -149,12 +150,10 @@ impl Deployer {
     }
 
     async fn deploy_registry(&self) -> eyre::Result<()> {
-        let (registry_distro, registry_spell_distros) = Self::get_registry_distro();
+        let (registry_distro, registry_spell_distro) =
+            Self::get_registry_distro(self.config.registry.clone());
         let _deployed = self.deploy_service_common(registry_distro)?;
```
folex, line 155:
```suggestion
        self.deploy_service_common(registry_distro)?;
```

crates/system-services/src/deployer.rs
```diff
@@ -198,29 +197,30 @@ impl Deployer {
...
-        (distro, spells_distro)
+
+        let registry_config = registry_distro::RegistryConfig {
+            expired_interval: config.expired_period_sec,
+            renew_interval: config.renew_period_sec,
+            replicate_interval: config.replicate_period_sec,
+        };
+        let spell_distro = registry_distro::registry_spell(registry_config);
+        let mut trigger_config = TriggerConfig::default();
+        trigger_config.clock.start_sec = 1;
```
folex, line 214: looks like this deserves a comment. Why not keep it default `0`?

folex (approved):

crates/server-config/src/defaults.rs
gurinderu: Done

crates/system-services/src/deployer.rs
gurinderu: Used the same value as for decider

folex: why?

gurinderu: why not)

gurinderu: Added comment

PR #1647, opened 2023-06-22
fix(avm): avm 0.40, avm-server 0.32

folex:


rust-toolchain.toml
```diff
@@ -4,7 +4,3 @@ components = [
   "rustfmt",
   "clippy"
 ]
-targets = [
```
folex, line 7: i'm not sure if it is still needed.

let's see if that passes CI.

justprosh (approved):

crates/local-vm/src/local_vm.rs
```diff
@@ -123,6 +123,14 @@ pub fn client_functions(data: &HashMap<String, JValue>, args: Args) -> ClientFun
             outcome: FunctionOutcome::Empty,
             returned: None,
         },
+        ("run-console", "print") => {
+            println!("run-console: {}", json!(args.function_args));
```
kmd-fl, line 127: why double?

kmd-fl (approved):

crates/local-vm/src/local_vm.rs
folex: because I'm not sure when logs are enabled and when not

PR #1648, opened 2023-06-23
chore: Renaming things

nahsi:
Please approve and I will merge all PR at once, reverting to using main/master branches.


.github/workflows/container.yml
```diff
@@ -3,7 +3,7 @@ name: Publish docker container
 on:
   workflow_call:
     inputs:
-      rust-peer-sha:
+      nox-sha:
```
gurinderu, line 6: why not just "checksum" ?

nahsi: There could be other checksums, like builtins etc in a future.

PR #1649, opened 2023-06-23
fix(system-services): update system spell without redeploying [fixes NET-490]

kmd-fl:
Also updated the spell distro version to the one allowing to update of the spell script.

_Note_: before merging need to update the decider which isn't working properly after the update, so for now it's a draft.


crates/system-services/src/deployer.rs
```diff
@@ -312,30 +312,99 @@ impl Deployer {
         }
     }
 
-    async fn deploy_system_spell(&self, spell_distro: SpellDistro) -> eyre::Result<ServiceStatus> {
+    async fn deploy_system_spell_common(
```
justprosh, line 315: You forgot to rename `self.deploy_system_spell(registry_spell_distro).await?;` on line 156

crates/system-services/src/deployer.rs
```diff
@@ -312,30 +313,99 @@ impl Deployer {
...
```
folex, line 316: why `_common`?

crates/system-services/src/deployer.rs
```diff
@@ -312,30 +313,99 @@ impl Deployer {
...
-                self.clean_old_spell(&spell_name, spell_id).await;
+                if let Err(err) = self.update_spell(&spell_distro, &spell_id).await {
+                    tracing::warn!(
+                        spell_id,
+                        spell_name,
+                        "can't update a spell (will remove the old spell and deploy a new one): {err}"
+                    );
+                    self.clean_old_spell(&spell_name, spell_id).await;
+                    let spell_id = self.deploy_spell_common(spell_distro).await?;
+                    tracing::info!(spell_name, spell_id, "redeployed a system spell",);
```
folex, line 336:
```suggestion
                    tracing::info!(spell_name, spell_id, "redeployed a system spell");
```
Does log show `spell_name` and `spell_id`?

crates/system-services/src/deployer.rs
```diff
@@ -312,30 +313,99 @@ impl Deployer {
...
                     spell_id,
-                    "found existing spell that needs to be updated; will remove tha old spell and deploy a new one",
+                    "found existing spell that needs to be updated; will try to update script, trigger config and init data",
                 );
-                self.clean_old_spell(&spell_name, spell_id).await;
+                if let Err(err) = self.update_spell(&spell_distro, &spell_id).await {
+                    tracing::warn!(
+                        spell_id,
+                        spell_name,
+                        "can't update a spell (will remove the old spell and deploy a new one): {err}"
```
folex, line 332:
```suggestion
                        "can't update a spell (will redeploy it): {err}"
```

crates/system-services/src/deployer.rs
```diff
@@ -312,30 +313,99 @@ impl Deployer {
...
+                    );
+                    self.clean_old_spell(&spell_name, spell_id).await;
+                    let spell_id = self.deploy_spell_common(spell_distro).await?;
+                    tracing::info!(spell_name, spell_id, "redeployed a system spell",);
+                    return Ok(ServiceStatus::Created(spell_id));
+                }
+                tracing::info!(spell_name, spell_id, "updated a system spell");
```
folex, line 339:
```suggestion
                } else {
                    tracing::info!(spell_name, spell_id, "updated a system spell");
                }
```
wdyt about making it like that? or maybe even do a `match`. Right now, it's hard to follow, because `updated` message comes after `redeployed` block

crates/system-services/src/deployer.rs
kmd-fl: yes

folex (approved):

PR #1663, opened 2023-07-06
fix(system-services): allow overriding binary paths for system services [fixes NET-503]

kmd-fl:
## Description

Allow overriding binary paths for some system services:
- aqua-ipfs (path to the `ipfs` binary)
- connector (path to the `curl` binary)

Overriding happens in the node config `Config.toml` like that:
```toml
# Note that we also must allow the node to use those binaries
allowed_binaries = [
        "/home/maar/curl",
        "/home/maar/ipfs"
]

[system_services.connector]
curl_binary_path = "/home/maar/curl"

[system_services.aqua_ipfs]
ipfs_binary_path = "/home/maar/ipfs"

```

## Motivation

Mainly for local development.

The default paths for these services are the following;
- `/usr/bin/ipfs`
- `/usr/bin/curl`

However, some team members run Nox on MacBooks, where those paths are invalid, the default ones there are, for example, `/opt/homebrew/bin/ipfs`.

It should help with the development process.

## Proposed Changes

* Modify the config to add paths to binaries:
  - add `ipfs_binary_path` to `AquaIpfsConfig`
  - introduce `ConnectorConfig` with `curl_binary_path` field
* On deployment, change paths in the Marine service configuration provided by the system service crates
  - At the moment, the easiest way to do it is to find the necessary fields by their names (like module name `curl_adapter` and binary name `curl`).

## Additional Notes

 - If we find this feature useful in the future, we will find a more general approach with the help of the crates.
 - Note that at the moment, we **do not redeploy** services when their marine config changes. I guess, we need a linear task for this.

## Checklist

## Reviewer Checklist


gurinderu (approved):

crates/server-config/src/system_services_config.rs
```diff
@@ -78,17 +81,26 @@ pub struct AquaIpfsConfig {
     pub external_api_multiaddr: String,
     #[serde(default = "default_ipfs_multiaddr")]
     pub local_api_multiaddr: String,
+    #[serde(default)]
+    pub ipfs_binary_path: Option<String>,
```
folex, line 85: I would expect default to be `/usr/bin/ipfs`. Where does the default value come from?

kmd-fl: It's value for overriding. By default, we don't override anything.

folex: Hm, I see.

I think it might be better to always override it, and have `/usr/bin/ipfs` by default. That would make it far easier to find the path of the binary. 

Because otherwise, you have to know to go to the distro crate, find Toml file there and look it up – and that's a pretty long discovery path, only a few people know how to do that :)

While if we'd have this path in `defaults.rs`, it is immediately obvious, and we can even generate documentation out of that.

folex:
Maybe it makes sense to add these binaries to `allowed_binaries` automatically? 🤔

justprosh (approved):

crates/system-services/src/deployer.rs
```diff
@@ -640,4 +660,19 @@ impl Deployer {
             .add_blueprint(AddBlueprint::new(service_distro.name, hashes))?;
         Ok(blueprint_id)
     }
+
+    fn apply_binary_path_override(
```
folex, line 664: Let's add a doc, plz

folex (approved):

folex (approved):

PR #1755, opened 2023-08-12
fix(env, log)!: RUST_LOG and other envs parsing

folex:
## Description

- Fix parsing certain config fields of Array type from ENV variables.
- Fix RUST_LOG parsing: trim whitespaces
- Load RUST_LOG only once, I believe it was twice before (in main.rs and layers.rs). @gurinderu, wdyt?
- ⚠️ Crash if RUST_LOG is incorrect. @nahsi, wdyt?

## Motivation

A lot of lists in config did not parse from ENV before. Tabs after newlines in RUST_LOG didn't parse.

## Checklist

## Reviewer Checklist


nox/src/layers.rs
```diff
@@ -14,9 +14,13 @@ pub fn log_layer<S>(log_config: &Option<LogConfig>) -> impl Layer<S>
 where
     S: Subscriber + for<'span> LookupSpan<'span>,
 {
+    let rust_log = std::env::var("RUST_LOG")
```
gurinderu, line 17: why do we need that?

nahsi (approved):

nox/src/layers.rs
folex: it's in the pr description. it trims whitespaces, otherwise it doesn't work with multiline indentation

sorcerer/src/spell_builtins.rs
```diff
@@ -454,3 +454,20 @@ pub(crate) fn store_response(
         ))
     })
 }
+
+#[cfg(test)]
+mod tests {
+    use fluence_spell_dtos::trigger_config::TriggerConfig;
+
+    #[test]
+    fn deserialize_trigger_config() {
```
kmd-fl, line 463: why do we need this test here? why no in the fluence_spell_dtos or smth?

folex: because that's where this logic happens

kmd-fl: but deserialization logic implemented in the trigger config crate, so the test should there, shouldn't it? it looks like a random test tbh

folex: it was useful to me, it does not cost anything to be here. if anyone would experiment with processing triggers in builtins like I did, she would find that test useful.

folex: i'll remove it

crates/server-config/src/resolved_config.rs
```diff
@@ -206,7 +206,11 @@ pub fn load_config_with_args(
         .prefix_separator("_")
         .separator("__")
         .list_separator(",")
-        .with_list_parse_key("allowed_binaries");
+        .with_list_parse_key("allowed_binaries")
+        .with_list_parse_key("external_multiaddresses")
+        .with_list_parse_key("bootstrap_nodes")
+        .with_list_parse_key("listen_config.listen_multiaddrs")
```
gurinderu, line 212: probably we need tests for each of this types

nox/src/layers.rs
folex: what escaping? take a look at Makefile, it doesn't work otherwise.

gurinderu (approved):

crates/server-config/src/resolved_config.rs
folex: would be nice to automate that ofc

PR #1770, opened 2023-08-25
feat(system-services, tests): composable and overridable system services [NET-530, NET-531]

kmd-fl:
## Description

Decompose system services crate to be able to override and extend used system services.

## Motivation

ATM the system services are build in the system service deployer. To modify them or to add new ones, you need to change the deployer itself. In this PR we introduce a way to define a system service separately from the deployer and deploy any provided set of services. Also, we extend the tests interface to be able to modify and created custom system services.

## Proposed Changes
1. Introduced the `PackageDistro` structure to define a system service with all aux spells and services. I call a _package_, not a _system service_ (since it's not a single service but a set of services) or a _worker_ (since it's not installed on a separate worker) for clarity of the purpose, but we way rename it to `SystemServiceDistro`, although I find the _package_ concept more general. Probably, we can use this concept in FCLI to make re-usable and composable services.
2. Put definition of the default system services (`aqua-ipfs`, `registry` and so on) outside of the deployer.
3. Made it possible to override default system services and introduce new ones.
4. Extended the test interface to override or extend used system services.

## Additional Notes
1. In the future, I plan to create `PackageDistro` for each system service in the corresponding system service distro crate, and import it.
2. We need to reorganize the cargo configuration of the project to be able to publish `create-swarm` crate, but do it in a separate PR.


crates/created-swarm/Cargo.toml
```diff
@@ -19,6 +19,7 @@ script-storage = { workspace = true }
 fs-utils = { workspace = true }
 air-interpreter-fs = { workspace = true }
 toy-vms = { workspace = true }
+system-services = { workspace = true }
```
gurinderu, line 22: is it really needed to define that explicitly because it is a transitive dependency in the `nox` crate?

kmd-fl: I define it explicily to use `SystemServiceDistros` to create default system service distros.

crates/server-config/src/system_services_config.rs
```diff
@@ -36,6 +36,10 @@ impl ServiceKey {
             Self::Decider,
         ]
     }
+
+    pub fn from_string(name: &str) -> Option<ServiceKey> {
```
gurinderu, line 40: Why not Result? You can lose the error on the upper level

crates/system-services/src/deployer.rs
```diff
@@ -102,261 +66,50 @@ impl Deployer {
...
-        args: Vec<JValue>,
-    ) -> eyre::Result<()> {
-        let result = self.services.call_function(
-            self.root_worker_id,
-            service_id,
-            function_name,
-            args,
-            None,
-            self.root_worker_id,
-            Duration::from_millis(DEPLOYER_TTL),
```
gurinderu, line 324: you can make a constant with a duration time like a 
```
const DEPLOYER_TTL: Duration = Duration::from_millis(60_000);
```

crates/system-services/src/distro.rs
```diff
@@ -0,0 +1,273 @@
...
+
+                let timestamp = now_millis::now_sec();
+                for cert_chain in &certs.certs {
+                    call(
+                        name.clone(),
+                        "insert_cert".to_string(),
+                        vec![json!(cert_chain), json!(timestamp)],
+                    )?;
+                }
+                log::info!("initialized `{}` [{id}] service", name);
```
gurinderu, line 112: is the tracing avaliable here?

crates/server-config/src/system_services_config.rs
kmd-fl: 1) it doesn't matter, 2) there can only be one error -- no such key, 3) it doesn't matter

crates/nox-tests/tests/services.rs
```diff
@@ -16,13 +17,118 @@
...
 use base64::{engine::general_purpose::STANDARD as base64, Engine};
 use connected_client::ConnectedClient;
 use created_swarm::{make_swarms, make_swarms_with_cfg};
 use service_modules::load_module;
+use system_services::{PackageDistro, ServiceDistro};
+
+#[tokio::test]
+async fn test_system_service_override() {
+    // We need to include bytes, not read them, since the ServiceDistro expects module bytes as `&'static [u8]`
+    // It's unnecessary to allow not static links or even vectors since in real life in real life we need only this
```
justprosh, line 33:
```suggestion
    // It's unnecessary to allow not static links or even vectors since real life we need only this
```

crates/system-services/src/lib.rs
```diff
@@ -1,6 +1,125 @@
...
+/// Call service functions. Accepts
+/// - service name
+/// - function name
+/// - function arguments
+///
+/// Restriction:
+/// The functions called via this callback must return a result with execution status in field `status: bool`
+/// and error message in the field `error: string`.
+/// Otherwise, the output will be consider invalid.
+pub type CallService = Box<dyn Fn(String, String, Vec<Value>) -> eyre::Result<()> + Send + Sync>;
```
justprosh, line 26: Maybe add aliases for ServiceName and FunctionName?

crates/system-services/src/lib.rs
```diff
@@ -1,6 +1,125 @@
...
+    Box<dyn Fn(&CallService, DeploymentStatus) -> eyre::Result<()> + Send + Sync>;
+
+/// Package distribution description
+/// Contains enough information about all services and spells used by the package for installation
+/// It's not the same thing as a worker since this kind of package doesn't require to be installed
+/// in a separate scope
+#[derive(Clone)]
+pub struct PackageDistro {
+    /// High-level name of the package.
+    /// For the system services the field is supposed to be a string of `ServiceKey` enum
```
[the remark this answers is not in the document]
kmd-fl:
```rust
pub enum ServiceKey {
    AquaIpfs,
    TrustGraph,
    Registry,
    Decider,
}
```
https://github.com/fluencelabs/nox/blob/3a3f776de4ed9013cac3e2ebfb40620695371528/crates/server-config/src/system_services_config.rs#L23

Since we want to use it in tests, I thought it would be better to use an arbitrary `String` as a name, not a restricted `ServiceKey`.

kmd-fl:
@justprosh @nick let's merge it maybe?

crates/system-services/src/distro.rs
```diff
@@ -0,0 +1,284 @@
...
+    let service_distro = ServiceDistro {
+        modules: connector_service_distro.modules,
+        config: marine_config,
+        name: connector_service_distro.name.to_string(),
+    };
+
+    let wallet_key = match decider_config.wallet_key.clone() {
+        // TODO: set default wallet key somewhere in nox-distro, etc
+        //None => return Err(eyre!("Decider enabled, but wallet_key is not set. Please set it via env FLUENCE_ENV_CONNECTOR_WALLET_KEY or in Config.toml")),
+        None => "0xfdc4ba94809c7930fe4676b7d845cbf8fa5c1beae8744d959530e5073004cf3f".to_string(),
```
folex, line 250: I think throwing error makes more sense. Because `decider` won't work with that key anyway, and then Nox Admin will have hard time learning about that.

So fail early would be better than "fail somewhere deep after deploying someone else's deals".

kmd-fl: @justprosh you had some problems with this place, can you comment to resolve what to do with it?

justprosh: We should carefully disable decider for all non-relevant e2e tests and provide wallet addr where we need it. Also we should disable decider deployment by default, otherwise you just can't start nox container (also we have nothing about this in docs or readme)

folex: > otherwise you just can't start nox container (also we have nothing about this in docs or readme)

Maybe we should provide a CLI flag to pass that private key. So that it's documented in `--help`, at least.

And maybe also a CLI flag to enable/disable system services. The ENV-based approach is too cumbersome to be the only one, I think.

> we should disable decider deployment by default

I'm not sure about that. Being able to deploy deals is crucial to Nox. It's just that now Nox requires wallet key and chain connection to work.

justprosh: Yeah, required flag sounds suitable

PR #1797, opened 2023-09-20
fix(spell-bus): resubscribe spell on double subscription

kmd-fl:
## Description

The `spell-event-bus` interface doesn't forbid subscription of the same spell on the same event on their own; the code around it was supposed to avoid such situations. However, there's a possibility that the `system-service-deployer` can deploy system spells before the `sorcerer` resubscribes old spells (created on the previous run of the node), so the `sorcerer` can also resubscribe the newly created spells since we don't distinguish active and not-yet-active spells.

This problem occurred during running nox in tests via `CreatedSwarm` and doesn't yet affect the fully-fledged nox.  

I think the best approach to the situation is to create a spell registry that would account for the spell's state: unsubscribed, subscribed, ended execution, etc. However, this approach requires more work and design and will be done in the future, but at the moment, we can introduce a simpler solution that won't resubscribe the running spells. 

## Motivation

The `spell-event-bus` doesn't track the list of active spells and will subscribe the already active spells as new ones. This can lead to running spells more often than configured. 

This problem can occur only at the start of the nox when the sorcerer resubscribes spells; the public interfaces (used by decider and direct deployment) aren't affected.

## Proposed Changes

* Keep track of active subscriptions in the bus itself
* If we receive a doubling subscription, resubscribe the spell. I think there are no _nice_ solutions since we can't properly report on the events, but for now, it may be fine.


crates/spell-event-bus/src/bus.rs
```diff
@@ -97,17 +97,27 @@ impl PartialOrd for Scheduled {
 struct SubscribersState {
     subscribers: PeerEventSubscribers,
     scheduled: BinaryHeap<Scheduled>,
+    active: HashSet<Arc<SpellId>>,
```
gurinderu, line 100: why do u need the Arc here?

kmd-fl: I used it because it was used in `Scheduled` and `PeerEventSubscribers`. I guess, `Rc` should be enough

crates/spell-event-bus/src/bus.rs
```diff
@@ -97,17 +97,27 @@ impl PartialOrd for Scheduled {
...
     fn new() -> Self {
         Self {
             subscribers: PeerEventSubscribers::new(),
             scheduled: BinaryHeap::new(),
+            active: HashSet::new(),
         }
     }
 
     fn subscribe(&mut self, spell_id: SpellId, config: &SpellTriggerConfigs) -> Option<()> {
+        if self.active.contains(&spell_id) {
```
gurinderu, line 113: I assume it is better to use Result type here without unsubscribe here and check and log this error on upper level

crates/spell-event-bus/src/bus.rs
gurinderu: can it be used in a different threads?

gurinderu: checked.

folex (approved):

crates/spell-event-bus/src/bus.rs
folex: Same question

crates/spell-event-bus/src/bus.rs
```diff
@@ -582,4 +601,80 @@ mod tests {
...
+            },
+            || {
+                bus.abort();
+            },
+        );
+    }
+
+    #[tokio::test]
+    async fn test_resubscribing_same_spell() {
+        //log_utils::enable_logs();
```
justprosh, line 646:
```suggestion
```

crates/spell-event-bus/src/bus.rs
```diff
@@ -582,4 +601,80 @@ mod tests {
...
+        let (bus, api, mut event_receiver) = SpellEventBus::new(None, vec![]);
+        let bus = bus.start();
+        let _ = api.start_scheduling().await;
+        //let event_stream = UnboundedReceiverStream::new(event_receiver);
```
justprosh, line 650:
```suggestion
```

PR #1811, opened 2023-09-25
feat: verify particle signatures [NET-539]

justprosh:
## Description
Verify particle signatures and drop incorrectly signed ones.

## Motivation
Particle signatures are crucial for security and authentication.

## Proposed Changes
- drop incorrectly signed particles
- particle signature is a concatenation of:
    - `id` as bytes
    - `timestamp` u64 as little-endian bytes
    - `ttl` u32 as little-endian bytes
    - `script` as bytes

## Checklist

## Reviewer Checklist


aquamarine/src/error.rs
```diff
@@ -52,6 +58,11 @@ impl AquamarineApiError {
             AquamarineApiError::ParticleExpired { particle_id } => Some(particle_id),
             AquamarineApiError::OneshotCancelled { particle_id } => Some(particle_id),
             AquamarineApiError::ExecutionTimedOut { particle_id, .. } => Some(particle_id),
+            // Should it be `None`  considering usage of signature as particle id?
+            // It can compromise valid particles into thinking they are invalid.
+            // But still there can be a case when signature was generated wrong
+            // and client will never know about it.
```
kmd-fl, line 64: Not sure what the problem is exactly, but it feels like we need to deliver an understandable and debuggable error message no matter what, or we will die in pain.

kmd-fl (approved):

aquamarine/src/error.rs
folex: I think this that's the place to decide on what should be sent back to the client.

If we'd be to keep the behaviour of sending errors back to client _over network_, than I'd suggest to have a `valid_particle` function. It would signify if we can safely "trust" the particle not to be a part of Amplification Attack. And then, depending on that knowledge, send error to the client or not.

But it seems that we'd be removing that behaviour altogether.

aquamarine/src/plumber.rs
```diff
@@ -92,6 +92,16 @@ impl<RT: AquaRuntime, F: ParticleFunctionStatic> Plumber<RT, F> {
             return;
         }
 
+        if let Err(err) = particle.verify() {
+            tracing::info!(target: "signature_failed", particle_id = particle.id, "Particle signature verification failed: {err:?}");
```
folex, line 96:
```suggestion
            tracing::warn!(target: "signature_failed", particle_id = particle.id, "Particle signature verification failed: {err:?}");
```

folex (approved):

aquamarine/src/plumber.rs
```diff
@@ -92,6 +92,16 @@ impl<RT: AquaRuntime, F: ParticleFunctionStatic> Plumber<RT, F> {
             return;
         }
 
+        if let Err(err) = particle.verify() {
```
folex, line 95: for future readers: verify is implemented here https://github.com/fluencelabs/nox/blob/7e658ce092e274b358c751da516149438126d67a/particle-protocol/src/particle.rs#L128

folex:
Can you check on the ed25519 exploit, please? https://github.com/MystenLabs/ed25519-unsafe-libs/tree/main/ed25519-chalkias-exploit

Check if it applies to us on current setup, and follow their recommendation regarding using `verify_strict`. https://docs.rs/ed25519-dalek/latest/ed25519_dalek/struct.VerifyingKey.html#method.verify_strict

aquamarine/src/plumber.rs
gurinderu, line 96: why not warn with target signature

aquamarine/src/error.rs
```diff
@@ -15,6 +15,7 @@
  */
 
 use humantime::FormattedDuration;
+use particle_protocol::ParticleError;
```
folex, line 18: sort imports

aquamarine/src/error.rs
folex: It's marked as resolved, yet it still returns None. I don't understand.

script-storage/Cargo.toml
folex, line 1: remove?

script-storage/src/script_storage.rs
folex, line 1: remove?

PR #1824, opened 2023-10-09
fix(tests): Allow override system services config in tests [NET-576]

kmd-fl:
## Motivation
IPFS and curl paths can differ in CI, so we need some way to override them. Also, it may be nice to control other configs also.  

## Proposed Changes
Add the `override_system_services_config: Option<SystemServicesConfig>` field in `SwarmConfig`.
Requires importing `server-config` crate to get access to the definition and default values. 
Proposed use:
```rust
make_swarm_with_cfg(1, move |mut cfg| {
   let mut services_cfg = SystemServicesConfig::default();
   services_cfg.aqua_ipfs.ipfs_binary_path = "/new/path/to/ipfs".to_string()
   cfg.override_system_services_config = Some(services_cfg);
   cfg
}) 
```

### DIFF
--- a/crates/created-swarm/src/swarm.rs
+++ b/crates/created-swarm/src/swarm.rs
@@ -244,6 +244,7 @@ pub struct SwarmConfig {
     pub allowed_binaries: Vec<String>,
     pub enabled_system_services: Vec<String>,
     pub extend_system_services: Vec<system_services::PackageDistro>,
+    pub override_system_services_config: Option<system_services_config::SystemServicesConfig>,
     pub http_port: u16,
     pub connector_api_endpoint: Option<String>,
 }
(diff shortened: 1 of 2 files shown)


crates/created-swarm/src/swarm.rs
```diff
@@ -244,6 +244,7 @@ pub struct SwarmConfig {
     pub allowed_binaries: Vec<String>,
     pub enabled_system_services: Vec<String>,
     pub extend_system_services: Vec<system_services::PackageDistro>,
+    pub override_system_services_config: Option<system_services_config::SystemServicesConfig>,
```
folex, line 247: Is `SystemServicesConfig` struct exported from `created-swarm` crate? Would users of the crate be able to instantiate that config?

folex (approved):

crates/created-swarm/src/swarm.rs
folex: Yeah, I saw that. Let's reexport so it is self-suficient.

PR #1858, opened 2023-10-26
fix(metrics): measure call time and wait time separately

folex:
## Description
- Report actor mailbox size and actor counts in every poll. Before this PR, it was only reported when there are `interpretation_stats`.
- Measure lock wait time separately from marine execution time in AppService
- Measure time it takes for service call execution to be scheduled on Blocking Pool in Actor
- Separate call execution time measurement from schedule wait time in Actor

## Motivation
Including queuing time from latency measurements skews measurements when system is saturated. In such form, measurement doesn't make much sense: it will obviously grow as more work arrives.

Instead, execution latency should be measured separately, to allow for applying eg Little's Law to estimate system's throughput  at given execution latencies.

Another benefit of that separation is that we get a clear picture of actual Marine performance.

## Concerns

We have two sets of metrics that describe execution latencies. 

One is one the higher level: in Plumber-Actor. It measures Builtin calls latencies as close as possible, which is good. However, for Service Calls its measurements include RWLock and Mutex latencies, which makes these measurements skewed.

To the rescue comes another measurement, lower in the stack, at AppService level. That one (now) measures pure Marine call latency, and time it took to wait for a Service lock to be available.

The mixture of measurement sites and absence of clear naming makes it hard to navigate metrics.

The dashboards in Grafana must be upgraded with great care, then. And documenting these caveats is crucial for external users' ability to observe Nox through these metrics.

### DIFF
--- a/crates/peer-metrics/src/services_metrics/external.rs
+++ b/crates/peer-metrics/src/services_metrics/external.rs
@@ -86,7 +86,8 @@ pub struct ServicesMetricsExternal {
     pub modules_in_services_count: Histogram,
 
     /// Service call time
-    pub call_time_msec: Family<ServiceTypeLabel, Histogram>,
+    pub call_time_sec: Family<ServiceTypeLabel, Histogram>,
+    pub lock_wait_time_sec: Family<ServiceTypeLabel, Histogram>,
     pub call_success_count: Family<ServiceTypeLabel, Counter>,
     pub call_failed_count: Family<ServiceTypeLabel, Counter>,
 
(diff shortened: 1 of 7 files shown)


crates/peer-metrics/src/services_metrics/external.rs
```diff
@@ -182,13 +183,20 @@ impl ServicesMetricsExternal {
             "number of modules per services",
         );
 
-        let call_time_msec: Family<_, _> = register(
+        let call_time_sec: Family<_, _> = register(
             sub_registry,
             Family::new_with_constructor(|| Histogram::new(execution_time_buckets())),
             "call_time_msec",
```
folex, line 189: I don't understand why it's called `msec`

gurinderu (approved):

PR #1879, opened 2023-11-02
chore: Move Dockerfile from nox-distro [fixes FLU-436, FLU-493]

nahsi:
## Description
1. merge of nox-distro repo with nox and deprecation of nox-distro repo
2. get rid of S6 and  get rid of IPFS inside container

## Motivation
1. It is easier from CI point of view and also more logical to have docker container build in the same repo with the artifact.
2. Container should have only one daemon running inside.

## Proposed Changes
- no IPFS sidecar
- nox container version == nox version
- nox container is released on nox release
- no more flavours like minimal ipfs or rich
- no more ceramic
- updated default Config.toml with comments and various cli in allowed effectors
- addition of FLUENCE_UID conf var that is not present in Config.toml, only docker

## Checklist
- [x] The code follows the project's coding conventions and style guidelines.
- [x] All tests related to the changes have passed successfully.
- [x] Documentation has been updated to reflect the changes (if applicable).
- [x] All new and existing unit tests have passed.
- [x] I have self-reviewed my code and ensured its quality.
- [x] I have added/updated necessary comments to aid understanding.

## Reviewer Checklist


entrypoint.sh
```diff
@@ -0,0 +1,11 @@
+#! /usr/bin/env bash
```
folex, line 1: i don't think it's a good idea to put these to /

deploy/docker-compose/docker-compose.yml
```diff
@@ -0,0 +1,93 @@
+name: nox-network
```
folex, line 1: why do you need these in this repo?

nahsi: This was in nox-distro, since we duplicating nox-distro I had to move it somethere.
Another thing is that I was inspired by deployment examples in [Loki repo](https://github.com/grafana/loki/tree/main/production).


BUT it would be better to have deployment examples in a separate repo since we will have a release process for some of them (helm, ansible role).

So I guess I will create another repo for deployment examples and add docker-compose example there.

entrypoint.sh
nahsi: It is easier when it is in root - `docker build` command will work with default context and without specifying path to Dockerfile.

But of course I can hide it inside `./docker` directory.

.github/workflows/build.yml
```diff
@@ -77,7 +74,7 @@ jobs:
         uses: Swatinem/rust-cache@v2
         with:
           cache-directories: ~/.cache/sccache
-          shared-key: nox-${{ matrix.arch }}
+          shared-key: nox-${{ matrix.cache-hack }}
```
gurinderu, line 77: cache-key sounds much better

nahsi: I will change `cache-hack` to `arch` after this PR is approved and ready to merge.
This hack is needed to use old cache in this PR in e2e because arch value was aarch64 and x86_64 and now it is amd64 and arm64. And cache is only generated on merge on master.
It is to speed up e2e in other words, that is why it is a hack

folex (approved):

PR #1888, opened 2023-11-08
feat(nox): add new flags for system services [NET-547, NET-548]

kmd-fl:
## Description

Add two new flags:
- `--wallet-key KEY`
- `--enable-system-services=LIST`
   - `--enable-system-services=all` to enable all available services
   - `--enable-system-services=service1,service2` to enable only services from the list
   - `--enable-system-services=none` to disable all services (disabling now doesn't work, it will be introduce by some other task)

Also, make nox fail when a wallet key isn't set when decider is enabled. 

## Motivation
To be user-friendly

## Additional Notes

These flags are introduced via the existing mechanism of layer configs, so we don't have a lot of control here.

### DIFF
--- a/crates/server-config/src/args.rs
+++ b/crates/server-config/src/args.rs
@@ -240,7 +240,6 @@ impl Serialize for SystemServicesArgs {
                     struct_serializer.serialize_field::<Vec<ServiceKey>>("enable", &vec![])?;
                 }
             }
-            //struct_serializer.serialize_field("generate_on_absence", generate_on_absence)?;
         }
         if let Some(wallet_key) = &self.wallet_key {
             #[derive(Serialize)]
--- a/crates/server-config/src/args.rs
+++ b/crates/server-config/src/args.rs
@@ -14,7 +14,9 @@
  * limitations under the License.
  */
 
+use crate::system_services_config::ServiceKey;
 use crate::LogFormat;
+use clap::error::ErrorKind;
 use clap::{Args, Parser};
 use config::{ConfigError, Map, Source, Value};
 use serde::ser::SerializeStruct;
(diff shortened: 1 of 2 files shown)


crates/server-config/src/args.rs
```diff
@@ -150,6 +152,112 @@ pub enum TracingType {
...
+    #[arg(
+        long,
+        id = "WALLET_KEY",
+        help = "The private wallet key for signing transactions for joining deals",
+        help_heading = "System services configuration"
+    )]
+    wallet_key: Option<String>,
+}
+
+impl Serialize for SystemServicesArgs {
```
folex, line 212: Why do you need `Serialize`?

folex: Ah, because `DerivedArgs` has it. But why does `DerivedArgs` need `Serialize`? 🤔

particle-builtins/src/builtins.rs
```diff
@@ -786,6 +786,7 @@ where
 
         let alias: String = Args::next("alias", &mut args)?;
         let service_id: String = Args::next("service_id", &mut args)?;
+        log::warn!("add_alias {:?} {:?}", alias, service_id);
```
folex, line 789: remove?

kmd-fl: yeah, committed by mistake

crates/server-config/src/args.rs
kmd-fl: it's serialized to toml and used to update the node's config when certain flags are passed

folex: Tough

folex (approved):

PR #1891, opened 2023-11-09
feat(subnet): Update resolver to new contracts [NET-619]

kmd-fl:
## Description
Tested the nox and decider with new [deal lifecycle update](https://github.com/fluencelabs/cli/pull/593).

On deploy to stage/testnet/kras, need to set new addresses in the configs.

## Proposed Changes
The format of the call was changed. Now the function named `getComputeUnits` and the result doesn't have the `index` field (which was ignored anyway)

## Additional Notes
What I did:
- Built fcli from the PR above. Started local network.
- Started local nox with the updated matcher contract address.
- Registered local nox.
- Deployed a deal from the quickstart. 
- Re-deployed to update
- Looked at logs

Results:
- decider found and installed the deal
- decider couldn't find the update since it didn't move the queried window (created a task in linear). After manually changing the block number in KV, decider found and installed the update
- `deal logs` didn't work; needed to change `subnet.resolve`. After the change, it worked.


gurinderu (approved):

crates/subnet-resolver/src/resolve.rs
```diff
@@ -109,7 +104,7 @@ pub fn resolve_subnet(deal_id: String, api_endpoint: &str) -> SubnetResolveResul
         // Description of the `getPATs` function from the `chain.workers` smart contract on chain
```
justprosh, line 104:
```suggestion
        // Description of the `getComputeUnits` function from the `chain.workers` smart contract on chain
```

justprosh (approved):

folex (approved):

.github/workflows/e2e.yml
```diff
@@ -88,6 +88,6 @@ jobs:
   flox:
     needs:
       - nox-snapshot
-    uses: fluencelabs/flox/.github/workflows/tests.yml@main
+    uses: fluencelabs/flox/.github/workflows/tests.yml@deal-lifecycle
     with:
       nox-image: "${{ needs.nox-snapshot.outputs.nox-image }}"
```
nahsi, line 93:
```suggestion
    uses: fluencelabs/flox/.github/workflows/tests.yml@main
    with:
      nox-image: "${{ needs.nox-snapshot.outputs.nox-image }}"
      ref: deal-lifecycle
```

PR #1968, opened 2023-12-19
feat(plumber): allow root to access deactivated workers [fixes NET-695]

kmd-fl:
## Description
Allow root to access deactivated workers.

## Motivation
Decider needs to be able to update the worker spell's KV even if the worker is deactivated.

### DIFF
--- a/aquamarine/src/plumber.rs
+++ b/aquamarine/src/plumber.rs
@@ -120,6 +120,7 @@ impl<RT: AquaRuntime, F: ParticleFunctionStatic> Plumber<RT, F> {
             && !self
                 .key_manager
                 .is_management(particle.particle.init_peer_id)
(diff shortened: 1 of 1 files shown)


justprosh (approved):

aquamarine/src/plumber.rs
```diff
@@ -120,6 +120,7 @@ impl<RT: AquaRuntime, F: ParticleFunctionStatic> Plumber<RT, F> {
             && !self
                 .key_manager
                 .is_management(particle.particle.init_peer_id)
+            && !self.key_manager.is_host(particle.particle.init_peer_id)
```
folex, line 123: could you please move these conditions to variables and add comments?

folex (approved):

PR #2012, opened 2024-01-04
feat(workers): Create a ThreadPool for each worker [fixes NET-688]

gurinderu:
## Description
Create a ThreadPool for each worker.

## Motivation
We needed to isolate each worker on a separate CU

## Related Issue(s)
[NET-688](https://linear.app/fluence/issue/NET-688/create-a-threadpool-for-each-worker)

## Proposed Changes
- Added a runtime creation on worker creation 
- On the actor creation get a worker runtime handle if it exists 
- Introduced a Spawner entity that allows to shift executions on the different pools 

## Checklist
- [x] The code follows the project's coding conventions and style guidelines.
- [x] All tests related to the changes have passed successfully.
- [x] Documentation has been updated to reflect the changes (if applicable).
- [x] All new and existing unit tests have passed.
- [x] I have self-reviewed my code and ensured its quality.
- [x] I have added/updated necessary comments to aid understanding.

## Reviewer Checklist

### DIFF
--- a/aquamarine/src/actor.rs
+++ b/aquamarine/src/actor.rs
@@ -23,17 +23,16 @@ use std::{
 };
 use tracing::{instrument, Instrument, Span};
 
-use fluence_keypair::KeyPair;
-use fluence_libp2p::PeerId;
-use particle_execution::{ParticleFunctionStatic, ServiceFunction};
-use particle_protocol::{ExtendedParticle, Particle};
-use tokio::runtime::Handle;
-
 use crate::deadline::Deadline;
 use crate::particle_effects::RoutingEffects;
 use crate::particle_executor::{FutResult, ParticleExecutor};
 use crate::particle_functions::{Functions, SingleCallStat};
+use crate::spawner::Spawner;
 use crate::{AquaRuntime, InterpretationStats, ParticleDataStore, ParticleEffects};
+use fluence_keypair::KeyPair;
+use fluence_libp2p::PeerId;
+use particle_execution::{ParticleFunctionStatic, ServiceFunction};
+use particle_protocol::{ExtendedParticle, Particle};
 
 struct Reusables<RT> {
     vm_id: usize,
@@ -66,7 +65,7 @@ pub struct Actor<RT, F> {
     current_peer_id: PeerId,
     key_pair: KeyPair,
     data_store: Arc<ParticleDataStore>,
-    runtime_handle: Handle,
+    spawner: Spawner,
     deal_id: Option<String>,
 }
 
@@ -81,8 +80,8 @@ where
         current_peer_id: PeerId,
         key_pair: KeyPair,
         data_store: Arc<ParticleDataStore>,
-        runtime_handle: Handle,
         deal_id: Option<String>,
+        spawner: Spawner,
     ) -> Self {
         Self {
             deadline: Deadline::from(particle),
@@ -98,7 +97,7 @@ where
             current_peer_id,
             key_pair,
             data_store,
-            runtime_handle,
+            spawner,
             deal_id,
         }
     }
@@ -160,15 +159,18 @@ where
 
             self.future.take();
 
+            let spawner = self.spawner.clone();
             let waker = cx.waker().clone();
             // Schedule execution of functions
-            self.functions.execute(
-                self.runtime_handle.clone(),
-                self.particle.id.clone(),
-                effects.call_requests,
-                waker,
-                parent_span.clone(),
-            );
+            {
+                self.functions.execute(
+                    spawner,
+                    self.particle.id.clone(),
+                    effects.call_requests,
+                    waker,
+                    parent_span.clone(),
+                );
+            }
 
             let effects = RoutingEffects {
                 particle: ExtendedParticle::linked(
@@ -226,33 +228,34 @@ where
 
         let waker = cx.waker().clone();
         let data_store = self.data_store.clone();
-        let handle = self.runtime_handle.clone();
         let key_pair = self.key_pair.clone();
         let peer_id = self.current_peer_id;
 
         let (async_span, linking_span) =
             self.create_spans(call_spans, ext_particle, particle.id.as_str());
 
+        let spawner = self.spawner.clone();
         self.future = Some(
             async move {
-                handle
-                    .spawn(async move {
-                        let res = vm
-                            .execute(data_store, (particle.clone(), calls), peer_id, key_pair)
-                            .in_current_span()
-                            .await;
+                let res = vm
+                    .execute(
+                        spawner,
+                        data_store,
+                        (particle.clone(), calls),
+                        peer_id,
+                        key_pair,
+                    )
+                    .in_current_span()
+                    .await;
 
-                        waker.wake();
+                waker.wake();
 
-                        let reusables = Reusables {
-                            vm_id,
-                            vm: res.runtime,
-                        };
+                let reusables = Reusables {
+                    vm_id,
+                    vm: res.runtime,
+                };
 
-                        (reusables, res.effects, res.stats, linking_span)
-                    })
-                    .await
-                    .expect(" Failed to shift execution to handle")
+                (reusables, res.effects, res.stats, linking_span)
             }
             .instrument(async_span)
             .boxed(),
--- a/aquamarine/src/actor.rs
+++ b/aquamarine/src/actor.rs
@@ -66,7 +66,7 @@ pub struct Actor<RT, F> {
     current_peer_id: PeerId,
     key_pair: KeyPair,
     data_store: Arc<ParticleDataStore>,
-    handle: Handle,
+    runtime_handle: Handle,
     deal_id: Option<String>,
 }
 
@@ -81,7 +81,7 @@ where
         current_peer_id: PeerId,
         key_pair: KeyPair,
         data_store: Arc<ParticleDataStore>,
-        handle: Handle,
+        runtime_handle: Handle,
         deal_id: Option<String>,
     ) -> Self {
         Self {
@@ -98,7 +98,7 @@ where
             current_peer_id,
             key_pair,
             data_store,
-            handle,
+            runtime_handle,
             deal_id,
         }
     }
@@ -163,7 +163,7 @@ where
             let waker = cx.waker().clone();
             // Schedule execution of functions
             self.functions.execute(
-                self.handle.clone(),
+                self.runtime_handle.clone(),
                 self.particle.id.clone(),
                 effects.call_requests,
                 waker,
@@ -226,7 +226,7 @@ where
 
         let waker = cx.waker().clone();
         let data_store = self.data_store.clone();
-        let handle = self.handle.clone();
+        let handle = self.runtime_handle.clone();
         let key_pair = self.key_pair.clone();
         let peer_id = self.current_peer_id;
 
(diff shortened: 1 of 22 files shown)


aquamarine/src/particle_functions.rs
```diff
@@ -135,6 +145,7 @@ impl<F: ParticleFunctionStatic> Functions<F> {
     #[instrument(level = tracing::Level::INFO, skip_all)]
     fn call(
         &self,
+        handle: Handle,
```
folex, line 148: Is it possible to not pass `Handle` here, and use `spawn_blocking_on` outside the `call` function? 

To keep this layer separate from the threading model.

for example, `call` could return a future which you would submit to `handle`

aquamarine/src/particle_functions.rs
```diff
@@ -188,35 +199,38 @@ impl<F: ParticleFunctionStatic> Functions<F> {
...
-                        // Builtins were called, return their outcome
-                        outcome => outcome,
-                    };
-                    // How much time it took to execute the call
-                    // TODO: Time for ParticleFunction includes lock time, which is not good. Low priority cuz ParticleFunctions are barely used.
-                    let call_time = schedule_wait_start.elapsed() - schedule_wait_time;
-                    (outcome, call_kind, call_time, schedule_wait_time)
-                })
-            })
+            .spawn_blocking_on(
```
folex, line 202: why do you use `spawn_blocking_on` instead of `spawn_on`? It's CPU bounded here, not IO bounded, afaik

aquamarine/src/plumber.rs
```diff
@@ -149,12 +150,17 @@ impl<RT: AquaRuntime, F: ParticleFunctionStatic> Plumber<RT, F> {
                 let deal_id = self.workers.get_deal_id(worker_id).ok();
                 let data_store = self.data_store.clone();
                 key_pair.map(|kp| {
+                    let handle = self
+                        .workers
+                        .get_handle(worker_id)
+                        .unwrap_or(Handle::current());
```
folex, line 156: What would `Handle::current` return here? The root? if so, let's leave a comment at least.

ideally, it would be great to hold a reference to root handle, and pass it here explicitly, ie `self.root_thread_pool.clone()` instead of `Handle::current`. 

context-dependent code is harder to understand, cuz you never know if you missed anything and it might return a wrong thing

crates/workers/src/error.rs
```diff
@@ -137,4 +137,9 @@ pub enum WorkersError {
     },
     #[error("Keypair for peer_id {0} not found")]
     KeypairNotFound(PeerId),
+    #[error("Failed to create runtime: {err}")]
```
folex, line 140: Would be great to add worker id here as well, so it is possible to associate error with a specific worker

crates/workers/src/workers.rs
```diff
@@ -17,6 +18,8 @@ pub struct WorkerInfo {
     pub creator: PeerId,
     /// A read-write lock indicating whether the worker is active.
     pub active: RwLock<bool>,
+
+    pub cu_count: usize,
```
folex, line 22: a comment please

crates/workers/src/workers.rs
```diff
@@ -30,7 +33,25 @@ pub struct Workers {
     /// Key storage for managing worker key pairs.
     key_storage: Arc<KeyStorage>,
     /// Scope information used to determine the host and manage key pairs.
-    scope: Scope,
+    scope: PeerScope,
+    /// Mapping of worker IDs to worker runtime.
+    runtimes: RwLock<HashMap<WorkerId, Runtime>>,
```
folex, line 38:
```suggestion
    thread_pools: RwLock<HashMap<WorkerId, Runtime>>,
```
maybe?

not sure how else to convey that it's the tokio runtime, and it's about threads. Otherwise it might be Docker runtime or WASM runtime, or something else.

crates/workers/src/workers.rs
```diff
@@ -30,7 +33,25 @@ pub struct Workers {
...
+}
+
+pub struct CreateWorkerParams {
```
folex, line 41:
```suggestion
pub struct WorkerParams {
```

crates/workers/src/workers.rs
```diff
@@ -51,24 +72,42 @@ impl Workers {
...
             let deal_id = w.deal_id.clone();
+            let cu_count = w.cu_count;
             worker_infos.insert(worker_id, w.into());
             worker_ids.insert(deal_id, worker_id);
+
+            let runtime = tokio::task::spawn_blocking(move || {
+                let runtime = tokio::runtime::Builder::new_multi_thread()
+                    .thread_name(format!("worker-pool-{}", worker_id))
+                    .worker_threads(cu_count)
+                    .max_blocking_threads(cu_count)
```
folex, line 93: please leave an elaborate comment on how exactly each of these thread pools will be used.

for example, 
```
// we create a total of cu_count*2 threads, and that makes sense because we assume X(???) number of threads per logical processor
// we don't verify that number against what OS sees because ???
let runtime = ...
...

// worker threads will be used to execute service calls
// because blah blah blah, so that blablabla
.worker_threads(cu_count)

// blocking threads will be used to execute ???
// because ??? so that ???
.max_blocking_threads(cu_count)
```

crates/workers/src/workers.rs
```diff
@@ -51,24 +72,42 @@ impl Workers {
...
+                    .build()
+                    .map_err(|err| WorkersError::CreateRuntime { err })
+                    .unwrap();
```
folex, line 96: panic in the constructor, is it the only way? maybe return an error and bubble it up, so that it finally can be handled in Aqua?

crates/workers/src/workers.rs
```diff
@@ -106,20 +145,28 @@ impl Workers {
...
                         if worker_ids.contains_key(&deal_id) {
                             return Err(WorkersError::WorkerAlreadyExists { deal_id });
                         }
 
+                        let runtime = tokio::runtime::Builder::new_multi_thread()
+                            .worker_threads(params.cu_count)
+                            .max_blocking_threads(params.cu_count)
+                            .build()
+                            .map_err(|err| WorkersError::CreateRuntime { err })?;
+
```
folex, line 166: Why take locks before checking the error and creating a runtime?

Maybe like this instead? The suggestion simply moves lines 154, 155 to 166, 167.
```suggestion
                        if worker_ids.contains_key(&deal_id) {
                            return Err(WorkersError::WorkerAlreadyExists { deal_id });
                        }

                        let runtime = tokio::runtime::Builder::new_multi_thread()
                            .worker_threads(params.cu_count)
                            .max_blocking_threads(params.cu_count)
                            .build()
                            .map_err(|err| WorkersError::CreateRuntime { err })?;
                            
                        let mut worker_infos = self.worker_infos.write();
                        let mut runtimes = self.runtimes.write();
```

crates/workers/src/workers.rs
```diff
@@ -523,6 +597,8 @@ mod tests {
             .get_worker_id("deal_id_1".to_string())
             .expect("Failed to get worker id");
         assert_eq!(worker_id_1, worker_id);
+        //tokio doesn't allow to drop runtimes in async context, so shifting workers drop to the blocking thread
```
folex, line 600:
```suggestion
        // tokio doesn't allow to drop runtimes in async context, so shifting workers drop to the blocking thread
```

crates/workers/src/workers.rs
```diff
@@ -562,14 +642,20 @@ mod tests {
...
+            ))
             .await;
 
         assert!(res.is_err());
         assert_eq!(
             res.err().unwrap().to_string(),
             "Worker for deal_id_1 already exists"
-        )
+        );
+        //tokio doesn't allow to drop runtimes in async context, so shifting workers drop to the blocking thread
```
folex, line 657:
```suggestion
        // tokio doesn't allow to drop runtimes in async context, so shifting workers drop to the blocking thread
```

aquamarine/src/particle_functions.rs
gurinderu: So, this code still is used for the common service calls, before that they was on blocking pool. Decided to have the same behaviour

aquamarine/src/actor.rs
```diff
@@ -79,6 +81,7 @@ where
         current_peer_id: PeerId,
         key_pair: KeyPair,
         data_store: Arc<ParticleDataStore>,
+        handle: Handle,
```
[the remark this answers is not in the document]
gurinderu: maybe runtime_handle ?

crates/workers/src/error.rs
gurinderu: Done

crates/workers/src/workers.rs
gurinderu: No, it was mistake. FIxed

crates/workers/src/workers.rs
gurinderu: Fixed

crates/workers/src/workers.rs
gurinderu: Fixed

crates/workers/src/workers.rs
gurinderu: Fixed

aquamarine/src/plumber.rs
gurinderu: Fixed

crates/workers/src/workers.rs
gurinderu: Fixed

aquamarine/src/actor.rs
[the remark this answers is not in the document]
folex: runtime is very generic and overloaded, too

aquamarine/src/particle_functions.rs
folex: ah, understood. I hope if you succeed in moving `Handle` a few layers higher, then you might have just enough flexibility to run service calls from `root` worker on `blocking`, and service calls from other workers on usual threads.

folex (approved):
nice!

aquamarine/src/particle_executor.rs
```diff
@@ -83,12 +87,14 @@ impl<RT: AquaRuntime> ParticleExecutor for RT {
         tracing::trace!(target: "execution", particle_id = particle_id, "Executing particle");
 
         let prev_data = data_store
+            .clone()
```
folex, line 90: why?

aquamarine/src/spawner.rs
```diff
@@ -0,0 +1,196 @@
...
+        let task_name = format!("Call function {}:{}", self.worker_id, &function_identity);
+
+        let builder = tokio::task::Builder::new().name(task_name.as_str());
+
+        builder
+            .spawn_on(fut, &self.runtime_handle)
+            .expect("Failed to spawn a task")
+    }
+
+    fn spawn_avm_call<F, R>(&self, fut: F) -> JoinHandle<F::Output>
```
folex, line 181: it's not a future, so name `fut` is rather misleading here

folex: `closure` or something like that would make more sense

folex (approved):
What is `enum_dispatch`, and how it affects the code?

gurinderu:
> enum_dispatch

https://crates.io/crates/enum_dispatch

aquamarine/src/spawner.rs
gurinderu: Fixed

PR #2039, opened 2024-01-26
feat(app-services): Support worker-id in CallParameters [NET-717]

kmd-fl:
## Description

Support the new CallParamters with `worker_id` field. 

## Motivation
This is required for the new spell kv permission system.

## Proposed Changes

1. When calling services on a worker, set `worker_id` to the worker_id of the worker and `host_id` to the underlying `host_id` of the host
2. When installing spells, set initial data as a spell (we emulate CallParameters as if a spell is doing the call). This is a dubious decision, but now we don't see any nicer ones.


sorcerer/src/spell_builtins.rs
```diff
@@ -84,7 +84,15 @@ pub async fn install_spell(
     // Save the script to the spell
     spell_service_api.set_script(params.clone(), script)?;
     // Save init_data to the spell's KV
-    spell_service_api.update_kv(params.clone(), init_data)?;
+    let self_particle_id = format!("spell_{spell_id}_0");
+    let init_kv_params = CallParams::new(
```
folex, line 88: Why do you need different `CallParams` for `set_script` and `update_kv`? 🤔

kmd-fl: Not sure if we _need_ it. We need special CallParams for update_kv, but yeah, we may use the same callparams for everything

kmd-fl: If you want some changes, I can do them in a different PR, when I'll be integrating the Spell KV Restrictions update, okay? I fight for some time to make the tests pass

gurinderu (approved):

sorcerer/src/spell_builtins.rs
kmd-fl: I mark this resolve to merge rn, since I need to check the PR in the spell repo

PR #2098, opened 2024-02-21
feat(particle-vault)!: introduce new particle vault path format

kmd-fl:
## Description

Introduce new particle vault features: new path format and write-only `/tmp/vault`

## Motivation

Need to secure the particle vault to prevent unauthorized access to the vaults of others since we map shared vault directory into services (due to marine limitation) 

## Proposed Changes
1. Use `worker-id` (`host-id` on host) as a part of the particle vault name
2. In marine service, map `/tmp/vault -> vault_dir/$worker_id`
3. Store `particle_token` in Actor to be able to clean the vault (:c) 
4. The vault name format now is `/tmp/vault/$particle-id-$particle-token/`
5. Create a worker vault with write-only permission and a particle vault with RW permission for the cleanup procedure (it can't remove a directory recursively without the read permission). I create it on each service creation since it's the easiest place, but we'll need to move it 
6. Note that Worker Particle Vault **won't be cleared**, and if a worker is removed, the directory remains. It should be fixed with the ephemeral feature, I guess. 
7. Also, fix the problem with [Spell KV Restriction](https://github.com/fluencelabs/nox/pull/2094): deployer can't re-deploy a spell due to the lack of privileges

## Additional Notes

TODO:
- [x] need to use new aqua-ipfs that use the updated particle vault path format https://github.com/fluencelabs/aqua-ipfs/pull/165
- [x] need to merge updated Installation Spell https://github.com/fluencelabs/spell/pull/276
- [x] need to use new decider with updated Installation Spell  https://github.com/fluencelabs/decider/pull/236
- [x] manually check the `ls` problem:
       - I created a simple service that does ` fs::read_dir("/tmp/vault") `, it returned Permission denied.
       - I tried to call `set_readonly("/tmp/vault", ..)`, and the result was the error: `operation not supported on this platform`
       - I tried to use PermissionsExt to use `chmod` directly, and it didn't compile. So, it seems impossible to change permissions inside the service.
       - Also, `ls` and `chmod` commands are considered mounted binaries so they won't be accessible.


aquamarine/src/actor.rs
```diff
@@ -74,10 +76,13 @@ where
     RT: AquaRuntime,
     F: ParticleFunctionStatic,
 {
+    // TODO: temporary (I hope), need to do smth clever with particle_token
+    #[allow(clippy::too_many_arguments)]
```
folex, line 80: that's fine

crates/nox-tests/tests/file_share/Cargo.toml
```diff
@@ -11,6 +11,6 @@ path = "src/main.rs"
 [workspace]
 
 [dependencies]
-marine-rs-sdk = "0.7.1"
-rand = { workspace = true }
+marine-rs-sdk = "0.13.0"
+rand = "0.8.5"
```
gurinderu, line 15:
```suggestion
rand = { workspace = true }
```

kmd-fl: it doesn't buld this way for me

aquamarine/src/actor.rs
```diff
@@ -63,6 +63,8 @@ pub struct Actor<RT, F> {
     /// Particles and call results will be processed in the security scope of this peer id
     /// It's either `host_peer_id` or local worker peer id
     current_peer_id: PeerId,
+    /// TODO: for the fucking clean-up, I don't think we need it here!
```
folex, line 66: Let's remove curse words :)

kmd-fl: lol, sry

particle-services/src/app_services.rs
```diff
@@ -993,13 +1001,16 @@ impl ParticleAppServices {
 
     fn create_app_service(
         &self,
+        current_peer_id: PeerId,
         blueprint_id: String,
         service_id: String,
     ) -> Result<AppService, ServiceError> {
+        // TODO: Better to create once, when the worker is created but I don't have mental capacity to do it rn
+        self.vault.initialize_worker(current_peer_id)?;
```
kmd-fl, line 1009: here

folex, line 1009:
```suggestion
        // Create Particle File Vault for Worker
        self.vault.initialize_worker(current_peer_id)?;
```

PR #2124, opened 2024-02-26
feat: introduce DealId type

justprosh:


crates/types/src/deal_id.rs
```diff
@@ -0,0 +1,124 @@
...
+
+impl PartialEq<&str> for &DealId {
+    fn eq(&self, other: &&str) -> bool {
+        self.0 == DealId::normalize(other)
+    }
+}
+
+impl PartialEq<String> for DealId {
+    fn eq(&self, other: &String) -> bool {
+        self.0 == DealId::normalize(other)
```
folex, line 51: So you copy on every comparison? Maybe it's better to force everything to do `normalize` once?

folex (approved):
I feel like it's a bit overcomplicated in terms of number of trait impls. 

Probably you could get away with just a constructor that always does normalise.

Otherwise, lgtm.

PR #2126, opened 2024-03-01
feat: introduce dev mode with allowing all effectors

kmd-fl:
## Description

Introduce developer mod to allow any effectors

## Motivation
Need less secure node for local developing.

## Proposed Changes
- add new flag `--dev-mode` to enter the developer mode
- add mounted binaries configuration with binary names and their pathes to be able to generate marine configs for them

## Additional Notes
I'll make it look nicer when there's time to think.  

## Example
To enable dev mode you can either use the `--dev-mode` flag or write it in the Config.toml file
```toml
[dev_mode]
enable = true
[dev_mode.binaries]
ls = "/bin/ls
```
By default, the dev mode is disabled. Also, we provide settings for curl and ipfs by default. 

You can also provide only config for your binaries and then regulate the mode by the `--dev-mode` flag
```
[dev_mode.binaries]
ls = "/bin/ls
``` 

Or you can set the dev mode enabled in the config without providing the setting for the mounted binaries. In this case the default ones will be used.
```
[dev_mode]
enable = true
```


crates/server-config/src/args.rs
```diff
@@ -50,23 +50,23 @@ pub struct RootKeyPairArgs {
     )]
     path: Option<PathBuf>,
     #[arg(
-        short('f'),
-        long("keypair-format"),
-        value_parser(["ed25519", "secp256k1", "rsa"]),
-        id = "ROOT_KEY_FORMAT",
-        help_heading = "Node keypair",
-        display_order = 12,
+    short('f'),
```
kmd-fl, line 53: wtf

crates/server-config/src/effectors_config.rs
```diff
@@ -0,0 +1 @@
+
```
justprosh, line 1: ?

crates/server-config/src/node_config.rs
```diff
@@ -603,3 +602,20 @@ fn default_effectors_config() -> EffectorsConfig {
...
 }
+
+#[derive(Clone, Deserialize, Serialize, Derivative)]
+#[derivative(Debug)]
+pub struct DevModeConfig {
+    #[serde(default)]
+    pub enable: bool,
+    /// Mounted binaries mapping: binary name (used in the effector modules) to binary path
+    #[serde(default = "default_binaries_mapping")]
+    pub binaries: HashMap<String, String>,
```
justprosh, line 613: Better to have `HashMap<String, PathBuf>`

crates/server-config/src/services_config.rs
```diff
@@ -64,28 +68,55 @@ impl ServicesConfig {
...
-                            Ok(false) => log::warn!("effector `{path_str}` does not exist"),
-                            _ => {}
-                        };
-                        (name, path.to_path_buf())
-                    })
-                    .collect::<_>();
-                (cid, effector)
-            })
-            .collect::<_>();
+        let allowed_effectors = if is_dev_mode {
```
justprosh, line 77: Why do you have this if here? It's not so transparent. I suggest to left there fields as it is. You have `if is_dev_mode` in other parts of your code

justprosh (approved):

crates/server-config/src/services_config.rs
kmd-fl: Just to avoid checks for the config that won't be in use. Can remove it, no problem

crates/server-config/src/effectors_config.rs
kmd-fl: removed

PR #2127, opened 2024-03-01
feat: introduce dev mode with allowing all effectors 

kmd-fl:
## Description

Introduce developer mod to allow any effectors

## Motivation
Need less secure node for local developing.

## Proposed Changes
- add new flag `--dev-mode` to enter the developer mode
- add mounted binaries configuration with binary names and their pathes to be able to generate marine configs for them

## Additional Notes
I'll make it look nicer when there's time to think.  

## Example
To enable dev mode you can either use the `--dev-mode` flag or write it in the Config.toml file
```toml
[dev_mode]
enable = true
[dev_mode.binaries]
ls = "/bin/ls"
```
By default, the dev mode is disabled. Also, we provide settings for curl and ipfs by default. 

You can also provide only config for your binaries and then regulate the mode by the `--dev-mode` flag
```toml
[dev_mode.binaries]
ls = "/bin/ls"
``` 

Or you can set the dev mode enabled in the config without providing the setting for the mounted binaries. In this case the default ones will be used.
```toml
[dev_mode]
enable = true
```


justprosh (approved):

kmd-fl:
Re-opened PR to fix broken GitHub

PR #2141, opened 2024-03-08
feat: restrict builtins, aqua-ipfs, worker-spell alias [NET-551,NET-730,NET-729,NET-792]

kmd-fl:
## Description
Forbid usage of several builtins, aqua-ipfs service and worker-spell alias.

## Motivation
Only Host and Worker Spells (and a host manager) should be able to call important built-ins like Srv.create or Worker.create.  

## Proposed Changes
1. Make some builtins available only to Host, Worker-Spell, and Host Manager (Here's [the list of all protected builtins](https://www.notion.so/fluencenetwork/Builtins-List-dc366787660f44faaba367e282fc4d52))
2. Make `aqua-ipfs` available only to Host, Worker-Spell, and Host Manager.
3. Allow calling services and spells `worker-spell` only Host and Host Manager
4. Fix incorrect spell resubscription for workers

## Additional Notes
- We shouldn't forget to design a nice permission system when we'll be discussing nox re-design later. It's really easy to miss now an important `if is_host` when it's needed.


particle-builtins/src/builtins.rs
```diff
@@ -294,7 +296,12 @@ where
                 }
                 wrap_unit(Ok(()))
             }
-
+            // NOTE: must be the same as in ServiceKey::AquaIpfs
+            // TODO: come up with some better way of restricting service access like aqua-ipfs
+            ("aqua-ipfs", _) => {
```
gurinderu, line 301: what will happen if I will deploy a service aqua-ipfs in the worker scope ?

particle-builtins/src/lib.rs
```diff
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+#![feature(async_closure)]
```
gurinderu, line 17: Hm, I don't see any async closure in the particle builtins module, could you provide the line where it is located ?

kmd-fl: No, I need to live it.

particle-builtins/src/builtins.rs
kmd-fl: It won't be available, yeah, need to fix it

kmd-fl: fixed, checked via fcli

particle-builtins/src/lib.rs
gurinderu: still here )

gurinderu (approved):

particle-builtins/src/lib.rs
kmd-fl: no, it isn't

PR #2158, opened 2024-03-10
fix(listener): save pending units on start; refresh state on subs drop

justprosh:


crates/chain-listener/src/listener.rs
```diff
@@ -155,9 +154,25 @@ impl ChainListener {
         }
     }
 
+    async fn refresh_commitment_params(&mut self) -> eyre::Result<()> {
+        let init_params =
```
folex, line 158: So you don't do anything with them? Why `&mut self`?

justprosh: fixed

crates/chain-listener/src/listener.rs
```diff
@@ -155,11 +157,33 @@ impl ChainListener {
...
+                    tracing::info!(target: "chain-listener", "Error getting Commitment initial params: {err}");
+                    err
+                })?;
+
+        self.difficulty = init_params.difficulty;
+        self.init_timestamp = init_params.init_timestamp;
+        self.global_nonce = init_params.global_nonce;
+        self.epoch_duration = init_params.epoch_duration;
+        self.current_epoch = init_params.current_epoch;
+
```
folex, line 175: is it ok that you don't persist anything here?

justprosh: yep, i get it from chain every time

PR #2168, opened 2024-03-15
feat(listener): restart ws client + refactoring

justprosh:


crates/chain-listener/src/listener.rs
```diff
@@ -314,155 +416,99 @@ impl ChainListener {
...
+        );
 
+        Ok(())
+    }
 
-                            if let Err(err) = self.poll_deal_statuses().await {
-                                tracing::warn!(target: "chain-listener", "Failed to poll deal statuses: {err}");
-                            }
+    async fn refresh_subscriptions(&mut self) -> Result<(), client::Error> {
+        if !self.ws_client.is_connected() {
```
folex, line 476: Are you sure RestartNeeded always causes websocket to set `is_connected` to false? 🤔

folex: Like, could it be that RestartNeeded happened, but you will not recreate WsClient due to this?

I guess you could read `client.error` 🤔

justprosh: i will recreate, i catch restart error on subscribe

crates/chain-listener/src/listener.rs
```diff
@@ -314,155 +416,99 @@ impl ChainListener {
...
+            self.ws_client =
+                ChainListener::create_ws_client(&self.listener_config.ws_endpoint).await?;
```
folex, line 478:
```suggestion
                // Ws Client creation retries through backoff inside, will fail only after many attempts
                // So it is OK that its error will propagate out of the loop
                ChainListener::create_ws_client(&self.listener_config.ws_endpoint).await?;
```

folex (approved):
nice!

PR #2208, opened 2024-03-28
feat(metrics): add metrics in chain-listener [NET-806]

kmd-fl:
## Description

Add some metrics to the ChainListener module. 

## Proposed Changes
Add a pack of metrics:
```rust
pub struct ChainListenerMetrics {
    // how many request Nox sends to ccp
    ccp_requests_total: Counter,
    // how many success replies Nox receives from ccp
    // an error is either error from ccp or connection errors
    ccp_replies_total: Counter,
    // how long we wait for a reply from ccp
    ccp_request_duration_msec: Histogram,
    // how many proofs we submitted
    cpp_proofs_submitted: Counter,
    // how many proofs we failed to submit
    cpp_proofs_submit_failed: Counter,
    // how many proofs transaction are ok
    ccp_proofs_tx_success: Counter,
    // how many proofs transaction are failed
    ccp_proofs_tx_failed: CounterWithExemplar<TxLabel>,
    // How many blocks we have received from the newHead subscription
    blocks_seen: CounterWithExemplar<BlockLabel>,
    // How many block we manage to process while processing the block
    blocks_processed: Counter,
}
```

## Additional Notes

We were discussing some refactoring, and I did the tiniest amount of refactoring, but I think we should do it in a separate PR.

### DIFF
--- a/crates/chain-listener/src/lib.rs
+++ b/crates/chain-listener/src/lib.rs
@@ -2,6 +2,8 @@
 #![feature(try_blocks)]
 #![feature(extract_if)]
 #![feature(btree_extract_if)]
+#![feature(result_option_inspect)]
+#![feature(async_closure)]
 
 extern crate core;
 
--- a/crates/chain-listener/src/lib.rs
+++ b/crates/chain-listener/src/lib.rs
@@ -3,7 +3,6 @@
 #![feature(extract_if)]
 #![feature(btree_extract_if)]
 #![feature(result_option_inspect)]
-#![feature(async_closure)]
 
 extern crate core;
 
--- a/crates/chain-listener/src/listener.rs
+++ b/crates/chain-listener/src/listener.rs
@@ -2,7 +2,7 @@ use alloy_primitives::{Address, FixedBytes, Uint, U256};
 use alloy_sol_types::SolEvent;
 use backoff::Error::Permanent;
 use std::collections::{BTreeMap, BTreeSet, HashMap};
-use std::future::pending;
+use std::future::{pending, Future};
 use std::ops::Add;
 use std::path::PathBuf;
 use std::process::exit;
@@ -27,7 +27,7 @@ use libp2p_identity::PeerId;
 use serde::de::DeserializeOwned;
 use serde_json::{json, Value};
 use tokio::task::JoinHandle;
-use tokio::time::interval;
+use tokio::time::{interval, Instant};
 use tokio_stream::wrappers::IntervalStream;
 use tokio_stream::StreamExt;
 
@@ -39,6 +39,7 @@ use chain_connector::{
 use chain_data::{parse_log, peer_id_to_hex, Log};
 use core_manager::types::{AcquireRequest, WorkType};
 use core_manager::{CoreManager, CoreManagerFunctions, CUID};
+use peer_metrics::ChainListenerMetrics;
 use server_config::{ChainConfig, ChainListenerConfig};
 use types::DealId;
 
@@ -53,25 +54,33 @@ pub struct ChainListener {
     listener_config: ChainListenerConfig,
 
     chain_connector: Arc<ChainConnector>,
+    // To subscribe to chain events
     ws_client: WsClient,
+
     ccp_client: Option<CCPRpcHttpClient>,
+
     core_manager: Arc<CoreManager>,
 
     host_id: PeerId,
 
+    // These settings aren't changed
+    // We refresh them on some errors, but it's enough to get them only on start without refreshing
     difficulty: Difficulty,
+    // when the contract was deployed
     init_timestamp: U256,
-    global_nonce: GlobalNonce,
-    current_epoch: U256,
     epoch_duration: U256,
     min_proofs_per_epoch: U256,
     max_proofs_per_epoch: U256,
+    // These settings are changed each epoch
+    global_nonce: GlobalNonce,
+    current_epoch: U256,
 
     proof_counter: BTreeMap<CUID, U256>,
     current_commitment: Option<CommitmentId>,
 
+    // the compute units that are in the commitment
     cc_compute_units: BTreeMap<CUID, ComputeUnit>,
-
+    // the compute units that are in deals
     active_deals: BTreeMap<DealId, CUID>,
 
     /// Resets every epoch
@@ -79,11 +88,16 @@ pub struct ChainListener {
     pending_proof_txs: Vec<(String, CUID)>,
     persisted_proof_id_dir: PathBuf,
 
+    // TODO: move out to a separate struct, get rid of Option
+    // Subscriptions that are polled when we have commitment
     unit_activated: Option<Subscription<JsonValue>>,
     unit_deactivated: Option<Subscription<JsonValue>>,
+    // Subscriptions that are polled always
     heads: Option<Subscription<JsonValue>>,
     commitment_activated: Option<Subscription<JsonValue>>,
     unit_matched: Option<Subscription<JsonValue>>,
+
+    metrics: Option<ChainListenerMetrics>,
 }
 
 async fn poll_subscription<T>(s: &mut Option<Subscription<T>>) -> Option<Result<T, client::Error>>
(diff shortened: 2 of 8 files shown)


crates/chain-listener/src/listener.rs
```diff
@@ -53,37 +54,50 @@ pub struct ChainListener {
...
     ccp_client: Option<CCPRpcHttpClient>,
+
     core_manager: Arc<CoreManager>,
 
     host_id: PeerId,
 
+    // These settings aren't changed
+    // We refresh them on some errors, but it's enough to get them only on start without refreshing
     difficulty: Difficulty,
+    // when the contract was deployed
```
justprosh, line 69: Probably better to say that it's timestamp of the beginning of the 1st epoch

crates/chain-listener/src/listener.rs
```diff
@@ -53,37 +54,50 @@ pub struct ChainListener {
...
     min_proofs_per_epoch: U256,
     max_proofs_per_epoch: U256,
+    // These settings are changed each epoch
+    global_nonce: GlobalNonce,
+    current_epoch: U256,
 
     proof_counter: BTreeMap<CUID, U256>,
     current_commitment: Option<CommitmentId>,
 
+    // the compute units that are in the commitment
```
justprosh, line 81:
```suggestion
    // the compute units that are in the commitment and not in deals
```
Because all CUs in deals in CC also, but don't need to produce proofs

crates/chain-listener/src/listener.rs
```diff
@@ -698,6 +721,9 @@ impl ChainListener {
             self.refresh_subscriptions().await?;
         }
 
+        // TODO: Why `current_commitment` is updated here? It looks like a deliberate choice.
+        // If `subscribe_unit_events` fails, the `refresh_subscribptions` will resubscribe to
+        // the `unit_activate` and `unit_deactivate` events, but for the older `current_commitment`.
```
justprosh, line 726: good point, let's move it to line 718

crates/chain-listener/src/listener.rs
```diff
@@ -1024,6 +1057,7 @@ impl ChainListener {
     }
 
     async fn submit_proof(&mut self, proof: CCProof) -> eyre::Result<()> {
+        // TODO: why can this happen?
```
justprosh, line 1060: if Unit moved to Deal and shortly after that (but before cc refresh) ccp found proof for it

crates/peer-metrics/src/chain_listener.rs
```diff
@@ -0,0 +1,155 @@
...
+pub struct ChainListenerMetrics {
+    // how many request Nox sends to ccp
+    ccp_requests_total: Counter,
+    // how many success replies Nox receives from ccp
+    // an error is either error from ccp or connection errors
+    ccp_replies_total: Counter,
+    // how long we wait for a reply from ccp
+    ccp_request_duration_msec: Histogram,
+    // how many proofs we submitted
+    cpp_proofs_submitted: Counter,
```
justprosh, line 28:
```suggestion
    ccp_proofs_submitted: Counter,
```

crates/peer-metrics/src/chain_listener.rs
```diff
@@ -0,0 +1,155 @@
...
+    // how many proofs we failed to submit
+    cpp_proofs_submit_failed: Counter,
```
justprosh, line 30: ditto

crates/chain-listener/src/listener.rs
```diff
@@ -626,8 +645,10 @@ impl ChainListener {
     ) -> eyre::Result<()> {
         let header = event.ok_or(eyre!("Failed to process newHeads event: got None"))?;
 
-        // TODO: add block_number to metrics
-        let (block_timestamp, _block_number) = Self::parse_block_header(header?)?;
+        let (block_timestamp, block_number) = Self::parse_block_header(header?)?;
+        self.metrics
+            .as_ref()
+            .inspect(|m| m.observe_new_block(block_number.to_string()));
```
justprosh, line 651: maybe we can wrap these long calls with smth?

crates/chain-listener/src/listener.rs
```diff
@@ -664,7 +685,9 @@ impl ChainListener {
                 }
             }
         }
-
+        self.metrics
+            .as_ref()
+            .inspect(|m| m.observe_processed_block());
```
justprosh, line 690: ditto

justprosh (approved):
Overall very cool! Thx very much for explanatory comments 🙏🙏🙏

crates/chain-listener/src/listener.rs
kmd-fl: it's already wrapped with the experimental `inspect` feature instead of using `if let Some` everywhere. :see_no_evil: I'm not sure how we can wrap it more. Maybe if macros, but I don't see much use in it

justprosh: Maybe it's okay to make method for listener `observe_new_block` which will do that or just `observe` and pass lambda to it

crates/chain-listener/src/lib.rs
```diff
@@ -2,6 +2,8 @@
 #![feature(try_blocks)]
 #![feature(extract_if)]
 #![feature(btree_extract_if)]
+#![feature(result_option_inspect)]
+#![feature(async_closure)]
```
gurinderu, line 6: hm, looks like you don't have async closures

crates/chain-listener/src/listener.rs
```diff
@@ -399,17 +418,18 @@ impl ChainListener {
...
-                    .map_err(|err| {
-                        tracing::warn!(target: "chain-listener", "Error reallocating utility core {utility_core} to CCP, error: {err}. Retrying...");
-                        eyre::eyre!("Error reallocating utility core {utility_core} to CCP, error: {err}")
-                    })?;
-                Ok(())
-            }).await?;
+            measured_request(&self.metrics, async ||
+                retry(ExponentialBackoff::default(), || async {
+                    ccp_client
+                        .realloc_utility_cores(vec![utility_core])
```
gurinderu, line 424: I assume here we can use all sets of utility cores, not the first

kmd-fl: I think behavioral changes should be in a different PR

crates/chain-listener/src/listener.rs
justprosh: Hm, but we have really long method and it becomes less readable with these multiline calls, so I prefer to hide ugliness inside

crates/chain-listener/src/listener.rs
```diff
@@ -1239,3 +1284,22 @@ fn all_min_proofs_found(priority_units: &[CUID]) -> bool {
...
+    metrics: &Option<ChainListenerMetrics>,
+    f: Fn,
+) -> Result<R, E>
+where
+    Fn: FnOnce() -> Fut,
+    Fut: Future<Output = Result<R, E>> + Sized,
+{
+    metrics.as_ref().inspect(|m| m.observe_ccp_request());
+    let start = Instant::now();
+    let result = f().await;
```
gurinderu, line 1299: Do you think you need a function here? Future enough?

gurinderu: so, try to use async block instead of async closure, should work and allows you to drop all async closures

folex (commented):
nice!

crates/peer-metrics/src/chain_listener.rs
```diff
@@ -0,0 +1,155 @@
...
+    // how many proofs we submitted
+    ccp_proofs_submitted: Counter,
+    // how many proofs we failed to submit
+    ccp_proofs_submit_failed: Counter,
+    // how many proofs transaction are ok
+    ccp_proofs_tx_success: Counter,
+    // how many proofs transaction are failed
+    ccp_proofs_tx_failed: CounterWithExemplar<TxLabel>,
+    // How many blocks we have received from the newHead subscription
+    blocks_seen: CounterWithExemplar<BlockLabel>,
```
folex, line 36: shouldn't this be a gauge with a block number?

crates/peer-metrics/src/chain_listener.rs
```diff
@@ -0,0 +1,155 @@
...
+    // How many block we manage to process while processing the block
+    blocks_processed: Counter,
```
folex, line 38: and this, too

PR #2219, opened 2024-04-08
fix: add block number value to the metrics

kmd-fl:
## Description

Add new fields `last_seen_block` and `last_processed_block` displaying the block number of the block from newHeader subscription handler.

## Motivation
It will be helpful for alerts to detect chain de-synch

## Proposed Changes
- add `last_seen_block: Gauge` 
- add `last_processed_block: Gauge` 
- use lower bits of the block number in metrics 

## Additional Notes
- We use `i64` since `u64` and hight aren't supported in prometheus due to [some protobuf restriction](https://github.com/prometheus/client_rust/pull/189)
- It's fine to use only lower bits of the u256 block number since:
   1. most likely, we won't reach i64 _soon_
   2. these metrics will be used in alerts to detect the chain de-sync, lower bits should be enough for this task.

### DIFF
--- a/crates/chain-listener/src/listener.rs
+++ b/crates/chain-listener/src/listener.rs
@@ -647,7 +647,7 @@ impl ChainListener {
         let header = event.ok_or(eyre!("Failed to process newHeads event: got None"))?;
 
         let (block_timestamp, block_number) = Self::parse_block_header(header?)?;
-        self.observe(|m| m.observe_new_block(block_number.to_string()));
+        self.observe(|m| m.observe_new_block(block_number));
 
         // `epoch_number = 1 + (block_timestamp - init_timestamp) / epoch_duration`
         let epoch_number =
@@ -684,7 +684,7 @@ impl ChainListener {
                 }
             }
         }
-        self.observe(|m| m.observe_processed_block());
+        self.observe(|m| m.observe_processed_block(block_number));
         Ok(())
     }
 
@@ -1117,7 +1117,7 @@ impl ChainListener {
         }
     }
 
-    fn parse_block_header(header: Value) -> eyre::Result<(U256, U256)> {
+    fn parse_block_header(header: Value) -> eyre::Result<(U256, i64)> {
         let obj = header.as_object().ok_or(eyre::eyre!(
             "newHeads: header is not an object; got {header}"
         ))?;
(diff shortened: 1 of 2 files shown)


crates/chain-listener/src/listener.rs
```diff
@@ -1117,7 +1117,7 @@ impl ChainListener {
         }
     }
 
-    fn parse_block_header(header: Value) -> eyre::Result<(U256, U256)> {
+    fn parse_block_header(header: Value) -> eyre::Result<(U256, i64)> {
```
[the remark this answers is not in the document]
kmd-fl: yes

crates/chain-listener/src/listener.rs
```diff
@@ -1138,7 +1138,13 @@ impl ChainListener {
             ))?
             .to_string();
 
-        Ok((U256::from_str(&timestamp)?, U256::from_str(&block_number)?))
+        let block_number = U256::from_str(&block_number)?;
+        // take lower bits from the number (they are in the order 'least significant first')
+        // we're okay with ignoring higher bits since they 1) probably, won't be reached in practice
+        // 2) don't help alerts very much, lower bits are enough
+        let block_number = block_number.as_limbs()[0] as i64;
```
folex, line 1145: I think this should live inside Metrics only. Imagine someone uses this function not for metrics (it looks appealing, right), and gets burned by `i64`.

Maybe make `observe_new_block` and `observe_processed_block` to take U256, wdyt?

crates/peer-metrics/src/chain_listener.rs
```diff
@@ -144,12 +157,13 @@ impl ChainListenerMetrics {
...
-    pub fn observe_new_block(&self, block_number: String) {
-        self.blocks_seen
-            .inc_by(1, Some(BlockLabel { block_number }));
+    pub fn observe_new_block(&self, block_number: i64) {
+        self.blocks_seen.inc();
+        self.last_seen_block.set(block_number);
     }
 
-    pub fn observe_processed_block(&self) {
+    pub fn observe_processed_block(&self, block_number: i64) {
```
folex, line 165: Like, take U256 here, and take lower limbs, all that

folex (approved):

crates/chain-listener/src/listener.rs
kmd-fl: I think it would also give a wtf moment if a Metrics interface accepts U256 but prints only i64.

kmd-fl: Also, it would require introducing a new dependency with the U256 type to peer-metrcis module only for a type that's not even used

kmd-fl: Omg, in the allow_primitives that exposes the U256 type, they have a `BlockNumber` type that is a alias for `u64`.

PR #2249, opened 2024-05-16
fix(listener): listener fixes

justprosh:
- don't print misleading logs if all units will be activated next epoch
- reset proof id on global nonce change
- persist corrent proof id
- release and acquire core for deal in refresh


crates/chain-listener/src/listener.rs
```diff
@@ -848,6 +849,16 @@ impl ChainListener {
             return Ok(());
         }
 
+        if self
+            .cc_compute_units
+            .iter()
+            .all(|(_, cu)| cu.startEpoch > self.current_epoch)
```
folex, line 855: I'm not sure I understand this condition. "If all CUs have started, then stop all CCs" – doesn't make much sense, I guess I read it wrong?

Could you please move the condition to a separate variable, and give it a sensible name, so your intention is clear?

crates/chain-listener/src/listener.rs
```diff
@@ -189,6 +189,11 @@ impl ChainListener {
                 }
                 tracing::info!(target: "chain-listener", "Subscribed successfully");
 
+                if let Err(err) = self.load_proof_id().await {
```
folex, line 192: If every such hotfix would have a comment explaining the motivation for it, the future refactoring would be much much easier.

crates/chain-listener/src/listener.rs
```diff
@@ -366,6 +370,8 @@ impl ChainListener {
     }
 
     async fn set_proof_id(&mut self, proof_id: ProofIdx) -> eyre::Result<()> {
+        self.last_submitted_proof_id = proof_id;
```
folex, line 373: Wouldn't it make more sense to first write new proof to disk, and only then mutate in-memory state?

I'm not sure what the failure semantics of the `set_proof_id` is, though.

crates/chain-listener/src/listener.rs
```diff
@@ -543,29 +544,35 @@ impl ChainListener {
...
+
         tracing::info!(target: "chain-listener",
             "Compute units mapping: in cc {}/[{} pending], in deal {}",
             self.cc_compute_units.len(),
             pending.clone().count(),
-            self.active_deals.len()
+            in_deal.len()
         );
 
+        for cu in in_deal {
```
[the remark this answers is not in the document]
justprosh: I wanted to move interactions with core-manager after all the logs, so in case of failure with cores allocation we will see all the data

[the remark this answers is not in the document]
folex: Let's state that in comments. Maybe you also want to print more info on error?

PR #2291, opened 2024-06-14
feat: implement batching [VM-614]

justprosh:


crates/chain-listener/src/proof_tracker.rs
```diff
@@ -0,0 +1,171 @@
...
+
+    pub async fn confirm_proof(&mut self, cu_id: CUID) -> U256 {
+        let proof_id = Uint::from(
+            *self
+                .proof_counter
+                .entry(cu_id)
+                .and_modify(|c| {
+                    *c = c.add(Uint::from(1));
+                })
+                .or_insert(Uint::from(1)),
```
monoid, line 96:
```
let proof_count = self.proof_counter.entry(cu_id).or_insert(Uint::from(0));
*proof_count += 1;
let proof_id = Uint::from(*proof_count);
```
would be more explicit.

monoid: The top-level `Uint::from`, is it really needed?

crates/chain-listener/src/listener.rs
```diff
@@ -1075,72 +1041,130 @@ impl ChainListener {
...
-        let result_hash = ResultHash::from_slice(*self.difficulty.as_ref());
-
-        // proof_id is used only by CCP and is not sent to chain
-        let proof_id = CCProofId::new(self.global_nonce, self.difficulty, ProofIdx::zero());
-        let units = self.cc_compute_units.keys().cloned().collect::<Vec<_>>();
-        for unit in units {
-            let local_nonce = LocalNonce::random();
-            self.submit_proof(CCProof::new(proof_id, local_nonce, unit, result_hash))
-                .await?;
+    // /// Submit Mocked Proofs for all active compute units.
```
monoid, line 1044: More commented code.

crates/chain-listener/src/listener.rs
```diff
@@ -1075,72 +1041,130 @@ impl ChainListener {
...
+            let proof_batches = if self.is_epoch_ending() {
+                let last_known_proofs = batch_requests
+                    .into_iter()
+                    .map(|(cu_id, req)| (cu_id, req.last_seen_proof_idx))
+                    .collect();
+
+                tracing::debug!(target: "chain-listener", "Polling proofs after {:?}", last_known_proofs);
+                measured_request(
+                    &self.metrics,
+                    async { ccp_client.get_proofs_after(last_known_proofs, PROOF_POLL_LIMIT) }
```
monoid, line 1098: The `async { X }.await` construction may be replaced by just `X`.  Here and elsewhere.

crates/chain-connector/src/connector.rs
```diff
@@ -806,6 +816,14 @@ impl ChainConnector for HttpChainConnector {
     }
 }
 
+fn parse_str_field(value: Option<Value>, field: &str) -> eyre::Result<String> {
+    value
+        .ok_or(eyre!("Field {} not found in response", field))?
```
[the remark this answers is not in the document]
gurinderu: it should be enum without eyre)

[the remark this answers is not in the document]
monoid: Even better.

.github/workflows/build.yml
```diff
@@ -11,6 +11,10 @@ on:
         description: "Cargo dependencies map"
         type: string
         default: "null"
+    secrets:
```
gurinderu, line 14: Why did you change that in this pr?

justprosh: it was done by @enjenjenje because of private repo deps

crates/chain-connector/src/connector.rs
```diff
@@ -468,25 +473,28 @@ impl HttpChainConnector {
             gas_limit: Uint256::from_le_bytes(&gas_limit.to_le_bytes_vec()),
             to: to.parse()?,
             value: 0u32.into(),
-            data,
+            data: data.clone(),
```
gurinderu, line 479: Data can be a pretty big, probably it is better to apply hex::decode before this line and do not use cloning ?

crates/chain-listener/src/proof_tracker.rs
```diff
@@ -0,0 +1,171 @@
...
+            self.last_submitted_proof_ids.clear();
+            self.persist().await;
+            true
+        } else {
+            false
+        }
+    }
+
+    async fn persist(&self) {
+        let backoff = ExponentialBackoff {
```
gurinderu, line 134: Why do we need a backoff for the persistence ?

crates/chain-data/src/block_header.rs
```diff
@@ -0,0 +1,56 @@
...
+            .ok_or(eyre::eyre!(" timestamp field not found; got {json}"))?
+            .to_string();
+
+        let block_number = obj
+            .get("number")
+            .and_then(Value::as_str)
+            .ok_or(eyre::eyre!("number field not found; got {json}"))?
+            .to_string();
+
+        Ok(Self {
```
monoid, line 45: The type may be deserialized from a `Value` with serde in more declarative manner.
https://docs.rs/serde_json/latest/serde_json/fn.from_value.html

PR #2298, opened 2024-07-02
chore: Upload nox binaries to s3

nahsi:


.github/workflows/publish.yml
```diff
@@ -118,3 +111,20 @@ jobs:
...
+          files: |
+            ./docker/nox-amd64
+            ./docker/nox-arm64
+            ./docker/nox_SHA256_SUMS
+          tag_name: ${{ inputs.tag }}
+
+      - name: Upload nox binary to s3
+        uses: osiegmar/s3-publisher-action@v1
+        with:
+          bucket: fluence-artifacts
```
folex, line 127: Let's use a less generic bucket? Like `nox-releases` or `nox-distribution`. 

So that eg all CCP binaries and all Nox binaries are clearly laid out without any noise around.

nahsi: It is not only for nox but for all artifacts we have/we can have. Creating and configuring a bucket for each distribution adds a lot of code.

folex: We have like just two S3 distributions: CLI and Nox+CCP.

It's benefical to keep things organised. Can change permissions there, easier to review and so on. Less noise, less brain power needed to manage it.

folex (approved):

PR #2306, opened 2024-07-07
feat: track nonce

justprosh:


crates/chain-connector/src/connector.rs
```diff
@@ -487,12 +487,25 @@ impl HttpChainConnector {
...
                 .await,
-        )?;
-        Ok(resp)
+        );
+
+        match result {
+            Ok(resp) => {
+                let new_nonce = nonce + Uint::from(1);
+                *nonce_guard = Some(new_nonce);
+                tracing::debug!(target: "chain-connector", "Incrementing nonce. New nonce {new_nonce}");
```
folex, line 500: maybe move to info for now?

or better ask providers to enable debug for chain connector?

crates/chain-connector/src/connector.rs
```diff
@@ -82,7 +79,7 @@ pub trait ChainConnector: Send + Sync {
 pub struct HttpChainConnector {
     client: Arc<jsonrpsee::http_client::HttpClient>,
     config: ChainConfig,
-    tx_nonce_mutex: Arc<Mutex<()>>,
+    tx_nonce_mutex: Arc<Mutex<Option<U256>>>,
```
folex, line 82: i think you could be fine with atomic u64, no mutex and probably no error/result

folex (approved):

crates/chain-connector/src/connector.rs
justprosh: Nope, I need mutex to send only one tx at a time. In theory decider can send txs at the same time

crates/chain-connector/src/connector.rs
justprosh: I think better to ask for debug for now, I don't like to make it info temporary

PR #2314, opened 2024-07-19
fix(chain): prefix data with 0x in eth_calls

folex:
Per Ethereum Spec https://eips.ethereum.org/EIPS/eip-1474#data, Data values must be prefixed with 0x.

Some blockchains – ie IPC – did allow us to pass it without `0x` prefix, but others – ie Gelato – return an error in that case.

> invalid argument 0: json: cannot unmarshal hex string without 0x prefix into Go struct field TransactionArgs.data of type hexutil.Bytes

This PR brings Nox one step closer to the ETH RPC spec. I will test it manually before merging.


justprosh (approved):

gurinderu:
Where is the pr description? I didn't understand why do we need that?

gurinderu (approved):

PR #2321, opened 2024-07-23
chore: Upload snapshots to s3

nahsi:


.github/workflows/build.yml
```diff
@@ -33,6 +33,10 @@ jobs:
             target: aarch64-unknown-linux-gnu
             runner: linux-arm64-c-2xlarge
 
+    permissions:
+      contents: read
+      id-token: write
```
folex, line 38: Can you add a comment for that block? no idea what it means

.github/workflows/build.yml
```diff
@@ -90,3 +94,27 @@ jobs:
...
+          method: jwt
+          jwtGithubAudience: "https://github.com/fluencelabs"
+          jwtTtl: 300
+          secrets: |
+            kv/ci/fluence-artifacts id | AWS_ACCESS_KEY_ID ;
+            kv/ci/fluence-artifacts secret | AWS_SECRET_ACCESS_KEY
+
+      - name: Rename nox binary
+        run: |
+          mkdir s3
```
folex, line 113:
```suggestion
          mkdir -p s3
```
so it doesn't fail even if dir already exists?

.github/workflows/build.yml
```diff
@@ -90,3 +94,27 @@ jobs:
...
+          cp target/${{ matrix.target }}/${{ steps.profile.outputs.profile }}/nox s3/nox-${{ matrix.arch }}
```
folex, line 114:
```suggestion
          cp "target/${{ matrix.target }}/${{ steps.profile.outputs.profile }}/nox" "s3/nox-${{ matrix.arch }}"
```

.github/workflows/build.yml
```diff
@@ -33,6 +33,10 @@ jobs:
             target: aarch64-unknown-linux-gnu
             runner: linux-arm64-c-2xlarge
 
+    permissions:
```
nahsi, line 36:
```suggestion
    # requirement for vault-action
    permissions:
```

folex (approved):

.github/workflows/build.yml
nahsi: No need really cause no bash spaces or bash variables there

folex:
Will it print links into the Summary tab?

nahsi:
> Will it print links into the Summary tab?

Summary already has artifacts uploaded directly

PR #2340, opened 2024-08-01
feat(vm): add vm builtins for reboot, reset and status [BRND-23]

kmd-fl:
## Description
Add builtins for user commands for reboot/reset/start/stop/status.

## Motivation
A user should have a way to control their VM in case something wrong happens with the VM. 

## Proposed Changes
- Add new builtins
- Allow only workers to call them
- Users will interact with the builtins via the vm-spell which will call the builtins
- Except for the `status` builtin, which can be called by everyone


crates/vm-utils/src/vm_utils.rs
```diff
@@ -64,17 +65,79 @@ pub enum VMUtilsError {
...
+    Shutdown,
+    Shutoff,
+    Crashed,
+    PMSuspended,
+    UnknownState(u32),
 }
 
-pub fn create_domain(uri: &str, params: &CreateVMDomainParams) -> Result<(), VMUtilsError> {
-    let conn = Connect::open(Some(uri)).map_err(|err| VMUtilsError::FailedToConnect { err })?;
+impl Display for VmStatus {
```
gurinderu, line 107: maybe https://docs.rs/strum_macros/latest/strum_macros/derive.Display.html ?

kmd-fl: why add a dependency on a crate for one small thing?

crates/vm-utils/src/vm_utils.rs
```diff
@@ -64,17 +65,79 @@ pub enum VMUtilsError {
...
+            virt::sys::VIR_DOMAIN_SHUTOFF => VmStatus::Shutoff,
+            virt::sys::VIR_DOMAIN_CRASHED => VmStatus::Crashed,
+            virt::sys::VIR_DOMAIN_PMSUSPENDED => VmStatus::PMSuspended,
+            _ => VmStatus::UnknownState(value),
+        }
+    }
+}
+
+pub fn create_domain(uri: &str, params: &CreateVMDomainParams) -> Result<(), VmError> {
+    let conn = Connect::open(Some(uri)).map_err(|err| VmError::FailedToConnect { err })?;
```
monoid, line 140: If `FailedToConnect` was a tuple variant, it would be just `.map_err(VmError::FailedToConnect)`.  But that's very minor thing.

crates/workers/src/workers.rs
```diff
@@ -699,12 +699,64 @@ impl Workers {
...
+    }
+
+    fn on_vm<F, R>(&self, worker_id: WorkerId, f: F) -> Result<R, WorkersError>
+    where
+        F: Fn(&str, &str) -> Result<R, vm_utils::VmError>,
+    {
+        match &self.config.vm {
+            Some(vm_config) => {
+                if self.has_vm(worker_id)? {
+                    let result = f(vm_config.libvirt_uri.as_str(), &worker_id.to_string())?;
```
monoid, line 741: The `result` is not `Result`, is it?

kmd-fl: It's `Result`.

kmd-fl: oh yeah, I meant, `f` returns a `Result`, but the whole expression is not, it's with `?`

monoid: I meant the variable type.

crates/vm-utils/src/vm_utils.rs
gurinderu: true

gurinderu (approved):

PR #2350, opened 2024-08-14
fix: suspend/resume on worker deactivate/activate

kmd-fl:
## Description
Pause VMs on worker deactivation and resume on activation

## Motivation
To reflect the deal state on chain

## Additional Notes
- Tested manually, deactivation and activation worked as expected


mikevoronov (approved):
utACK

PR #2355, opened 2024-08-22
feat(vm): add all GPUs on the host to a VM

kmd-fl:
## Description

Add all the GPUs to the XML of a newly created VM 

## Proposed changes

1. When creating VMs, get the info about GPUs and add them to the VM XML
2. Added the config setting that allows to enable/disable this functionallity
```
[vm]
allow_gpu = true
```
By default, GPUs are disabled


crates/gpu-utils/src/lib.rs
```diff
@@ -0,0 +1,38 @@
...
+pub fn get_gpu_pci() -> Result<Vec<PciLocation>, PciError> {
+    let info = PciInfo::enumerate_pci()?;
+    let mut result = Vec::new();
+    for device in info {
+        let device = device?;
+        let device_class = process_property_result(device.device_class())?;
+        if device_class == DisplayController {
+            result.push(process_property_result(device.location())?);
+        }
+    }
```
mikevoronov, line 28: Let's do it in the functional style.

mikevoronov (approved):
utACK

crates/gpu-utils/src/lib.rs
kmd-fl: I wanted but then with all those `?` I decided to do it straightforward

PR #2362, opened 2024-09-09
fix(vm): allow only nvidia and amd gpus

kmd-fl:
## Description

Allow only NVidia and AMD GPUs.  This way we try to increase the probability that iGPU aren't imported. Still need to point out that at least AMD iGPUs exist.


mikevoronov (approved):
utACK

PR #2365, opened 2024-09-12
feat(listener): skip proofs with wrong nonce

justprosh:


crates/chain-listener/src/listener.rs
```diff
@@ -1154,34 +1154,24 @@ impl ChainListener {
...
-                tracing::info!(target: "chain-listener", "Found {} proofs in {} batches from polling", total_proofs, proof_batches.len());
-
                 let mut unit_ids = Vec::new();
                 let mut local_nonces = Vec::new();
                 let mut result_hashes = Vec::new();
+                let batch_count = proof_batches.len();
+                let mut skipped_proofs_count = 0;
                 for batch in proof_batches.into_iter() {
                     for proof in batch.proof_batches.into_iter() {
+                        if proof.id.global_nonce != self.global_nonce {
```
monoid, line 1169: Perhaps, the difficulty should be checked too.

monoid: Yes, I think the difficulty should be checked definitely.
@mikevoronov your opinion?

justprosh: i am adding it, no problem

mikevoronov (approved):
utACK

PR #2375, opened 2024-09-16
feat(vm): check if virtualization is enabled

kmd-fl:
## Description
Check if virtualization is enabled if there's VM configuration in Nox

### DIFF
--- a/nox/src/main.rs
+++ b/nox/src/main.rs
@@ -302,6 +302,7 @@ fn check_virtualization() -> eyre::Result<()> {
     match kvm_ok {
         Ok(output) => {
             if output.contains("KVM acceleration can be used") {
+                log::info!("Virtualization is enabled");
                 Ok(())
             } else {
                 Err(eyre::eyre!(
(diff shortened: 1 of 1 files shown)


nox/src/main.rs
```diff
@@ -284,3 +288,32 @@ fn vm_config(config: &ResolvedConfig) -> VmConfig {
...
+//    If it's not available, return a warning.
+fn check_virtualization() -> eyre::Result<()> {
+    let kvm_ok = std::process::Command::new("kvm-ok")
+        .output()
+        .map(|output| String::from_utf8_lossy(&output.stdout).trim().to_string());
+
+    match kvm_ok {
+        Ok(output) => {
+            if output.contains("KVM acceleration can be used") {
+                Ok(())
```
mikevoronov, line 305: let's add log here that virtualization check was successful

mikevoronov (approved):
utACK

nox/src/main.rs
kmd-fl: done

PR #2386, opened 2024-09-23
feat(metrics): add chain info metrics

kmd-fl:
## Description
Add a chain info metric with chain configuration

## Proposed Changes
- Refactor a bit to remove `server-config` dependency on `peer-metrics`. It turned out to be useless in this PR, but maybe will be helpful in the next one with more configs
- Add `current_commitment_status`. The status is a number which corresponds to the CCStatus enum in chain_listener.
- Add `nox_chain_info` metric. Example:
```
# HELP nox_chain Chain Nox Info.
# TYPE nox_chain info
nox_chain_info{peer_id="0x123",http_endpoint="http://localhost:8545",diamond_contract_address="0x123",network_id="1",default_base_fee="1",default_priority_fee="1",ws_endpoint="ws://localhost:8546",proof_poll_period="1s",min_batch_count="1",max_batch_count="1",max_proof_batch_size="1",epoch_end_window="1s"} 1
# EOF
```

### DIFF
--- a/crates/peer-metrics/src/info.rs
+++ b/crates/peer-metrics/src/info.rs
@@ -100,35 +100,3 @@ pub fn add_info_metrics2(
     ]);
     sub_registry.register("build", "Nox Info", info);
 }
-
-#[test]
-fn test() {
-    let mut reg = Registry::default();
-    let info = NoxInfo {
-        versions: NoxVersions {
-            node_version: "0.1.0".to_string(),
-            air_version: "0.1.0".to_string(),
-            spell_version: "0.1.0".to_string(),
-        },
-        chain_info: ChainInfo {
-            peer_id: "0x123".to_string(),
-            http_endpoint: "http://localhost:8545".to_string(),
-            diamond_contract_address: "0x123".to_string(),
-            network_id: 1,
-            default_base_fee: Some(1),
-            default_priority_fee: Some(1),
-            ws_endpoint: "ws://localhost:8546".to_string(),
-            proof_poll_period_secs: 1,
-            min_batch_count: 1,
-            max_batch_count: 1,
-            max_proof_batch_size: 1,
-            epoch_end_window_secs: 1,
-        },
-    };
-    add_info_metrics(&mut reg, info);
-
-    let mut buf = String::new();
-
-    encode(&mut buf, &reg);
-    println!("{buf}");
-}
--- a/crates/peer-metrics/src/info.rs
+++ b/crates/peer-metrics/src/info.rs
@@ -22,13 +22,13 @@ use prometheus_client::metrics::info::Info;
 use prometheus_client::registry::Registry;
 
 pub struct NoxInfo  {
-    pub versions: NoxVersions,
+    pub version: NoxVersion,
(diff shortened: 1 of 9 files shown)


crates/peer-metrics/src/info.rs
```diff
@@ -17,20 +17,118 @@
...
 use prometheus_client::registry::Registry;
 
+pub struct NoxInfo  {
+    pub versions: NoxVersions,
+    pub chain_info: ChainInfo,
+}
+
+
+#[derive(Debug, Clone, Hash, Eq, PartialEq, EncodeLabelSet)]
+pub struct NoxVersions {
```
folex, line 31: imo singular "version" suits better

crates/peer-metrics/src/info.rs
```diff
@@ -17,20 +17,118 @@
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
 
+use prometheus_client::encoding::EncodeLabelSet;
 use prometheus_client::metrics::info::Info;
 use prometheus_client::registry::Registry;
 
+pub struct NoxInfo  {
+    pub versions: NoxVersions,
```
folex, line 25: was expecting array here

crates/peer-metrics/src/info.rs
```diff
@@ -17,20 +17,118 @@
...
+    pub ws_endpoint: String,
+    pub proof_poll_period_secs: u64,
+    pub min_batch_count: usize,
+    pub max_batch_count: usize,
+    pub max_proof_batch_size: usize,
+    pub epoch_end_window_secs: u64,
+}
+
+impl ChainInfo {
+    pub fn empty(peer_id: String) -> ChainInfo {
```
folex, line 57: default?

crates/peer-metrics/src/info.rs
```diff
@@ -17,20 +17,118 @@
...
+    let sub_registry = registry.sub_registry_with_prefix("nox");
+
+    let info = Info::new(nox_info.versions);
+    sub_registry.register("build", "Nox Info", info);
+
+    let chain_info = Info::new(nox_info.chain_info);
+    sub_registry.register("chain", "Chain Nox Info", chain_info);
+}
+
+pub fn add_info_metrics2(
```
folex, line 88: why 2? maybe remove 1st?

folex (approved):
cc status in the next pr?

crates/peer-metrics/src/info.rs
kmd-fl: Why? `NoxVersions` implement the `EncodeLabelSet` which creates the array automatically from the structure fields and values

crates/peer-metrics/src/info.rs
kmd-fl: Well, maybe. I would expect that `default` is what implements `Default`, and I'm not sure if this thing should implement default, but whatever

folex: I think it's a good idea to move `empty` to Default implementation, that's the rust way of things

PR #2390, opened 2024-10-02
feat(metrics): add CC metrics

kmd-fl:
## Description
Add more CC metrics:
```
    // max amount of proofs are allowed
    ccp_proofs_per_epoch_allowed_max: Gauge,
    // min amount of proofs are allowed
    ccp_proofs_per_epoch_allowed_min: Gauge,

    // CUs metrics
    cus_total: Gauge,
    cus_in_deals: Gauge,

    // Epoch Metrics
    current_epoch: Gauge,
    current_epoch_start_timestamp_sec: Gauge,
    current_epoch_duration_sec: Gauge,
```

### DIFF
--- a/crates/peer-metrics/src/chain_listener.rs
+++ b/crates/peer-metrics/src/chain_listener.rs
@@ -17,7 +17,6 @@
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
 
-use crate::{execution_time_buckets, register};
 use prometheus_client::encoding::EncodeLabelSet;
 use prometheus_client::metrics::counter::Counter;
 use prometheus_client::metrics::exemplar::CounterWithExemplar;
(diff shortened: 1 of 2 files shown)


crates/peer-metrics/src/chain_listener.rs
```diff
@@ -53,14 +54,32 @@ pub struct ChainListenerMetrics {
...
     current_commitment: Family<CommitmentLabel, Gauge>,
+
+    // CUs metrics
+    cus_total: Gauge,
+    cus_in_deals: Gauge,
+
+    // Epoch Metrics
+    current_epoch: Gauge,
+    current_epoch_start_timestamp_sec: Gauge,
+    current_epoch_duration_sec: Gauge,
```
folex, line 82: I wonder if you can use `Gauge<u64, AtomicU64>` to avoid `as i64` everywhere

folex (approved):

PR #2392, opened 2024-10-09
fix(vm): close libvirt connections

kmd-fl:
## Description
Close libvirt connection when we're done.

## Motivation
We found a problem with `Too many files` on one of our provider's machines when deploying a deal on it: the nox process has too many open eventfd connections. Further investigation showed that the problem is that the `Connection` struct of the `virt` crate doesn't implement `Drop`, so the connection isn't closed. 

## Proposed Changes
Auto-close connection.


mikevoronov (approved):
utACK

